PR #2, opened 2020-06-16
Update the instructions to setup the Gradle plugin.

vRallev:


README.md
```diff
@@ -33,14 +33,28 @@ Notice that `AppComponent` automatically includes `DaggerModule` and extends `Co
...
 ```groovy
-apply plugin: 'com.squareup.hephaestus'
+plugins {
+  id 'com.squareup.hephaestus' version "${latest_version}"
+}
 ```
 
-The Gradle plugin needs to be applied in all modules that either contribute classes to the
-dependency graph or merge them.
+Or you can use the old way to apply a plugin:
```
FrostRocket, line 45: Minor comment here: if you know the Gradle version it would help to note it here. Your call.

i.e. "Or, you can apply the plugin if you're on Gradle X.XX or older."

vRallev: I don't know with which version this was introduced, but it was a long time ago. Gradle wants you to use the new way, but we also use the old way.

PR #3, opened 2020-06-16
Run all CI jobs in parallel.

vRallev:


.github/workflows/ci.yml
```diff
@@ -7,41 +7,25 @@ on:
...
     strategy:
-      # Run all checks, even if some fail.
+      # Run all tasks, even if some fail. Note that they don't share an output, some tasks overlap
+      # which is expected. If they need to share their outputs, then we need a proper caching
+      # solution.
       fail-fast: false
       matrix:
         gradle-task:
-          # Unit tests
+          - assemble
```
zach-klippenstein, line 21: I think these can all be combined by just running `build`.

vRallev: It also runs `lint`, which I want to have separate.

PR #10, opened 2020-06-21
Add missing year and name to LICENSE.txt

tehras:
Noticed missing information in your LICENSE.txt


tehras:
Actually it was right to begin with. Please ignore.

PR #18, opened 2020-06-27
Configure tasks lazily when they're actually needed. 

vRallev:
This prevents creating task unnecessarily.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
@@ -37,7 +37,7 @@ open class HephaestusPlugin : Plugin<Project> {
   ) {
     project.tasks
         .withType(KotlinCompile::class.java)
-        .all { compileTask ->
+        .configureEach { compileTask ->
           val checkMixedSourceSet = project.tasks.register(
               compileTask.name + "CheckMixedSourceSetHephaestus",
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -37,7 +37,7 @@ open class HephaestusPlugin : Plugin<Project> {
   ) {
     project.tasks
         .withType(KotlinCompile::class.java)
-        .all { compileTask ->
+        .configureEach { compileTask ->
           val checkMixedSourceSet = project.tasks.register(
               compileTask.name + "CheckMixedSourceSetHephaestus",
               CheckMixedSourceSetTask::class.java
```
ZacSweers, line 43: I think you'll likely just want to add a `doFirst` here rather than use task ordering, as you can't register a task inside another's configuration like this. Only reason to keep it a separate task is if you wanted to make its checks cacheable/incremental, which this isn't currently. If you did keep it as a separate task, it's probably better to base this entirely on source sets rather than reading from the compile task.

vRallev: Yes, there is more work involved. The main issue is actually that instant execution / "configuration caching" isn't supported at the moment. This was another issue that somebody discovered internally and I thought it would be a quick fix. I was wrong. 

I'll work on the instant execution support in a separate PR.

PR #19, opened 2020-06-28
Support instant execution / configuration caching

vRallev:


gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -22,26 +24,41 @@ open class HephaestusPlugin : Plugin<Project> {
...
       }
     }
-    with(project.pluginManager) {
-      withPlugin("org.jetbrains.kotlin.android", innerApply)
-      withPlugin("org.jetbrains.kotlin.jvm", innerApply)
+
+    project.pluginManager.withPluginOnce("org.jetbrains.kotlin.android") {
+      realApply(project, true)
     }
+    project.pluginManager.withPluginOnce("org.jetbrains.kotlin.jvm") {
```
stephanenicolas, line 42: Can you add a comment on why this is necessary, please ?

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/CheckMixedSourceSet.kt
```diff
@@ -20,11 +18,12 @@ import java.io.File
  * these module using this task.
```
stephanenicolas, line 18: Can you add that this operation is performed at configuration time ?

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/DisableIncrementalCompilationTask.kt
```diff
@@ -1,60 +1,43 @@
...
+abstract class DisableIncrementalCompilationTask : DefaultTask() {
+
+  @Suppress("unused")
+  @get:Classpath
+  @get:InputFiles
+  abstract val pluginClasspath: ConfigurableFileCollection
 
   @Suppress("unused")
   @get:OutputFile @get:Optional
   val someFile = File(project.buildDir, "not-existing-file-because-gradle-needs-an-output")
```
stephanenicolas, line 28: Can you use a RegularFileProperty here ? It would be more homogenous with the property below.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/CheckMixedSourceSet.kt
vRallev: It's not, it's called in a doFirst and this is fine:
```
    project.tasks
        .configureEach { compileTask ->
          compileTask.doFirst {
            // Disable precise java tracking if needed. Note that the doFirst() action only runs
            // if the task is not up to date. That's ideal, because if nothing needs to be
            // compiled, then we don't need to disable the flag.
            CheckMixedSourceSet(project, compileTask).disablePreciseJavaTrackingIfNeeded()

            compileTask.logger.info(
                "Hephaestus: Use precise java tracking: ${compileTask.usePreciseJavaTracking}"
            )
          }
        }
```

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/DisableIncrementalCompilationTask.kt
vRallev: I keep it for now. It's less code, less configuration and no unnecessary file is created on the file system that needs to be cached, packaged and uploaded.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: Done

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/CheckMixedSourceSet.kt
stephanenicolas: I am surprised a `doFirst` block can access other tasks with CoCa.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/DisableIncrementalCompilationTask.kt
stephanenicolas: It's not so much of a change and it's the new standard way of doing things.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/CheckMixedSourceSet.kt
stephanenicolas: Oh, but it's the same task ! So yeah, it's all good.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
-            else -> {
-              val variant = project.androidVariants()
-                  .findVariantForCompileTask(compileTask)
-
-              "${variant.name}CompileClasspath"
-            }
+        .withType(KaptGenerateStubsTask::class.java)
+        .configureEach { stubsTask ->
+          // Disable incremental compilation for the stub generating task. If any of the
+          // dependencies in the compile classpath have changed, then this we need to trigger the
```
FrostRocket, line 105: "Trigger the compiler plugin if any dependencies in the compile classpath have changed."

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
+          compileTask.logger.lifecycle(
+              "Hephaestus: Incremental compilation enabled: ${compileTask.incremental} (compile)"
+          )
         }
+      }
+    }
+
+    if (isAndroidProject) {
+      project.androidVariantsConfigure { variant ->
+        val compileTaskName = "compile${variant.name.capitalize(US)}Kotlin"
```
FrostRocket, line 159: Clever. 👍

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
+        disableIncrementalCompilationAction(compileTaskName)
+      }
+    } else {
+      // The Java plugin has two Kotlin tasks we care about: compileKotlin and compileTestKotlin.
+      val taskNames = listOf("compileKotlin", "compileTestKotlin")
```
FrostRocket, line 164: Nit: I'm not a fan of the intermediate variables that don't serve a purpose. 

`listOf("compileKotlin", "compileTestKotlin")
.forEach(disableIncrementalCompilationAction)`

Would be cleaner in my opinion. Your call!

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -22,26 +24,44 @@ open class HephaestusPlugin : Plugin<Project> {
...
+        }
       }
     }
-    with(project.pluginManager) {
-      withPlugin("org.jetbrains.kotlin.android", innerApply)
-      withPlugin("org.jetbrains.kotlin.jvm", innerApply)
+
+    // Apply the Hephaestus plugin after the Kotlin plugin was applied. There could be a timing
+    // issue. Also make sure to apply it only once. A module could accidentally apply the JVM and
+    // Android Kotlin plugin.
```
cdeange, line 41: Could the ordering matter? eg: if I were to do

```
apply plugin: 'kotlin-android'
apply plugin: 'kotlin'
```

vs

```
apply plugin: 'kotlin'
apply plugin: 'kotlin-android'
```

that would have a difference in behaviour here, correct?

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
+        compileTask.doFirst {
+          // If the signal is set, then the plugin classpath changed. Apply the setting that
+          // DisableIncrementalCompilationTask requested.
+          val incremental = incrementalSignal.get().incremental
+          if (incremental != null) {
+            compileTask.incremental = incremental
           }
+
+          compileTask.logger.lifecycle(
+              "Hephaestus: Incremental compilation enabled: ${compileTask.incremental} (compile)"
```
cdeange, line 151: nit: I usually find that having an if-else and printing either `Incremental compilation enabled` or `Incremental compilation disabled` is easier to read, instead of the true/false.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
-                  DisableIncrementalCompilationTask::class.java
-              ) {
-                it.compileTask = compileTask
-                it.configurationName = configurationNameForIncrementalChecks
-              }
+    // Use this signal to share state between DisableIncrementalCompilationTask and the Kotlin
+    // compile task. If the plugin classpath changed, then DisableIncrementalCompilationTask sets
+    // the signal to false.
+    val incrementalSignal = project.gradle.sharedServices
+        .registerIfAbsent("incrementalSignal${project.path}", IncrementalSignal::class.java) { }
```
cdeange, line 122: Would it make sense to register just a single shared service, and have that maintain a `Map<String, Boolean?>` of incremental signals, instead of one service per project?

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -58,101 +78,129 @@ open class HephaestusPlugin : Plugin<Project> {
...
+
+    if (isAndroidProject) {
+      project.androidVariantsConfigure { variant ->
+        val compileTaskName = "compile${variant.name.capitalize(US)}Kotlin"
+        disableIncrementalCompilationAction(compileTaskName)
+      }
+    } else {
+      // The Java plugin has two Kotlin tasks we care about: compileKotlin and compileTestKotlin.
+      val taskNames = listOf("compileKotlin", "compileTestKotlin")
+      taskNames.forEach(disableIncrementalCompilationAction)
```
cdeange, line 165:
```
disableIncrementalCompilationAction("compileKotlin")
disableIncrementalCompilationAction("compileTestKotlin")
```

Takes the same number of lines, without the need to allocate a list 😃

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
```diff
@@ -22,26 +24,44 @@ open class HephaestusPlugin : Plugin<Project> {
...
+      realApply(project, false)
+    }
+
     project.afterEvaluate {
       if (!once.get()) {
-        throw GradleException("No supported plugins for Hephaestus found on project " +
-            "'${project.name}'. Only Android and Java modules are supported for now.")
+        throw GradleException(
+            "No supported plugins for Hephaestus found on project " +
+                "'${project.name}'. Only Android and Java modules are supported for now."
```
cdeange, line 53: `project.path`? `.name` is not unique within the project hierarchy.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
zach-klippenstein: I prefer the intermediate variables. It makes it more clear that the strings in the list represent task names.

vRallev: Same here in this case.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: I personally like it this way ¯\_(ツ)_/¯

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: I honestly have no idea. I would hope that this would have only a tiny impact and not be measurable at all. I'll change it, because it somehow sounds safer.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: Initially the forEach didn't look this way. Will change.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: Ah, yes!

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: I don't know who would want to do that. But at the same time I don't want to prevent that use case. Maybe in a future version the Kotlin plugin itself applies the plugins? I'd rather keep it as is.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: Done.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
vRallev: This should be `info` instead of `lifecycle`, though.

gradle-plugin/src/main/java/com/squareup/hephaestus/plugin/HephaestusPlugin.kt
cdeange: Sounds like you wanted to avoid that case, since you say above that applying both only happens "accidentally". I guess the behaviour here is undefined if that happens, depending on the ordering, hence why I brought it up.

PR #21, opened 2020-07-04
Use @OptIn(ExperimentalStdlibApi::class) instead of @ExperimentalStdlibApi

vRallev:
Use @OptIn(ExperimentalStdlibApi::class) instead of @ExperimentalStdlibApi to avoid for consumers of the Gradle plugin to add a flag to the Kotlin compiler.

### DIFF
--- a/sample/library/build.gradle
+++ b/sample/library/build.gradle
@@ -10,7 +10,8 @@ dependencies {
   kapt deps.dagger2.compiler
 }
 
-compileKotlin {
+//noinspection UnnecessaryQualifiedReference
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
   kotlinOptions {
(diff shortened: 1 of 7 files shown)


sample/library/build.gradle
```diff
@@ -10,7 +10,8 @@ dependencies {
   kapt deps.dagger2.compiler
 }
 
-compileKotlin {
+//noinspection UnnecessaryQualifiedReference
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
```
cdeange, line 14: Missing a `jvmTarget = JavaVersion.VERSION_1_8` inside this block?

vRallev: I do this in the root project for all subprojects. So it's not necessary here.
```
  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
    kotlinOptions.jvmTarget = JavaVersion.VERSION_1_8
  }
```

PR #22, opened 2020-07-04
Support Dagger modules that are interfaces

vRallev:


compiler/src/test/java/com/squareup/hephaestus/compiler/InterfaceMergerTest.kt
```diff
@@ -417,6 +417,31 @@ class InterfaceMergerTest(
...
+        @ContributesTo(Any::class)
+        interface ContributingInterface
+        
+        @dagger.Module
+        @ContributesTo(Any::class)
+        interface SecondContributingInterface
+        
+        $annotation(Any::class)
+        interface ComponentInterface
+    """
```
zach-klippenstein, line 438: Nit: misaligned triple-quotes

vRallev: I keep it as is for now, because this needs to be fixed in the whole file.

PR #23, opened 2020-07-04
Run a build in CI with a Kotlin dev build. 

vRallev:
These CI runs are optional and don't prevent a merge.


vRallev:
Every Kotlin version starting with `1.4` is optional: https://github.com/square/hephaestus/blob/2c2c943b97f0b9fd51fae166b5f41aed38d594a4/.github/workflows/ci.yml#L13

PR #24, opened 2020-07-04
Windows support

vRallev:


.github/workflows/ci.yml
```diff
@@ -29,8 +30,19 @@ jobs:
         with:
           java-version: 1.8
 
-      - name: ${{ matrix.gradle-task }} with Gradle
+      # On Windows the command looks a little bit different. Notice that we use the .bat file and
+      # quotes for the Kotlin version, because dots "." in the Kotlin version and parameter name
+      # cause issues.
+      #
+      # Expressions in Github actions are limited. If there would be an if expression, then we
+      # wouldn't need to duplicate the next step and depending on the OS enable / disable them.
```
vRallev, line 38: I wonder if there is a better way.

ursusursus: you can just do `gradlew foo` on windows, but yes, still needs to be separate command

PR #33, opened 2020-07-17
Disallow including and excluding types at the same time.

vRallev:
Ensure that types cannot be included and excluded at the same time, which leads to unexpected results.


compiler/src/main/java/com/squareup/hephaestus/compiler/InterfaceMerger.kt
```diff
@@ -107,4 +124,12 @@ internal class InterfaceMerger(
     supertypes += contributedClasses
     super.addSyntheticSupertypes(thisDescriptor, supertypes)
   }
+
+  private fun List<KotlinType>.getAllSuperTypes(): List<FqName> =
+    generateSequence(this) { kotlinTypes ->
+      if (kotlinTypes.isEmpty()) null else kotlinTypes.flatMap { it.supertypes() }
```
FrostRocket, line 130: Before you merge, check if you should return emptyList() here instead of null.

vRallev: Yes, that's important and correct. Without returning `null` `generateSequence` would end in an endless loop and never finish.

PR #41, opened 2020-07-27
Build Anvil with the Kotlin 1.4 RC instead of a milestone release.

vRallev:


compiler/build.gradle
```diff
@@ -23,5 +23,10 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
   kotlinOptions {
     jvmTarget = JavaVersion.VERSION_1_8
     freeCompilerArgs += "-Xopt-in=kotlin.RequiresOptIn"
+
+    // TODO: remove if check when upgrading to Kotlin 1.4.
```
tehras, line 27: did you mean "remove *the* check .."?

tehras: Also maybe write an explanation why it's needed?

vRallev: Done

PR #43, opened 2020-07-28
Add the annotation artifact as an `implementation` dependency instead of `api`.

### LINKED ISSUE #40: Should display error when plugin not applied
Hi, since using this I shot my self in the foot bunch of times on this issue

for whatever reason I can include @ContributesTo annotation, and it compiles, however at runtime looking up components via the given annotated componen type fails

And then I noticed I didnt `apply plugin: "com.squareup.anvil"` in the given module. I apply it. Then it works

Unless plugin is applied, @ContributesTo should not be visible. I'm including all my project dependencies via `implementation`, so it could not have leaked as a transitive dependency? I'm not sure how plugins are handled though, or however else does the annotation get on class path for the incriminated module

vRallev:
This prevents misleading results where one could forget to apply the Gradle
plugin. Imagine module B depends on module A. A applies the Gradle plugin, which
adds an `api` dependency for the annotation. In module B you can reference
the annotations due the transitive dependency, but the compiler plugin won't
run, because the Gradle plugin wasn't applied in this module. By using an
`implementation` dependency you cannot reference the annotations in module B
without applying the Gradle plugin.

Fixes #40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Next Version
 
+* Add the annotation artifact as an `implementation` dependency instead of `api` #40.
+
 ## 1.0.5-1.4-M3 (2020-07-24)
 
 * Renamed the project from Hephaestus to Anvil #12. **IMPORTANT:** Anvil is not compatible with Hephaestus and you must upgrade the plugin in all of your libraries. The artifact coordinates changed, too.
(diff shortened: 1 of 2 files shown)


vRallev:
> Ooh I like this strategy of maintaining the change-log with the change commit itself.
Unfortunately that's hard to do with external contributions.

PR #44, opened 2020-07-28
Allow the plugin to be used in pure JVM modules

### LINKED ISSUE #31: Fails when using gradle worker api
Using the Gradle worker api, each worker gets it's own classpath, so trying to run Hephaestus fails with the following error message.

```
Cannot set the value of task ':some:module:compileReleaseKotlinCheckIncrementalCompilationHephaestus' property 'incrementalSignal' of type com.squareup.hephaestus.plugin.IncrementalSignal using a provider of type com.squareup.hephaestus.plugin.IncrementalSignal.
```

This is in a project with kapt using the worker API, and Room.

I've attempted to implement this in my [branch here](https://github.com/rharter/hephaestus/tree/rh/worker_registry), but need to take off for the day.

### LINKED ISSUE #39: Gradle plugin 1.0.5-1.4-M3 throws NoSuchMethodError on non-Android Kotlin project
I haven't looked too hard, but I think any AGP dependencies in AnvilPlugin need to be done in an optionally configured way?

I think this is the related file:
https://github.com/square/anvil/blob/3b58a5758b6d1fc56b388ff82980f435e39114b7/gradle-plugin/src/main/java/com/squareup/anvil/plugin/IncrementalSignal.kt

```
java.lang.NoSuchMethodError: 'java.io.Closeable com.android.build.gradle.internal.workeractions.WorkerActionServiceRegistry.registerServiceAsCloseable(com.android.build.gradle.internal.workeractions.WorkerActionServiceRegistry$ServiceKey, java.lang.Object)'
        at com.squareup.anvil.plugin.IncrementalSignalBuildService.registerIncrementalSignalService(IncrementalSignal.kt:69)
        at com.squareup.anvil.plugin.IncrementalSignalBuildService.registerIncrementalSignalService$default(IncrementalSignal.kt:64)
        at com.squareup.anvil.plugin.IncrementalSignalKt.registerIncrementalSignalBuildService(IncrementalSignal.kt:27)
        at com.squareup.anvil.plugin.AnvilPlugin.disableIncrementalKotlinCompilation(AnvilPlugin.kt:119)
        at com.squareup.anvil.plugin.AnvilPlugin.realApply(AnvilPlugin.kt:64)
        at com.squareup.anvil.plugin.AnvilPlugin.access$realApply(AnvilPlugin.kt:25)
        at com.squareup.anvil.plugin.AnvilPlugin$apply$3.invoke(AnvilPlugin.kt:47)
        at com.squareup.anvil.plugin.AnvilPlugin$apply$3.invoke(AnvilPlugin.kt:25)
        at com.squareup.anvil.plugin.AnvilPlugin$apply$1$1.execute(AnvilPlugin.kt:35)
        at com.squareup.anvil.plugin.AnvilPlugin$apply$1$1.execute(AnvilPlugin.kt:25)
```

vRallev:
This PR reverts #35, which accidentally added a hard dependency on the Android Gradle Plugin. It fixes #31 in a different way and resolves #39 due to the revert.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/IncrementalSignal.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/IncrementalSignal.kt
@@ -1,9 +1,42 @@
 package com.squareup.anvil.plugin
 
+import org.gradle.api.Project
+import org.gradle.api.provider.Provider
 import org.gradle.api.services.BuildService
 import org.gradle.api.services.BuildServiceParameters.None
+import kotlin.random.Random
+
+/*
+ * The key is important. Gradle uses different class loaders for different modules when the plugin
+ * is applied through the new plugins { } syntax. If we'd use the same key, then the
+ * IncrementalSignal class would be different for the same entry and Gradle would throw an error.
+ *
+ * By using a random key per class loader we avoid this issue. Since this service isn't shared
+ * across modules and only within a module it's not causing any problems tasks only use the service
+ * for their project.
+ *
+ * We could also create one service per module, but that seems wasteful. When the Gradle plugin is
+ * added to classpath of the root project, then the key is always the same and all modules share
+ * a single build service and a single cache.
+ */
+private val KEY = "incrementalSignal-" + Random.nextLong()
 
 /** This signal is used to share state between the task above and Kotlin compile tasks. */
+@Suppress("UnstableApiUsage")
 abstract class IncrementalSignal : BuildService<None> {
-  val incremental = mutableMapOf<String, Boolean?>()
+  private val incremental = mutableMapOf<String, Boolean?>()
+
+  @Synchronized fun setIncremental(
+    projectPath: String,
+    incremental: Boolean
+  ) {
+    this.incremental[projectPath] = incremental
+  }
+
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/IncrementalSignal.kt
```diff
@@ -1,9 +1,42 @@
...
+ *
+ * By using a random key per class loader we avoid this issue. Since this service isn't shared
+ * across modules and only within a module it's not causing any problems tasks only use the service
+ * for their project.
+ *
+ * We could also create one service per module, but that seems wasteful. When the Gradle plugin is
+ * added to classpath of the root project, then the key is always the same and all modules share
+ * a single build service and a single cache.
+ */
+private val KEY = "incrementalSignal-" + Random.nextLong()
```
cdeange, line 22: `UUID.randomUUID().toString` instead? It's guaranteed to be universally unique (though in a practical sense, Random is _technically_ just as likely to be unique).

vRallev: Yeah, I have no concerns here, 2^64 possible values should be more than enough. For some reason it felt wrong to import a Java class in this Kotlin project. That's the only reason why I changed it 😅

PR #46, opened 2020-07-30
Make code generation more flexible

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -0,0 +1,97 @@
...
+            "Could not clean file: $it"
+          }
+        }
+
+    fun generateCode(files: Collection<KtFile>): Collection<GeneratedFile> =
+      codeGenerators
+          .map {
+            it.generateCode(codeGenDir, module, files)
+          }
+          .flatten()
```
kirillzh, line 68: Could use `flatMap` here?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -0,0 +1,97 @@
...
+
+    val psiManager = PsiManager.getInstance(project)
+
+    var newFiles = generateCode(files)
+    while (newFiles.isNotEmpty()) {
+      // Parse the KtFile for each generated file. Then feed the code generates
```
FrostRocket, line 74: Looks like you forgot to finish your comment.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -0,0 +1,97 @@
...
+
+internal class CodeGenerationExtension(
+  private val codeGenDir: File
+) : AnalysisHandlerExtension {
+
+  private var didRecompile = false
+
+  private val codeGenerators = mutableListOf<CodeGenerator>()
+
+  fun registerCodeGenerator(codeGenerator: CodeGenerator) {
```
FrostRocket, line 26: If you didn't already know to call this, it'd be easy to forget to call `codeGenerationExtension.registerCodeGenerator(ContributesToGenerator())`. Would it be more beneficial to enforce that via a constructor argument instead?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -0,0 +1,97 @@
...
+      codeGenerators
+          .map {
+            it.generateCode(codeGenDir, module, files)
+          }
+          .flatten()
+
+    val psiManager = PsiManager.getInstance(project)
+
+    var newFiles = generateCode(files)
+    while (newFiles.isNotEmpty()) {
```
FrostRocket, line 73: There's never a scenario where you'd end up looping forever due to the newFiles list never being empty? I'm slightly worried that you'd run into an issue where you'd need to use an iterator instead. Just making sure.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerator.kt
```diff
@@ -0,0 +1,36 @@
+package com.squareup.anvil.compiler.codegen
+
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.psi.KtFile
+import java.io.File
+
+internal interface CodeGenerator {
+  /**
+   * Called multiple times in order to create new code. Note that instances should not rely on
```
FrostRocket, line 9: Great documentation 🥳

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
vRallev: Hm, I like this better. Constructors should be free of any logic.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
vRallev: No, that doesn't happen.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
FrostRocket: I'm asking because I don't see where unregisterCodeGenerator would be used, so getting rid of the internally managed list and instead pass in when you initialize the class. I might be missing something if those are used outside the scope of this PR though.

vRallev: That's fair, I like that actually better.

FrostRocket: There's no logic? You're injecting the list of codeGenerators instead of relying on someone to call that method beforehand. Without it this class won't work properly, right? Your call.

vRallev: Changed

PR #47, opened 2020-07-30
Introduce a helper function that requires an annotation to be present.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -36,7 +36,7 @@ internal val daggerModuleFqName = FqName(Module::class.java.canonicalName)
 
 internal const val HINT_PACKAGE_PREFIX = "hint.anvil"
 
-internal fun ClassDescriptor.findAnnotation(
+internal fun ClassDescriptor.annotationOrNull(
```
vRallev, line 39: That aligns better with the Kotlin naming conventions.

PR #48, opened 2020-07-31
Introduce the @ContributesBinding annotation

vRallev:
This PR only adds the annotation and one functionality in order to make the feature work. I thought it would be easier to review this way.

### DIFF
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
@@ -1,0 +1,59 @@
+package com.squareup.anvil.annotations
+
+import kotlin.annotation.AnnotationRetention.RUNTIME
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.reflect.KClass
+
+/**
+ * Generate a Dagger binding method for an annotated class and contributes this binding method to
+ * the given [scope]. Imagine this example:
+ * ```
+ * interface Authenticator
+ *
+ * class RealAuthenticator @Inject constructor() : Authenticator
+ *
+ * @Module
+ * @ContributesTo(AppScope::class)
+ * abstract class AuthenticatorModule {
+ *   @Binds abstract fun bindRealAuthenticator(authenticator: RealAuthenticator): Authenticator
+ * }
+ * ```
+ * This is a lot of boilerplate if you always want to use `RealAuthenticator` when injecting
+ * `Authenticator`. You can replace this entire module with the [ContributesBinding] annotation.
+ * The equivalent would be:
+ * ```
+ * interface Authenticator
+ *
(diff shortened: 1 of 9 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -30,11 +31,13 @@ internal val mergeSubcomponentFqName = FqName(MergeSubcomponent::class.java.cano
 internal val mergeInterfacesFqName = FqName(MergeInterfaces::class.java.canonicalName)
 internal val mergeModulesFqName = FqName(MergeModules::class.java.canonicalName)
 internal val contributesToFqName = FqName(ContributesTo::class.java.canonicalName)
+internal val contributesBindingFqName = FqName(ContributesBinding::class.java.canonicalName)
 internal val daggerComponentFqName = FqName(Component::class.java.canonicalName)
 internal val daggerSubcomponentFqName = FqName(Subcomponent::class.java.canonicalName)
 internal val daggerModuleFqName = FqName(Module::class.java.canonicalName)
 
-internal const val HINT_PACKAGE_PREFIX = "hint.anvil"
+internal const val HINT_CONTRIBUTES_PACKAGE_PREFIX = "hint.anvil"
```
vRallev, line 39: I wish I could change this package, but it would be a non-backward compatible change if you depend on library that was compiled with an older Anvil version.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -0,0 +1,59 @@
...
+ * )
+ * class FakeAuthenticator @Inject constructor() : Authenticator
+ * ```
+ */
+@Target(CLASS)
+@Retention(RUNTIME)
+annotation class ContributesBinding(
+  val scope: KClass<*>,
+  val boundType: KClass<*> = Unit::class,
+  val replaces: KClass<*> = Unit::class
```
vRallev, line 58: Again, a lot of things for this annotation are still missing and it's not working, yet. I'll be implemented in the next PRs.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -0,0 +1,59 @@
...
+ * }
+ * ```
+ * This is a lot of boilerplate if you always want to use `RealAuthenticator` when injecting
+ * `Authenticator`. You can replace this entire module with the [ContributesBinding] annotation.
+ * The equivalent would be:
+ * ```
+ * interface Authenticator
+ *
+ * @ContributesBinding(AppScope::class)
+ * class RealAuthenticator @Inject constructor() : Authenticator
```
kirillzh, line 28: Exciting!

PR #49, opened 2020-07-31
Generate an Anvil Dagger module with contributed bindings

vRallev:
This PR depends on #48 and actually implements some parts in order to make to `@ContributesBinding` work. The bounded type is still required and replacements and exclusions don't work, yet.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -1,0 +1,90 @@
+package com.squareup.anvil.compiler.codegen
+
+import com.squareup.anvil.compiler.AnvilCompilationException
+import com.squareup.anvil.compiler.MODULE_PACKAGE_PREFIX
+import com.squareup.anvil.compiler.annotationOrNull
+import com.squareup.anvil.compiler.codegen.CodeGenerator.GeneratedFile
+import com.squareup.anvil.compiler.contributesToFqName
+import com.squareup.anvil.compiler.daggerModuleFqName
+import com.squareup.anvil.compiler.mergeComponentFqName
+import com.squareup.anvil.compiler.mergeModulesFqName
+import com.squareup.anvil.compiler.mergeSubcomponentFqName
+import com.squareup.anvil.compiler.scope
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.descriptors.resolveClassByFqName
+import org.jetbrains.kotlin.incremental.KotlinLookupLocation
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtFile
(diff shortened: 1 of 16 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
```diff
@@ -0,0 +1,205 @@
...
+        interface ComponentInterface
+    """
+    ) {
+      val modules = if (annotationClass == MergeModules::class) {
+        componentInterface.daggerModule.includes.toList()
+      } else {
+        componentInterface.anyDaggerComponent.modules
+      }
+      assertThat(modules)
+          .containsExactly(componentInterfaceAnvilModule.kotlin, daggerModule1.kotlin)
```
kirillzh, line 85: Does expected order matter here? If yes, might need to add `.inOrder()`.

vRallev: No, the order doesn't matter for Dagger.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -0,0 +1,236 @@
...
+              "extend this type."
+      )
+    }
+  }
+
+  private fun generateClassName(psiClass: KtClassOrObject): String =
+    psiClass.parentsWithSelf
+        .filterIsInstance<KtClassOrObject>()
+        .toList()
+        .reversed()
```
FrostRocket, line 208: Why are you reversing the list?

vRallev: `com.abc.MyClass.InnerClass.InnerInnerClass`. Calling `parents`walks the structure from right to left. But I want classes from left to right. I want to generate `MyClassInnerClassInnerInnerClassAnvilModule`. Without reversing I'd generate `InnerInnerClassInnerClassMyClassAnvilModule`.

integration-tests/tests/src/test/java/com/squareup/anvil/test/Utils.kt
```diff
@@ -4,3 +4,7 @@ import kotlin.reflect.KClass
 
 internal infix fun KClass<*>.extends(other: KClass<*>): Boolean =
   other.java.isAssignableFrom(this.java)
+
+internal fun Array<KClass<*>>.withoutAnvilModule(): List<KClass<*>> = toList().withoutAnvilModule()
+internal fun Collection<KClass<*>>.withoutAnvilModule(): List<KClass<*>> =
+  filterNot { it.qualifiedName!!.startsWith("anvil.module") }
```
FrostRocket, line 10: MODULE_PACKAGE_PREFIX?

vRallev: Does not work here, because this module doesn't see code from a Kotlin compiler plugin. It's a different classpath.

PR #50, opened 2020-07-31
Make the bound type optional

vRallev:
This PR depends on #49


compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
```diff
@@ -173,6 +173,102 @@ class BindingModuleGeneratorTest(
...
+        """
+        package com.squareup.test
+        
+        import com.squareup.anvil.annotations.ContributesBinding
+        $import
+
+        interface ParentInterface
+
+        @ContributesBinding(Any::class)
+        interface ContributingInterface : ParentInterface
```
vRallev, line 187: Here is an example where the bound type is implied.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -52,13 +54,26 @@ internal class ModuleMerger(
...
         .findContributedClasses(
             module = codegen.descriptor.module,
             packageName = HINT_CONTRIBUTES_PACKAGE_PREFIX,
             annotation = contributesToFqName
         )
         .asSequence()
+        .filter {
+          // We generate a Dagger module for each merged component. We use Anvil itself to
+          // contribute this generated module. It could happen that there are multiple components
+          // merging the same scope or different Gradle modules depend on each other and in each
```
kirillzh, line 69: "depend" -> "depending"

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -52,13 +54,26 @@ internal class ModuleMerger(
...
+          // the same scope is merged. This would cause duplicate bindings, because the generated
```
kirillzh, line 70: "and in each the same scope is merged" reads a bit strange, possible to rephrase?

kirillzh, line 70: "duplicate" -> "duplicated"?

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -52,13 +54,26 @@ internal class ModuleMerger(
...
     val modules = classScanner
         .findContributedClasses(
             module = codegen.descriptor.module,
             packageName = HINT_CONTRIBUTES_PACKAGE_PREFIX,
             annotation = contributesToFqName
         )
         .asSequence()
+        .filter {
+          // We generate a Dagger module for each merged component. We use Anvil itself to
+          // contribute this generated module. It could happen that there are multiple components
```
FrostRocket, line 68: "It's possible" vs "It could happen". Flows a little better.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
FrostRocket: It reads fine as-is to me.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -173,6 +188,20 @@ internal class ModuleMerger(
     block(this)
     visitEnd()
   }
+
+  private fun createAnvilModuleName(classDescriptor: ClassDescriptor): FqName {
+    val name = "$MODULE_PACKAGE_PREFIX.${classDescriptor.findPackage().fqName}." +
+        classDescriptor.parentsWithSelf
+            .filterIsInstance<ClassDescriptor>()
+            .toList()
+            .reversed()
```
FrostRocket, line 197: I saw this in your last PR, but still curious why you need to reverse this.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -87,6 +91,11 @@ internal fun ConstantValue<*>.toType(
 // When the Kotlin type is of the form: KClass<OurType>.
 internal fun KotlinType.argumentType(): KotlinType = arguments.first().type
 
+internal fun FqName.isAnvilModule(): Boolean {
+  val string = asString()
```
FrostRocket, line 95: "name" instead of "string"? I know it's a bit obvious here but if you're going to declare an intermediate variable might as well name it better 😄

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
vRallev: Explained it there.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -98,10 +107,29 @@ internal fun AnnotationDescriptor.scope(module: ModuleDescriptor): ClassDescript
...
+    throw AnvilCompilationException(
+        classDescriptor = annotatedClass,
+        message = "${annotatedClass.fqNameSafe} contributes a binding, but does not " +
+            "specify the bound type. This is only allowed with one direct super type. If " +
+            "there are multiple, then the bound type must be explicitly defined in the " +
+            "@${contributesBindingFqName.shortName()} annotation."
+    )
+  }
+
+  return directSuperTypes[0]
```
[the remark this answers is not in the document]
FrostRocket: I like this suggestion, would make it cleaner and a bit less verbose.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
```diff
@@ -173,6 +173,102 @@ class BindingModuleGeneratorTest(
...
+        
+        $annotation(Any::class)
+        interface ComponentInterface
+    """
+    ) {
+      assertThat(exitCode).isEqualTo(COMPILATION_ERROR)
+
+      assertThat(messages).contains("Source.kt: (11, 11)")
+      assertThat(messages).contains(
+          "com.squareup.test.ContributingInterface contributes a binding, but does not specify " +
```
FrostRocket, line 264: I may have mentioned this before, but all these tests are going to fail on you (and be annoying) if you end up refactoring the wording of your logging strings. I wonder if you could leverage something Android-esque like strings.xml to have a centralized spot for them.

Not a ship stopper just something to think about moving forward.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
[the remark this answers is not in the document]
vRallev: Definitely agree.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
vRallev: In older tests I didn't check the error message at all and only the position. But I found those tests actually harder to maintain and read. A centralized spot could help, but is also hard to work with due to the placeholders. For now I think that's the best option knowing that I will change my mind in a few days :-)

PR #54, opened 2020-08-03
Change the `replaces` attribute from a single class to an array. 

vRallev:
This gives the API more flexibility and avoids redundant classes. E.g. one Dagger module with several binding and provider methods may wish to replace multiple other Dagger modules.


vRallev:
That's a breaking API change and the next version will be released as `1.1.0`.

PR #62, opened 2020-08-14
Remove the usage of the deprecated KotlinGradleSubplugin

### LINKED ISSUE #30: Remove KotlinGradleSubplugin with Kotlin 1.4
The class `KotlinGradleSubplugin` is deprecated in favor of `KotlinCompilerPluginSupportPlugin`. When Kotlin 1.4 is out this project is migrated, then `HephaestusSubplugin` should be removed and the new API be used instead.

vRallev:
Use the new replacement instead.

Fixes #30


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -8,21 +8,58 @@ import com.android.build.gradle.TestExtension
...
+
+  override fun applyToCompilation(
+    kotlinCompilation: KotlinCompilation<*>
+  ): Provider<List<SubpluginOption>> {
+    val project = kotlinCompilation.target.project
+
+    // Notice that we use the name of the Kotlin compile task as a directory name. Generated code
+    // for this specific compile task will be included in the task output. The output of different
+    // compile tasks shouldn't be mixed.
+    val srcGenDir = File(project.buildDir, "anvil/src-gen-${kotlinCompilation.name}")
```
cdeange, line 50: Should be using `File.separator` if you want this to be portable for windows users.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -8,21 +8,58 @@ import com.android.build.gradle.TestExtension
...
+import java.io.File
 import java.util.Locale.US
 import java.util.concurrent.atomic.AtomicBoolean
 
-open class AnvilPlugin : Plugin<Project> {
-  override fun apply(project: Project) {
+open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
+
+  override fun isApplicable(kotlinCompilation: KotlinCompilation<*>): Boolean {
+    return kotlinCompilation.target.project.plugins.hasPlugin(AnvilPlugin::class.java)
```
stephanenicolas, line 31: I was also surprised that the plugin is applied to all our projects. How does that work ? Why do we have to check this? Isn't the compiler plugin only applied to projects using the plugin ?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -39,18 +76,18 @@ open class AnvilPlugin : Plugin<Project> {
...
-      realApply(project, false)
+    target.pluginManager.withPluginOnce("org.jetbrains.kotlin.jvm") {
+      realApply(target, false)
     }
 
-    project.afterEvaluate {
+    target.afterEvaluate {
       if (!once.get()) {
         throw GradleException(
             "No supported plugins for Anvil found on project " +
```
stephanenicolas, line 89: The message is not very clear (gradle plugin, compiler plugin, how does a plugin requires plugins ?). 
-> The Anvil gradle plugin could not be applied to project...

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -8,21 +8,58 @@ import com.android.build.gradle.TestExtension
...
+      artifactId = "compiler",
+      version = VERSION
+  )
+
+  override fun applyToCompilation(
+    kotlinCompilation: KotlinCompilation<*>
+  ): Provider<List<SubpluginOption>> {
+    val project = kotlinCompilation.target.project
+
+    // Notice that we use the name of the Kotlin compile task as a directory name. Generated code
```
stephanenicolas, line 47: Maybe in this comment we should remove the mentions to the compile task and talk about compilation instead.
The whole reason for the new interface is to avoid referencing tasks and break Task Configuration Avoidance.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: It's a sanity check. You can still apply the Gradle plugin to a module and the compiler plugin to single Kotlin compilation tasks. I our case we apply it to all.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: Makes sense, done.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: That was resolved in another PR.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: Unrelated to this PR.

kirillzh (approved):

PR #66, opened 2020-08-24
Generate Dagger Factories

vRallev:
The code that Dagger generates for modules, field injection and inject constructors is trivial. The real magic happens when Dagger compiles components or subcomponents.

This PR gives us the option to use Anvil instead of the Dagger annotation processor to generate the factories that Dagger would generate. This improves build times for many modules. 

These numbers are from our code base when building one of the tiniest development apps:
  | Stub generation | Kapt | Javac | Kotlinc | Sum
--- | --- | --- | --- | --- | ---
With Dagger | 12.976 | 40.377 | 8.571 | 10.241 | 72.165
Anvil | 0 | 0 | 6.965 | 17.748 | 24.713

Notice that many tasks don't exist anymore with this optimization. Since no annotation processor needs to run Kapt stub generation and Kapt itself are gone. Anvil generates Kotlin code instead of Java unlike the Dagger annotation processor. In many modules javac is a no-op now, because there aren't any Java sources anymore. Koltinc on the other side is slower due to applying the Kotlin compiler plugin itself and that's where the work happens.

In CI for SPOS I've measured 9:37 with Dagger vs 7:57 with Anvil.

This optimization can only be enabled in modules that don't compile a Dagger component. If Anvil sees a component or subcomponent, then it throws a compilation error. This new feature can be enabled through a new Gradle extension:
```groovy
anvil {
  generateDaggerFactories = true // default is 'false'
}
```

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,11 +38,11 @@ jobs:
       # wouldn't need to duplicate the next step and depending on the OS enable / disable them.
       - name: ${{ matrix.gradle-task }} with Gradle Ubuntu
         run: ./gradlew ${{ matrix.gradle-task }} --no-build-cache --no-daemon --stacktrace -Psquare.kotlinVersion=${{ matrix.kotlin-version }}
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
+        if: ${{ startsWith(matrix.os, 'ubuntu') && (matrix.gradle-task != 'test' || matrix.kotlin-version == '1.3.72') }}
 
       - name: ${{ matrix.gradle-task }} with Gradle Windows
         run: ./gradlew.bat ${{ matrix.gradle-task }} --no-build-cache --no-daemon --stacktrace "-Psquare.kotlinVersion=${{ matrix.kotlin-version }}"
-        if: ${{ startsWith(matrix.os, 'windows') }}
+        if: ${{ startsWith(matrix.os, 'windows')  && (matrix.gradle-task != 'test' || matrix.kotlin-version == '1.3.72') }}
 
       - name: Upload Lint Results
         uses: actions/upload-artifact@v2
--- a/compiler/build.gradle
+++ b/compiler/build.gradle
@@ -7,12 +7,14 @@ dependencies {
   implementation deps.dagger2.dagger
   implementation deps.jsr250
   implementation deps.kotlin.stdlib
+  implementation deps.kotlinpoet
 
   compileOnly deps.auto.service.annotations
   compileOnly deps.kotlin.compiler
 
   kapt deps.auto.service.processor
 
+  testImplementation deps.dagger2.compiler
   testImplementation deps.kotlin.compile_testing
   testImplementation deps.kotlin.compiler
   testImplementation deps.truth
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
@@ -1,20 +1,33 @@
 package com.squareup.anvil.compiler.codegen
 
 import com.squareup.anvil.compiler.AnvilCompilationException
+import com.squareup.anvil.compiler.jvmSuppressWildcardsFqName
+import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.descriptors.resolveClassByFqName
 import org.jetbrains.kotlin.incremental.components.NoLookupLocation.FROM_BACKEND
 import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtClass
+import org.jetbrains.kotlin.psi.KtClassBody
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
+import org.jetbrains.kotlin.psi.KtNamedDeclaration
+import org.jetbrains.kotlin.psi.KtNamedFunction
+import org.jetbrains.kotlin.psi.KtNullableType
+import org.jetbrains.kotlin.psi.KtTypeArgumentList
+import org.jetbrains.kotlin.psi.KtTypeReference
+import org.jetbrains.kotlin.psi.KtUserType
 import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.KtValueArgumentName
+import org.jetbrains.kotlin.psi.psiUtil.parentsWithSelf
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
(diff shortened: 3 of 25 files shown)


.github/workflows/ci.yml
```diff
@@ -38,11 +38,11 @@ jobs:
       # wouldn't need to duplicate the next step and depending on the OS enable / disable them.
       - name: ${{ matrix.gradle-task }} with Gradle Ubuntu
         run: ./gradlew ${{ matrix.gradle-task }} --no-build-cache --no-daemon --stacktrace -Psquare.kotlinVersion=${{ matrix.kotlin-version }}
-        if: ${{ startsWith(matrix.os, 'ubuntu') }}
+        if: ${{ startsWith(matrix.os, 'ubuntu') && (matrix.gradle-task != 'test' || matrix.kotlin-version == '1.3.72') }}
 
       - name: ${{ matrix.gradle-task }} with Gradle Windows
         run: ./gradlew.bat ${{ matrix.gradle-task }} --no-build-cache --no-daemon --stacktrace "-Psquare.kotlinVersion=${{ matrix.kotlin-version }}"
-        if: ${{ startsWith(matrix.os, 'windows') }}
+        if: ${{ startsWith(matrix.os, 'windows')  && (matrix.gradle-task != 'test' || matrix.kotlin-version == '1.3.72') }}
```
vRallev, line 45: The interesting bit here is that the tests with Dagger fail with Kotlin 1.4. But there is progress: https://github.com/tschuchortdev/kotlin-compile-testing/issues/51

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ComponentDetectorCheck.kt
```diff
@@ -0,0 +1,41 @@
...
+              daggerSubcomponentFqName
+          )
+        }
+
+    if (component != null) {
+      throw AnvilCompilationException(
+          message = "Anvil cannot generate the code for Dagger components or subcomponents. In " +
+              "these cases the Dagger annotation processor is required. Enabling the Dagger " +
+              "annotation processor and turning on Anvil to generate Dagger factories is " +
+              "redundant. Set 'generateDaggerFactories' to false.",
```
vRallev, line 34: This check helped me to find the modules in our codebase that cannot make use of this feature.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
```diff
@@ -0,0 +1,608 @@
...
+@Suppress("UNCHECKED_CAST")
+@RunWith(Parameterized::class)
+class InjectConstructorFactoryGeneratorTest(
+  private val useDagger: Boolean
+) {
+
+  companion object {
+    @Parameters(name = "Use Dagger: {0}")
+    @JvmStatic fun useDagger(): Collection<Any> {
+      return listOf(true, false)
```
vRallev, line 26: I run all tests for each code generator one time with Dagger and one time with Anvil. In both cases the generated code must pass the expectations. This way I make sure that the Anvil generates what Dagger expects, even though Anvil generates Kotlin code. Another safety net is that Dagger references the generated code when compiling components. If something doesn't match, then there is a compile error. When testing this feature in our codebase I found smaller edge cases at compile time, but at runtime there never was an issue.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -159,3 +153,94 @@ private fun KtClassOrObject.findScopeClassLiteralExpression(
...
+      element = this
+  )
+
+  val classReference = when (this) {
+    // If a fully qualified name is used, then we're done and don't need to do anything further.
+    is KtDotQualifiedExpression -> return FqName(text)
+    is KtNameReferenceExpression -> getReferencedName()
+    is KtUserType -> referencedName ?: failTypeHandling()
+    is KtTypeReference -> text
+    is KtNullableType -> return innerType?.requireFqName(module) ?: failTypeHandling()
```
kirillzh, line 188: Should we be using `fqNameOrNull()` in case if type is present to get better error message from `failTypeHandling`?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/KotlinPoet.kt
```diff
@@ -0,0 +1,230 @@
...
+  val importDirectives = copiedImports
+      .plus(
+          originalLines
+              .filter {
+                it.startsWith("import ") && it.contains(".")
+              }
+              .map { it.trim() }
+      )
+      .distinct()
+      .toMutableList()
```
kirillzh, line 200: Do we care about order at all here?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
vRallev: That's not necessary here. If `innerType` is present, then it must have a fqName.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/KotlinPoet.kt
vRallev: The order is important for the build cache. In theory it shouldn't matter, because it doesn't impact the ABI, but with the Kotlin plugin you never know...

I think the order is stable. But to be sure the imports should be sorted before they're written to the file.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ComponentDetectorCheckTest.kt
```diff
@@ -0,0 +1,114 @@
...
+  @Test fun `a Dagger component causes an error`() {
+    compile(
+        """
+        package com.squareup.test
+        
+        import dagger.Component
+        
+        @Component
+        interface ComponentInterface
+    """
```
zach-klippenstein, line 19: Nit: different start/end indentation.

compiler/build.gradle
```diff
@@ -7,12 +7,14 @@ dependencies {
   implementation deps.dagger2.dagger
   implementation deps.jsr250
   implementation deps.kotlin.stdlib
+  implementation deps.kotlinpoet
```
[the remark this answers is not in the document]
vRallev: No, it needs to be source for the compiler itself. I'm generating source code and tell the compiler to load the generated code on top. The compiler then compiles this code as any other. There are no hooks for generating bytecode.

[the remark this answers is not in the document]
zach-klippenstein: TIL

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ComponentDetectorCheckTest.kt
vRallev: Will fix that everywhere in a follow-up PR.

vRallev: https://github.com/square/anvil/pull/69

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ComponentDetectorCheck.kt
Dkhusainov: What about modules(gradle modules) which declare only `Subcomponent`. Dagger does not generate anything(maybe I'm wrong) for them, so why prohibit subcomponents?

vRallev: Actually I missed that. You're right. I filed #74.

PR #67, opened 2020-08-24
Throw a compilation error when the contributed binding has a type argument.

vlad-kasatkin:
This is an example of a (more complicated) error for `ParentInterface<Map<String, List<Pair<String, Int>>>, SomeOtherType>`

>  com.squareup.anvil.compiler.AnvilCompilationException: Back-end (JVM) Internal error: Binding com.squareup.test.ParentInterface contains type parameters(s) <Map<String, List<Pair<String, Int>>>, SomeOtherType>. Type parameters in bindings are not supported. This binding needs to be contributed to a dagger module manually

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -34,7 +34,6 @@ import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.psiUtil.parentsWithSelf
-import org.jetbrains.kotlin.resolve.DescriptorUtils
 import org.jetbrains.kotlin.resolve.constants.ArrayValue
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
 import org.jetbrains.kotlin.resolve.descriptorUtil.getAllSuperClassifiers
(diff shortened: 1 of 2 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
```diff
@@ -339,6 +339,36 @@ class BindingModuleGeneratorTest(
...
+        
+        import com.squareup.anvil.annotations.ContributesBinding
+        $import
+
+        interface ParentInterface<T, S>
+        
+        class SomeOtherType
+        
+        @ContributesBinding(Any::class, ParentInterface::class)
+        interface ContributingInterface:
```
vRallev, line 355: Nit: missing space before the colon.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -224,6 +227,46 @@ internal class BindingModuleGenerator(
...
+            }
+            append(DescriptorUtils.getClassDescriptorForType(typeArgument.type).name)
+            describeTypeParameters(typeArgument.type)
+          }
+
+          append(">")
+        }
+
+        return this
+      }
```
vRallev, line 252: This entire function could be shrinked to:
```kotlin
      fun KotlinType.describeTypeParameters(): String =
        arguments
            .ifEmpty { return "" }
            .joinToString(prefix = "<", postfix = ">") { typeArgument ->
              typeArgument.type.classDescriptorForType().name.asString() +
                  typeArgument.type.describeTypeParameters()
            }
```
You're using a `StringBuilder`, but I don't think this optimization is worth it in an error scenario. I'd prefer better readability in this case.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -224,6 +227,46 @@ internal class BindingModuleGenerator(
...
+
+      val actualBoundType = type.typeConstructor
+          .supertypes
+          .first { DescriptorUtils.getClassDescriptorForType(it) == boundTypeDescriptor }
```
vRallev, line 256: Use the extension function: `it.classDescriptorForType()`

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vlad-kasatkin: Very fancy! Love it. Thanks!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vlad-kasatkin: Ah, damn. It was already there and used in the very file... Thanks!

PR #68, opened 2020-08-25
Override the Kotlin compilation task inputs

### LINKED ISSUE #65: Usage of absolute path as input

vRallev:
Override the Kotlin compilation task inputs to avoid build cache misses on different machines.

Fixes #65

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
@@ -38,13 +38,24 @@ class AnvilSubplugin : KotlinGradleSubplugin<AbstractCompile> {
     // Notice that we use the name of the Kotlin compile task as a directory name. Generated code
     // for this specific compile task will be included in the task output. The output of different
     // compile tasks shouldn't be mixed.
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
```diff
@@ -39,12 +39,23 @@ class AnvilSubplugin : KotlinGradleSubplugin<AbstractCompile> {
     // for this specific compile task will be included in the task output. The output of different
     // compile tasks shouldn't be mixed.
     val srcGenDir = File(project.buildDir, "anvil/src-gen-${kotlinCompile.name}")
```
cdeange, line 41: nit: use `File.separator` for platform-agnostic file-path building.

vRallev: You or someone else pointed this out already and then I forgot to fix it. This is a good PR to fix it.

cdeange: It was me 😄

vRallev: I feel bad

PR #78, opened 2020-08-29
Support different JVM targets when adding the @Generated annotation.

### LINKED ISSUE #75: generateDaggerFactories doesn't seem to work on windows (?)
https://github.com/ursusursus/AnvilRepro

:base contains the flag, and has removed the dagger compiler, and kapt altogether
dagger compiler and kapt is still present in :app

I think this is the correct usage, and therefore a bug in the library

```
> Task :base:compileDebugKotlin FAILED
e: C:\Users\ursus\AndroidStudioProjects\MyApplication9\base\build\anvil\src-gen-compileDebugKotlin\sk\foo\base\di\BaseModule_BarFactory.kt: (8, 14): Unresolved reference: annotation
e: C:\Users\ursus\AndroidStudioProjects\MyApplication9\base\build\anvil\src-gen-compileDebugKotlin\sk\foo\base\di\BaseModule_BarFactory.kt: (16, 2): Unresolved reference: Generated
```

In my real project I kept getting

```
error: incompatible types: NonExistentClass cannot be converted to Annotation` 
```
on a similar setup

vRallev:
Fixes #75


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/KotlinPoet.kt
```diff
@@ -173,15 +177,41 @@ private fun TypeName.wrapInLazy(): ParameterizedTypeName {
...
+private fun ModuleDescriptor.getGeneratedAnnotationClassName(): String? {
+  val targetVersion = (module.platform
+      ?.firstOrNull { it is JdkPlatform } as? JdkPlatform)
+      ?.targetVersion
+      ?: return null
+
+  return when {
+    targetVersion > JVM_1_8 -> "javax.annotation.processing.Generated"
+    targetVersion == JVM_1_8 -> "javax.annotation.Generated"
+    else -> null
```
gabrielittner, line 189: Will this return `null` for Android modules that target 1.8? On Android the annotation is never available. Dagger does not emit it in Android projects at all.

vRallev: I'd assume it wouldn't. Do you know how Dagger makes this distinction by any chance?

gabrielittner: It's checking the classpath https://github.com/google/auto/blob/master/common/src/main/java/com/google/auto/common/GeneratedAnnotations.java

vRallev: I definitely can reproduce what you said.

vRallev: Resolved.

gabrielittner: Great, thank you

vRallev:
I changed my mind. The annotation is annoying. I don't wanna resolve it during compilation. It adds little value and the comment is good enough for me. I removed the annotation in favor of a comment.

PR #88, opened 2020-08-31
Support @ContributessBindings for objects

vRallev:


vRallev:
Note that the first commit is adding the functionality. The remaining two commits are cleanup.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -49,6 +49,13 @@ import kotlin.reflect.KClass
  * )
  * class FakeAuthenticator @Inject constructor() : Authenticator
  * ```
+ * [ContributesBinding] supports Kotlin objects, e.g.
+ * ```
+ * @ContributesBinding(AppScope::class)
+ * object RealAuthenticator : Authenticator
+ * ```
+ * In this scenario instead of generating a `@Binds` method Anvil will generate a `@Provides`
+ * method returning [boundType].
```
zach-klippenstein, line 58: Might be good to have a bit more documentation about how the bound type is determined? The default value is `Unit::class`, but this example suggests it would be inferred to be `Authenticator`. Is that true, or is the example wrong? And if it is true, what happens if `RealAuthenticator` implements more than one interface? If `Authenticator` itself extends another interface? This is clear in the error message below, but would be nice to call it out here as well.

vRallev: That's what I'm explaining above. It's more that `object`s are an exceptional case that I call out. All the constraints you mentioned are explained above for the general case.

zach-klippenstein: Ah, ok.

Issue #76, opened 2020-08-28
Generated code with Anvil is larger than Dagger generated code


vRallev:
This output is from the integration test library. I need to investigate where and how we can safe. There are a couple of ideas.

<details><summary>Report</summary>

<p>

```
ralf$ diffuse diff --jar old.jar new.jar
OLD: old.jar
NEW: new.jar

 JAR   │ old      │ new      │ diff
───────┼──────────┼──────────┼──────────
 class │ 27.4 KiB │ 32.1 KiB │ +4.7 KiB
 other │    468 B │    468 B │      0 B
───────┼──────────┼──────────┼──────────
 total │ 27.9 KiB │ 32.6 KiB │ +4.7 KiB


 CLASSES │ old │ new │ diff
─────────┼─────┼─────┼──────────────
 classes │  27 │  27 │   0 (+3 -3)
 methods │  68 │  78 │ +10 (+19 -9)
  fields │  21 │  24 │  +3 (+6 -3)


=================
====   JAR   ====
=================

 size     │ diff     │ path
──────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────
  2.2 KiB │ +2.2 KiB │ + com/squareup/anvil/test/AppModule2_ProvideFunctionFactory$Companion.class
  1.5 KiB │ +1.5 KiB │ + com/squareup/anvil/test/SubcomponentBinding_Factory$Companion.class
  1.4 KiB │ +1.4 KiB │ + com/squareup/anvil/test/AppBinding_Factory$Companion.class
          │   -788 B │ - com/squareup/anvil/test/AppModule2_ProvideFunctionFactory$InstanceHolder.class
          │   -752 B │ - com/squareup/anvil/test/SubcomponentBinding_Factory$InstanceHolder.class
          │   -698 B │ - com/squareup/anvil/test/AppBinding_Factory$InstanceHolder.class
  1.6 KiB │   +664 B │ ∆ com/squareup/anvil/test/AppBinding_Factory.class
  1.7 KiB │   +673 B │ ∆ com/squareup/anvil/test/SubcomponentBinding_Factory.class
  1.9 KiB │   +473 B │ ∆ com/squareup/anvil/test/AppModule2_ProvideFunctionFactory.class
──────────┼──────────┼──────────────────────────────────────────────────────────────────────────────────
 10.4 KiB │ +4.7 KiB │ (total)



=====================
====   CLASSES   ====
=====================

CLASSES:

   old │ new │ diff
[...]

PR #89, opened 2020-08-31
Generate an object implementing the Factory interface

vRallev:
Generate an object implementing the Factory interface rather than a class with a companion object when a provider method doesn't have any parameters and the Dagger module is stateless.

This reduces the bytecode size #76.

Old report:
```
 JAR   │ old      │ new      │ diff
───────┼──────────┼──────────┼──────────
 class │ 29.3 KiB │ 36.8 KiB │ +7.5 KiB
 other │    476 B │    476 B │      0 B
───────┼──────────┼──────────┼──────────
 total │ 29.7 KiB │ 37.2 KiB │ +7.5 KiB


 CLASSES │ old │ new │ diff
─────────┼─────┼─────┼──────────────
 classes │  28 │  29 │  +1 (+4 -3)
 methods │  74 │  89 │ +15 (+24 -9)
  fields │  23 │  27 │  +4 (+7 -3)
```

New report:
```
 JAR   │ old      │ new      │ diff
───────┼──────────┼──────────┼──────────
 class │ 29.3 KiB │ 34.8 KiB │ +5.5 KiB
 other │    476 B │    476 B │      0 B
───────┼──────────┼──────────┼──────────
 total │ 29.7 KiB │ 35.3 KiB │ +5.5 KiB


 CLASSES │ old │ new │ diff
─────────┼─────┼─────┼──────────────
 classes │  28 │  28 │   0 (+3 -3)
 methods │  74 │  84 │ +10 (+19 -9)
  fields │  23 │  26 │  +3 (+6 -3)
```
This is still worse than Dagger generating Java, but at least a little less worse.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -89,146 +89,147 @@ internal class ProvidesMethodFactoryGenerator : CodeGenerator {
     val moduleClass = clazz.asClassName()
 
     val content = FileSpec.builder(packageName, className)
-        .addType(
-            TypeSpec
-                .classBuilder(factoryClass)
-                .addAnvilAnnotation()
-                .addSuperinterface(Factory::class.asClassName().parameterizedBy(returnType))
-                .apply {
-                  if (!isObject || parameters.isNotEmpty()) {
-                    primaryConstructor(
-                        FunSpec.constructorBuilder()
+        .apply {
+          val canGenerateAnObject = isObject && parameters.isEmpty()
+          val classBuilder = if (canGenerateAnObject) {
+            TypeSpec.objectBuilder(factoryClass)
+          } else {
+            TypeSpec.classBuilder(factoryClass)
+          }
+
+          classBuilder.addAnvilAnnotation()
+              .addSuperinterface(Factory::class.asClassName().parameterizedBy(returnType))
+              .apply {
+                if (!canGenerateAnObject) {
+                  primaryConstructor(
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -89,146 +89,147 @@ internal class ProvidesMethodFactoryGenerator : CodeGenerator {
...
-                  if (!isObject || parameters.isNotEmpty()) {
-                    primaryConstructor(
-                        FunSpec.constructorBuilder()
+        .apply {
+          val canGenerateAnObject = isObject && parameters.isEmpty()
+          val classBuilder = if (canGenerateAnObject) {
+            TypeSpec.objectBuilder(factoryClass)
+          } else {
+            TypeSpec.classBuilder(factoryClass)
+          }
```
vRallev, line 98: That's the actual change.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -89,146 +89,147 @@ internal class ProvidesMethodFactoryGenerator : CodeGenerator {
...
+                        val argumentList = parameters.asArgumentList(
+                            asProvider = true,
+                            includeModule = !isObject
+                        )
+                        addStatement("return $functionName($argumentList)")
+                      }
+                      .build()
+              )
+              .apply {
+                val builder = if (canGenerateAnObject) this else TypeSpec.companionObjectBuilder()
```
vRallev, line 150: And here. If it can be an object, then we don't need an additional companion object.

PR #90, opened 2020-08-31
Allow using Anvil to generate Dagger factories even though the module uses a Subcomponent.

### LINKED ISSUE #74: Allow using Anvil to generate Factories for subcomponents
At the moment if a module contains a `@Component` or `@Subcomponent` class and Anvil is setup to generate Dagger factories the compiler plugin throws an error, because it cannot generate this code. This check is correct for `@Component`s, but not subcomponents. Subcomponents are generated in the same module as the root component is generated.

vRallev:
This is fine, because Dagger only generates the code for Subcomponents when generating the parent Component.

Fixes #74.


compiler/src/test/java/com/squareup/anvil/compiler/dagger/ComponentDetectorCheckTest.kt
```diff
@@ -30,7 +31,7 @@ class ComponentDetectorCheckTest {
     }
   }
 
-  @Test fun `a Dagger subcomponent causes an error`() {
+  @Test fun `a Dagger subcomponent is allowed`() {
```
zach-klippenstein, line 34: 🙌

PR #93, opened 2020-09-01
Add missing import

### LINKED ISSUE #79: Factory generation: incorrect import for nested classes
Class 1:

```
package com.example.myapplication

import com.example.myapplication.foo.MyClassFoo
import javax.inject.Inject

class MyClass @Inject constructor(d: Dependencies) : MyClassFoo(d)
```

Class 2:
```
package com.example.myapplication.foo

import javax.inject.Inject

open class MyClassFoo(d: Dependencies) {
  class Dependencies @Inject constructor()
}
```

Generated `MyClass_Factory` references `Dependencies` directly, without outer class name, which resuls in compilation error. Both files are otherwise valid Kotlin.

vRallev:
Add missing import for inject constructor factories when an injected type is an inner class of the class itself or are parent type.

Fixes #79.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
@@ -34,6 +34,7 @@ import org.jetbrains.kotlin.resolve.descriptorUtil.getSuperClassNotAny
 import org.jetbrains.kotlin.resolve.descriptorUtil.getSuperInterfaces
 import org.jetbrains.kotlin.resolve.descriptorUtil.module
 import org.jetbrains.kotlin.types.KotlinType
+import org.jetbrains.kotlin.types.typeUtil.supertypes
 import org.jetbrains.org.objectweb.asm.Type
 import javax.inject.Inject
 import javax.inject.Provider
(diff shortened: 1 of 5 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
```diff
@@ -676,6 +676,201 @@ public final class InjectClass_Factory implements Factory<InjectClass> {
...
+        
+        import javax.inject.Inject
+        
+        open class ParentOne(d: Dependencies) {
+          class Dependencies @Inject constructor()
+        }
+        
+        open class ParentTwo(d: Dependencies) : ParentOne(d)
+        
+        class InjectClass @Inject constructor(d: Dependencies) : ParentTwo(d)
```
vRallev, line 859: That's the issue here. Notice how `InjectClass` references `Dependencies`. Even if `ParentOne` and `ParentTwo` are in different packages `InjectClass` doesn't need to add an import for `Dependencies`.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -177,3 +178,10 @@ internal fun KtClassOrObject.generateClassName(
             .shortName()
             .asString()
       }
+
+internal fun List<KotlinType>.getAllSuperTypes(): Sequence<FqName> =
+  generateSequence(this) { kotlinTypes ->
+    if (kotlinTypes.isEmpty()) null else kotlinTypes.flatMap { it.supertypes() }
```
zach-klippenstein, line 184: Super minor nit: `kotlinTypes.takeUnless { it.isEmpty() }?.flatMap { it.supertypes() }`

vRallev: I think that's even easier to read `kotlinTypes.ifEmpty { null }?.flatMap { it.supertypes() }`.

PR #95, opened 2020-09-02
Support classes with multiple generic parameters for field injection

### LINKED ISSUE #91: Incorrect factory generated for class with type parameters
Given I have a class with a type parameter:
```
abstract class MyFragment<T> : Fragment() {

    @Inject
    lateinit var something: Something
```

Anvil will generated a factory as:
```
class MyFragment_MembersInjector(
        private val param0: Provider<Something>
) : MembersInjector<MyFragment> {
    override fun injectMembers(instance: MyFragment) {
        injectSomething(instance, param0.get())
    }

    companion object {
        @JvmStatic
        fun create(param0: Provider<Something>): MyFragment_MembersInjector =
            MyFragment_MembersInjector(param0)

        @JvmStatic
        @InjectedFieldSignature("com.foo.MyFragment")
        fun injectSomething(instance: MyFragment, something: Something) {
            instance.something = something
        }
    }
}
```
which doesn't compile. I'm no expert, but I believe adding a `<*>` to the generated code when there's a type parameter should be enough to fix this

quiro91:
Fixes #91

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
@@ -74,7 +74,9 @@ internal class MembersInjectorGenerator : CodeGenerator {
     val classType = clazz.asTypeName()
         .let {
           if (it is ClassName && clazz.isGenericClass()) {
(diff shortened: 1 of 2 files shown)


vRallev (approved):

vRallev:
Thanks!

Issue #100, opened 2020-09-02
Factory generation: provide a way to disable it in androidTest


Dkhusainov:
I have module which uses anvil Factory generation for it's main source set and an `androidTest` source set which wants to generate(with Dagger) `TestComponen`" and get the dependencies from it to test them.

Currently this leads to an error inside `androidTest` compilation, because Factories for `@Provides` functions in a `TestModule` generated twice(by Dagger and Anvil).

Currently I'm using a wierd workaround(which is another Anvil bug I believe):
If I mark `@Provides` functions in a `TestModule` as `internal`, dagger generates _Factories with different names_ for them, and so the duplicate class exception doesn't happen.
![image](https://user-images.githubusercontent.com/7382654/92042970-e794aa80-ed83-11ea-8635-d0a0abf40492.png)

vRallev:
You should not use Anvil and Dagger at the same time. You only need to use one of them. You also shouldn't use Anvil in modules that have `@Component`. Anvil should actually fail. I recommend to disable generating factories with Anvil in this module.

Dkhusainov:
I'm not, I'm only using Dagger in test source set. How am I gonna create the component otherwise? There is not components in the main source set. 

Are you saying I should use the old way of wiring Dagger modules in `androidTest` - including them manually in `TestComponent`. That's exactly the problem this project is trying to solve, no?

vRallev:
There are limitations:
> This feature can only be enabled in Gradle modules that don't compile any Dagger component. Since Anvil only processes Kotlin code, you shouldn't enable it in modules with mixed Kotlin / Java sources either.

Dkhusainov:
Why not implement it the way that allows to enable/disable the `factoryGeneration`(and potentially other generations besides hints) per android variant? 
One way to do it is read value override from `ext` on `KotlinCompile` task, which users can then modify via `tasks` container

vRallev:
For now this isn't supported. We have a fairly large codebase and didn't need this feature ourselves, yet. I'd suggest to create a separate feature request for that.

Dkhusainov:
How are you testing dagger components on emulator then?02:19, September 3, 2020, Ralf Wondratschek <notifications@github.com>:
For now this isn't supported. We have a fairly large codebase and didn't need this feature ourselves, yet. I'd suggest to create a separate feature request for that.

—You are receiving this because you authored the thread.Reply to this email directly, view it on GitHub, or unsubscribe.
-- Sent from Yandex.Mail for mobile

[...]

PR #101, opened 2020-09-03
Provide a way to override generateDaggerFactories for individual Kotl…

Dkhusainov:
#100

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,6 +251,14 @@ anvil {
 }
 ```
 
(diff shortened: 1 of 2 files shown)


vRallev (changes requested):
While this seems to work I'd rather want a solution with a proper DSL. I'll leave more information in the issue #100

vRallev:
See discussion in #100

PR #103, opened 2020-09-04
Missing Factory class

vRallev:
Generate a Dagger factory for provider methods that Anvil generates in its Anvil specific modules when the experimental feature is enabled to generate Dagger factories.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerator.kt
@@ -5,6 +5,7 @@ import org.jetbrains.kotlin.psi.KtFile
 import java.io.File
 
 internal interface CodeGenerator {
+
   /**
    * Called multiple times in order to create new code. Note that instances should not rely on
    * being able to resolve [projectFiles] to descriptors. They should rather use the Psi APIs to
@@ -22,12 +23,14 @@ internal interface CodeGenerator {
 
   /**
    * Called after the last round of [generateCode] when no new files were produced by any code
-   * generator anymore.
+   * generator anymore. The returned result should contain files that were added or changed one
+   * last time. Code generates that do not impact other code generators get a last chance to
+   * evaluate these results.
    */
   fun flush(
     codeGenDir: File,
     module: ModuleDescriptor
-  ) = Unit
+  ): Collection<GeneratedFile> = emptyList()
 
   data class GeneratedFile(
     val file: File,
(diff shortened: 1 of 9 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerator.kt
```diff
@@ -5,6 +5,9 @@ import org.jetbrains.kotlin.psi.KtFile
 import java.io.File
 
 internal interface CodeGenerator {
+
+  val impactsOtherCodeGenerators: Boolean
```
forrestbice, line 9: This could use some KDoc similar to `flush`. Seems like this encodes whether or not a 2nd pass is required. Would it ever be the case that more than a second pass of the code generators like Dagger would need to be run? If not maybe the name should different. Mostly wondering if there's a more targeted adjective we can use other than `impacts`. No suggestions right now though.

forrestbice (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerator.kt
vRallev: Yeah, I'm struggling with the name, too. Code generation happens in rounds. In the next round generators see the generated code of the previous round. In the initial round it's the source code of the module itself. Once no code was generated anymore, then `flush()` is called. 

Some code generated must see the code after calling `flush()`. They get another chance for an extra round, but only if they generate code that no other code generate should process. 

Maybe I should use a separate type instead of a boolean. I think this is cleaner. Suggestions for the name? `PrivateCodeGenerator`, `ConsumingCodeGenerator`?

forrestbice: I do like the idea of using the type system for this. `MetaCodeGenerator` could be the type for any `Generator` that wants to operate on previously generated code? Or you use `Tier` or `Pass`: `FirstTierGenerator`, `SecondTierGenerator`.

Maybe this is getting too abstract but while trying to come up with alternatives my mind went to recursion and the distinction between first and higher order functions.

How about `FirstOrderGenerator` for any `Generator` that produces code that is "complete" and does not require further processing by other "higher order generators". `HigherOrderGenerator`: A `Generator` which generates code that could be (must be?) processed / consumed by another `Generator`, first or higher order. This would result in all chains / rounds terminating with `FirstOrderGenerator`s.

The first-order and higher-order naming here changes depending on if the description is being applied to what the `Generator` consumes or produces. In the example above it's aimed at the resulting code but maybe it makes more sense to tie it to what the `Generator` consumes. A la `FirstOrderGenerator` a `Generator` that consumes non-generated code. `HigherOrderGenerator` a `Generator` that consumes code that has been generated and requires further processing.

WDYT?

vRallev: Hm, I personally think it's too abstract 😅  I'll go with `PrivateCodeGenerator` for now. I was considering `SideEffectFreeCodeGenerator`, but that's too long.

zach-klippenstein (approved):

PR #104, opened 2020-09-04
Fix the absolute paths as inputs for Kapt and Kotlin compile tasks.

### LINKED ISSUE #65: Usage of absolute path as input

vRallev:
Fixes #65

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
@@ -4,10 +4,12 @@ import com.google.auto.service.AutoService
 import org.gradle.api.Project
 import org.gradle.api.tasks.compile.AbstractCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinCommonOptions
+import org.jetbrains.kotlin.gradle.internal.KaptTask
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinGradleSubplugin
 import org.jetbrains.kotlin.gradle.plugin.SubpluginArtifact
 import org.jetbrains.kotlin.gradle.plugin.SubpluginOption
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
```diff
@@ -50,10 +53,17 @@ class AnvilSubplugin : KotlinGradleSubplugin<AbstractCompile> {
...
+      val relativePath = srcGenDir.relativeTo(project.buildDir).path
+
+      project.tasks.withType(KotlinCompile::class.java) { task ->
+        task.inputs.property(key, relativePath)
+      }
+      project.tasks.withType(KaptTask::class.java) { task ->
+        // We don't apply the compiler plugin for KaptTasks. But for some reason the Kotlin Gradle
+        // plugin copies the inputs with the "kotlinCompile" prefix.
+        task.inputs.property("kotlinCompile.$key", relativePath)
+      }
```
vRallev, line 66: This used to be way more dynamic, but I was seeing strange build errors in CI only. No idea what happened.
```kotlin
      val key = "${getCompilerPluginId()}.${srcGenDirOption.key}"

      project.tasks
          .filter { it is KotlinCompile || it is KaptTask }
          .forEach { task ->
            task.inputs.properties
                .filter { it.key.contains(key) && it.value == srcGenDirPath }
                .forEach {
                  task.inputs.property(it.key, srcGenDir.relativeTo(project.buildDir).path)
                }
          }
```
This new solution is goo enough.

zach-klippenstein (approved):

vRallev:
I'm hesitant to start writing Groovy :-/

runningcode:
It is not required that the test be in groovy, you may write it in whichever language you please. That was just an example I had. :) 
Here is a relocatability test in Kotlin if that helps. https://android.googlesource.com/platform/tools/base/+/mirror-goog-studio-master-dev/build-system/integration-test/application/src/test/java/com/android/build/gradle/integration/cacheability/CacheabilityTest.kt#181

PR #106, opened 2020-09-04
Throw an error if provider method names clash

### LINKED ISSUE #99: Factory generation: overloaded Provides functions
If a module has multiple overloaded `@Provides` functions with the same, only one Factory gets generated. 
Dagger gives an error `Cannot have more than one binding method with the same name in a single module` for such modules.

```
@Module
object MyModule {
    @Provides fun s(): String = ""
    @Provides fun s(s: String): Int = 1
}
```

vRallev:
Throw an error if provider method names clash and Anvil is supposed to generate Dagger factories. Dagger throws the same error.

Fixes #99

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -1,5 +1,6 @@
 package com.squareup.anvil.compiler.codegen.dagger
 
+import com.squareup.anvil.compiler.AnvilCompilationException
 import com.squareup.anvil.compiler.codegen.CodeGenerator.GeneratedFile
 import com.squareup.anvil.compiler.codegen.PrivateCodeGenerator
 import com.squareup.anvil.compiler.codegen.addGeneratedByComment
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -52,8 +53,22 @@ internal class ProvidesMethodFactoryGenerator : CodeGenerator {
...
+
+                if (duplicateFunctions.isNotEmpty()) {
+                  throw AnvilCompilationException(
+                      element = clazz,
+                      message = "Cannot have more than one binding method with the same name in " +
+                          "a single module: ${duplicateFunctions.keys.joinToString()}"
+                  )
+                }
+              }
+              .asSequence()
```
zach-klippenstein, line 71: Why move the asSequence?

zach-klippenstein (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: Because the outer chain is a sequence and Kotlin 1.3 doesn't have the overloads for lists, yet.

vRallev: Not sure why I moved it down, though. I'll revert that.

Issue #82, opened 2020-08-29
Factory generation: unncecessary private imports


Dkhusainov:
```
package com.example.myapplication

import javax.inject.Inject
import com.example.myapplication.MyClass.Result.*

class MyClass @Inject constructor() {

  private sealed class Result {
    class Some : Result()
    class Other : Result()
  }

  private fun some(): Result = Some()
  private fun other(): Result = Other()
}
```

Generated factory contains unnecessary(for compilation) import `import com.example.myapplication.MyClass.Result.*`, which fails to compile.

vRallev:
The compiler plugin uses the Psi API. For performance reasons it doesn't resolve all types and doesn't resolve all imports properly. As a workaround it inherits all imports from the origin file even though they might be unused. This is also necessary to work around issues with star imports.

PR #109, opened 2020-09-06
Fix wrong imports

### LINKED ISSUE #97: Factory generation: inject nested class into nested class
Anvil: 2.0,2

File1:
```
package com.example.myapplication

import com.example.myapplication.foo.Foo
import javax.inject.Inject

class MyClass (bar: Bar): Foo(bar) {
  class MyClassInner @Inject constructor(bar: Bar)
}
```

File2
```
package com.example.myapplication.foo

import javax.inject.Inject

abstract class Foo constructor(bar: Bar) {
    class Bar @Inject constructor()
}
```

Generated `MyClass_MyClassInner_Factory` has incorrect imports

vRallev:
This PR reworks how imports are resolved. Before I implemented an "easy fix, "workaround" or "hack" 🙃  As usual, those things bite one in the long run. Now I resolve all types properly so that Kotlin Poet can add the correct imports.

Fixes #97, #82

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
@@ -13,6 +13,7 @@ import com.squareup.kotlinpoet.LambdaTypeName
 import com.squareup.kotlinpoet.ParameterSpec
 import com.squareup.kotlinpoet.ParameterizedTypeName
 import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.STAR
 import com.squareup.kotlinpoet.TypeName
 import com.squareup.kotlinpoet.asClassName
 import com.squareup.kotlinpoet.jvm.jvmSuppressWildcards
@@ -26,6 +27,7 @@ import org.jetbrains.kotlin.psi.KtCallableDeclaration
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtFunctionType
 import org.jetbrains.kotlin.psi.KtNullableType
+import org.jetbrains.kotlin.psi.KtProjectionKind
 import org.jetbrains.kotlin.psi.KtTypeArgumentList
 import org.jetbrains.kotlin.psi.KtTypeElement
 import org.jetbrains.kotlin.psi.KtTypeProjection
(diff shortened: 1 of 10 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -83,18 +83,12 @@ internal class ProvidesMethodFactoryGenerator : CodeGenerator {
 
     val parameters = function.valueParameters.mapToParameter(module)
 
-    // This is a little hacky. The typeElement could be "String", "Abc", etc., but also a generic
-    // type like "List<String>". We simply copy this literal as return type into our generated
-    // code. Kotlinpoet will add an import for this literal like "import String", which we later
-    // will remove again.
-    //
-    // This solution is a lot easier than trying to resolve all FqNames for each type.
```
vRallev, line 91: That was the hack.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -204,58 +268,3 @@ internal fun FileSpec.writeToString(): String {
   }
   return stream.toString()
 }
-
-/**
- * Removes invalid imports that can be caused by [requireTypeName] and copies imports from the
- * source file. Copying imports is necessary for star imports.
- */
-internal fun String.replaceImports(clazz: KtClassOrObject): String {
```
vRallev, line 212: That workaround isn't need anymore. The test coverage is really high, that's why I'm confident in this new solution.

vRallev:
I was a little afraid of the overhead compared to the old solution, but it is negligible. I ran benchmarks over the weekend and the overhead is around 1% (I wonder if this actually came from somewhere else on the machine).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -9,61 +9,131 @@ import com.squareup.anvil.compiler.providerFqName
...
-  return try {
-    ClassName.bestGuess(asString())
-  } catch (e: IllegalArgumentException) {
-    // This happens when the class name starts with a lowercase character.
-    TypeVariableName(asString())
+private fun FqName.asClassName(module: ModuleDescriptor): ClassName {
+  try {
+    return ClassName.bestGuess(asString())
+  } catch (ignored: IllegalArgumentException) {
+    // Probably lowercase class.
```
steve-the-edwards, line 65: Nit: could comment intentional fall-through.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/TestClasses.kt
```diff
@@ -0,0 +1,11 @@
+@file:Suppress("unused")
+
+package com.squareup.anvil.compiler.dagger
+
+// These classes is used in some of the unit tests. Don't touch them.
```
steve-the-edwards, line 5: nit: "... are used ..."

steve-the-edwards (approved):

PR #114, opened 2020-09-11
Fix Confguration Caching: #113

### LINKED ISSUE #113: Configuration Caching: the anvil plugin is adding a doFirst block to the task KotlinCompileWithWorkers from KGP and this block uses the project object at execution time
The bug filed against KGP: https://youtrack.jetbrains.com/issue/KT-41783 was indeed a bug created by the Anvil plugin.

The line https://github.com/square/anvil/blob/main/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt#L86
is adding a doFirst block to the compileTask of KGP (in our case: `KotlinCompileWithWorkers` , but the problem would be the same if no workers would be used). 

And this block is using the project object to access source files and plugins: https://github.com/square/anvil/blob/main/gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt#L25

The problem is that the project object is not availalbe during execution time with Configuration Caching.

stephanenicolas:
The first commit solves the issue. 
The second commit makes configuration more lazy when it comes to IOs.

Fixes #113


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -79,12 +79,15 @@ open class AnvilPlugin : Plugin<Project> {
     project.tasks
         .withType(KotlinCompile::class.java)
         .configureEach { compileTask ->
+          val shouldDisablePreciseJavaTracking = CheckMixedSourceSet(compileTask.project, compileTask)
+              .shouldDisablePreciseJavaTrackingIfNeeded()
           compileTask.doFirst {
             // Disable precise java tracking if needed. Note that the doFirst() action only runs
-            // if the task is not up to date. That's ideal, because if nothing needs to be
+             // if the task is not up to date. That's ideal, because if nothing needs to be
```
vRallev, line 86: Wrong indent.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -79,12 +79,15 @@ open class AnvilPlugin : Plugin<Project> {
...
             // compiled, then we don't need to disable the flag.
-            CheckMixedSourceSet(compileTask.project, compileTask)
-                .disablePreciseJavaTrackingIfNeeded()
+            if(shouldDisablePreciseJavaTracking.get()) {
```
vRallev, line 88: Missing space. There are more KtLint violations: https://github.com/square/anvil/pull/114/checks?check_run_id=1099361039

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -22,44 +23,52 @@ open class CheckMixedSourceSet(
...
-        .findVariantForCompileTask(compileTask)
-        .sourceSets
-        .asSequence()
-        .flatMap { it.javaDirectories.asSequence() }
-        .flatMap { it.walk() }
-        .filter { it.isFile }
-  }
+  private fun Project.getSourceFilesAndroidProject(compileTask: KotlinCompile) =
+    this.files( this.provider {
+      this.androidVariants()
```
vRallev, line 54: All the `this.` can be removed.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -22,44 +23,52 @@ open class CheckMixedSourceSet(
...
-  private fun getSourceFilesAndroidProject(): Sequence<File> {
-    return project.androidVariants()
-        .findVariantForCompileTask(compileTask)
-        .sourceSets
-        .asSequence()
-        .flatMap { it.javaDirectories.asSequence() }
-        .flatMap { it.walk() }
-        .filter { it.isFile }
-  }
+  private fun Project.getSourceFilesAndroidProject(compileTask: KotlinCompile) =
```
vRallev, line 52: Why did you remove the return type and actually change it? Please specify the return type again.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -22,44 +23,52 @@ open class CheckMixedSourceSet(
...
+    this.files( this.provider {
```
vRallev, line 53: I actually missed the provider first. This would be easier to read if you assign the provider to a temporary variable.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -22,44 +23,52 @@ open class CheckMixedSourceSet(
...
+          .flatMap { it.javaDirectories.asSequence() }
+          .flatMap { it.walk() }
+          .filter { it.isFile }
+          .toSet()
+    })
 
-  private fun getSourceFilesJavaProject(): Sequence<File> {
-    return project.convention.getPlugin(JavaPluginConvention::class.java)
+  private fun Project.getSourceFilesJavaProject() = this.files( this.provider {
+    this.convention.getPlugin(JavaPluginConvention::class.java)
```
vRallev, line 65: I don't understand how this works. You make this lazy by wrapping it a provider. Inside of the lambda you shouldn't use the project reference anymore, because it's called at execution time. But you actually use it by calling `this.convention`. How is this different than before?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -22,44 +23,52 @@ open class CheckMixedSourceSet(
...
+    }
 
     // If Kapt is enabled then it usually generates Java code making Kotlin compilation use a
     // mixed source set.
-    val isKaptApplied = project.plugins.hasPlugin(Kapt3GradleSubplugin::class.java)
-
-    if (hasJavaFile || isKaptApplied) {
-      compileTask.usePreciseJavaTracking = false
+    val isKaptApplied = project.provider {
+      project.plugins.hasPlugin(Kapt3GradleSubplugin::class.java)
```
vRallev, line 42: You still reference `project` at execution time.

vRallev (changes requested):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
stephanenicolas: `./gradlew check` would be a good addition to `CONTRIBUTING.MD` ;)

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
stephanenicolas: When is the return type necessary ? We needed a lazy file representation here, I used a `FileCollection`.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
stephanenicolas: I just chatted with Gradle about this. The configuration will evaluate the boolean property, and CoCa will serialize it which leads to 2 issues:
1. the IO operations are performed too early still
2. the build wouldn't be aware of any change in what we evaluate (source file presence) after the configuration is cached once. 
The solution is to distinguish more formally the creation of the file collection we work on, and separate better execution, and io operations. I will work on that now.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
vRallev: When it's not obvious I specify the return type explicitly. Here it's not obvious. Public function should always specify the type explicitly.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -1,65 +0,0 @@
...
- * In a mixed Kotlin / Java source set the Kotlin compiler might crash with an error like this:
- *
- * java.lang.AssertionError: Duplicated JavaClassDescriptor a/b/c/Hello reported to IC.
- *
- * That's a known bug: https://youtrack.jetbrains.com/issue/KT-38576
- *
- * The workaround for now is to set the kotlin.incremental.usePreciseJavaTracking flag to false for
- * these module using this task.
- */
-open class CheckMixedSourceSet(
```
vRallev, line 20: The change itself is good. But I really dislike that you deleted this class and moved everything into the plugin class. You bloat the plugin class even more. I suggest to keep this class simply move the logic here. This is just a question of how to organize code.

You also removed the comment from this class which is very valuable to understand why we do this at all.

vRallev (changes requested):

vRallev (approved):

PR #116, opened 2020-09-11
Support named imports in @Provides annotated methods

### LINKED ISSUE #115: Support named import
The following class is processed successfully by Dagger compiler but not by Anvil
```
package com.squareup.test

import dagger.Module
import dagger.Provides
import java.lang.Runnable as NamedRunnable
        
@Module
object DaggerModule1 {
    @Provides fun provideRunner(): NamedRunnable = NamedRunnable {}
}
```

quiro91:
Fixes #115

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
@@ -282,9 +282,7 @@ internal fun PsiElement.requireFqName(
   containingKtFile.importDirectives
     .firstOrNull { classReference == it.importPath?.importedName?.asString() }
     ?.importedFqName
-    ?.let {
-      module.resolveClassByFqName(it, FROM_BACKEND)?.let { return it.fqNameSafe }
-    }
+    ?.let { return it }
 
   // Everything else isn't supported.
   throw AnvilCompilationException(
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -278,6 +278,13 @@ internal fun PsiElement.requireFqName(
         module.findClassOrTypeAlias(importFqName, classReference)?.let { return it.fqNameSafe }
       }
 
+  // Check if it's a named import
```
vRallev, line 281: Please finish the comment with a dot.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -278,6 +278,13 @@ internal fun PsiElement.requireFqName(
...
+  containingKtFile.importDirectives
+    .firstOrNull { classReference == it.importPath?.importedName?.asString() }?.importedFqName
```
vRallev, line 283: Put `?.importedFqName` on a new line for better readability.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -278,6 +278,13 @@ internal fun PsiElement.requireFqName(
...
+    ?.let {
+      module.resolveClassByFqName(it, FROM_BACKEND)?.let { return it.fqNameSafe }
```
vRallev, line 285: That seems unnecessary. Can't you just return `it` immediately without resolving? You already have the fully qualified name.

vRallev (changes requested):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
quiro91: Makes sense, if it matches the reference of the class I can just return it 🤦

vRallev (approved):

PR #118, opened 2020-09-14
Upgrade the Kotlin versions that are used in CI for testing.

vRallev:


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   build:
     runs-on: ${{ matrix.os }}
     timeout-minutes: 15
-    continue-on-error: ${{ startsWith(matrix.kotlin-version, '1.4') }}
+    continue-on-error: ${{ startsWith(matrix.kotlin-version, '1.4.2') }}
```
cdeange, line 13: `1.4.20` to be more precise?

cdeange (approved):

.github/workflows/ci.yml
vRallev: I don't think it makes a difference ¯\\_(ツ)_/¯

PR #121, opened 2020-09-17
Support the new IR backend

vRallev:
This PR adds support for the new IR backend. It uses APIs from Kotlin 1.4.20 that aren't supported in older compiler versions. To still be able to compile the project with older Kotlin versions the IR related source code is written in a different Gradle source set. This source set is only added to the module when Kotlin 1.4.20 is selected. That means published versions of Anvil won't support the new IR backend, because the compile the project with Kotlin 1.3. If we need to support the new IR backend, then I can publish special versions like I did initially with Kotlin 1.4.

In CI we now run all tests with and without the new IR backend for Kotlin 1.4.20. For older Kotlin versions we only use the old backend.

All of this is super experimental. E.g. with Kotlin `1.4.255-SNAPSHOT` I'm seeing compile errors.


compiler/build.gradle
```diff
@@ -20,6 +22,10 @@ dependencies {
   testImplementation deps.truth
 }
 
+if (rootProject.ext.kotlinVersion.startsWith('1.4.2')) {
+  sourceSets.main.java.srcDirs += ['src/main/ir']
+}
```
vRallev, line 27: That's the workaround with the source set mentioned in the description.

compiler/generate_build_properties.gradle
```diff
@@ -0,0 +1,26 @@
...
+  inputs.property 'useIR', rootProject.ext.kotlinUseIR
+
+  outputs.file buildPropertiesFile
+
+  doLast {
+    buildPropertiesFile.parentFile.mkdirs()
+    buildPropertiesFile.write """\
+      package com.squareup.anvil.compiler
+      
+      internal const val USE_IR = ${rootProject.ext.kotlinUseIR}
```
vRallev, line 19: This constant is generated for unit tests. I already use this workaround for the Gradle plugin. I miss build configs from AGP.

compiler/src/main/ir/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -0,0 +1,331 @@
...
+import org.jetbrains.kotlin.ir.types.defaultType
+import org.jetbrains.kotlin.ir.types.impl.IrDynamicTypeImpl
+import org.jetbrains.kotlin.ir.util.defaultType
+import org.jetbrains.kotlin.ir.util.getPackageFragment
+import org.jetbrains.kotlin.ir.util.isInterface
+import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.types.Variance.INVARIANT
+
+internal class ModuleMergerIr(
```
vRallev, line 27: I literally copied the old class and converted it piece by piece to the new IR backend APIs.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -68,6 +69,22 @@ class AnvilComponentRegistrar : ComponentRegistrar {
...
+
+    try {
+      // This extension depends on Kotlin 1.4.20 and the code fails to compile with older compiler
+      // versions. Anvil will only support the new IR backend with 1.4.20. To avoid compilation
+      // errors we only add the source code to this module when IR is enabled. So try to
+      // dynamically look up the class name and add the extension when it exists.
+      val moduleMergerIr = Class.forName("com.squareup.anvil.compiler.ModuleMergerIr")
+          .declaredConstructors
+          .single()
+          .newInstance(scanner) as IrGenerationExtension
```
vRallev, line 81: The comment describes why this is needed.

compiler/src/test/java/com/squareup/anvil/compiler/TestUtils.kt
```diff
@@ -26,7 +26,7 @@ internal fun compile(
   return KotlinCompilation()
       .apply {
         compilerPlugins = listOf(AnvilComponentRegistrar())
-        useIR = false
+        useIR = USE_IR
```
vRallev, line 29: That's the generated constant from the Gradle script.

.github/workflows/ci.yml
```diff
@@ -7,8 +7,8 @@ on:
   pull_request:
 
 jobs:
-  build:
-    runs-on: ${{ matrix.os }}
+  test-ubuntu:
```
vRallev, line 10: I had to rewrite the entire CI config, because shards got stuck on Github action and I couldn't figure out why. There's a lot of duplication right now, but it works better than before.

zach-klippenstein (approved):

.github/workflows/ci.yml
```diff
@@ -18,9 +18,54 @@ jobs:
...
+        uses: actions/upload-artifact@v2
+        if: ${{ failure() }}
+        with:
+          name: test-results
+          path: ./**/build/reports/tests/
+
+
+  test-windows:
+    runs-on: windows-latest
+    timeout-minutes: 15
```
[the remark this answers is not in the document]
vRallev: Yes, before it was using a matrix.

compiler/src/main/java/com/squareup/anvil/compiler/IrUtils.kt
```diff
@@ -0,0 +1,105 @@
...
+  return referenceClass(fqName) ?: throw AnvilCompilationException(
+      message = "Couldn't resolve reference for $fqName"
+  )
+}
+
+internal fun IrAnnotationContainer.annotationOrNull(
+  annotationFqName: FqName,
+  scope: FqName? = null
+): IrConstructorCall? {
+  val annotation = getAnnotation(annotationFqName)
```
cdeange: I think this whole thing could just be:

```kotlin
return getAnnotation(annotationFqName)?.takeIf { annotation ->
  scope != null && annotation.scope() == scope
}
```

dependencies.gradle
```diff
@@ -5,6 +5,13 @@ ext {
   kotlinVersion = project.hasProperty('square.kotlinVersion') ?
       project.getProperty('square.kotlinVersion') : '1.3.72'
 
+  kotlinUseIR = (project.hasProperty('square.useIR') ?
+      project.getProperty('square.useIR') : 'false').toBoolean()
+
+  if (kotlinUseIR && !kotlinVersion.startsWith('1.4.2')) {
+    throw new GradleException("The new IR backend is only testable with Kotlin 1.4.20 or newer.")
```
cdeange, line 12: A bit of a mismatch here – you're only allowing this on `1.4.20` _or newer_, but only checking `startsWith('1.4.2')`. This would break if you were, for example, running against kotlin version `1.4.30`.

Should you be doing some kind of semver-aware check here, comparing version codes with something like `org.gradle.util.VersionNumber`?

cdeange (approved):
A lot of the IR stuff is uhh kinda unknown to me so don't take my word that it works properly 🙈 @zach-klippenstein might be a better reviewer for that area than I am.

compiler/src/main/java/com/squareup/anvil/compiler/IrUtils.kt
vRallev: It's an OR, but nonetheless you're right!
```
  return getAnnotation(annotationFqName)?.takeIf { annotation ->
    scope == null || annotation.scope() == scope
  }
```

dependencies.gradle
vRallev: Added `kotlinIrSupported = VersionNumber.parse(kotlinVersion).baseVersion >= VersionNumber.parse("1.4.20")`.

cdeange: Nice!

PR #122, opened 2020-09-17
Support variance keywords and preserve annotations for types

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -108,11 +110,32 @@ internal fun KtTypeReference.requireTypeName(
...
+              typeArgumentList.arguments.map { typeProjection ->
+                if (typeProjection.projectionKind == KtProjectionKind.STAR) {
                   STAR
                 } else {
-                  (it.typeReference ?: it.fail()).requireTypeName(module)
+                  val typeReference = typeProjection.typeReference ?: typeProjection.fail()
+                  typeReference
+                      .requireTypeName(module)
+                      .let { typeName ->
+                        // Preserve annotations, e.g. List<@JvmSuppressWildcards Abc>.
```
cdeange, line 121: nit: unit test?

cdeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
vRallev: Part of the next commit.

PR #127, opened 2020-09-23
Support constructor injection for classes with bounded type parameters

### LINKED ISSUE #126: Support constructor injection for classes with bounded type parameters is unsupported
Looks like:
```
class InjectClass<T: CharSequence> @Inject constructor(val element: T)
```
is not currently supported by Anvil

quiro91:
Fixes #126


vRallev (approved):
The change looks good. I addressed smaller nits in a separate commit. Thanks for the contribution!

PR #131, opened 2020-09-23
Print a better user-friendlier error message for Dagger provider methods with an implicit return type.

### LINKED ISSUE #124: Improve error message for implicit return types
Provider methods for Dagger modules require an explicit return type. The current error message isn't userfriendly at all:
```
e: com.squareup.anvil.compiler.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't obtain type reference.
File being compiled: (41,5) in /../File.kt
```

vRallev:
Fixes #124

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
@@ -10,6 +10,7 @@ import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.COMPILATION_ERRO
 import com.tschuchort.compiletesting.KotlinCompilation.Result
 import dagger.Lazy
 import dagger.internal.Factory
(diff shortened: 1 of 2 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2249,6 +2250,28 @@ public final class DaggerComponentInterface implements ComponentInterface {
     }
   }
 
+  @Test fun `a return type for a provider method is required`() {
```
cdeange, line 2253: I love these kinds of integration tests. So easy to read and comprehend.

cdeange (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
vRallev: 😇

PR #133, opened 2020-09-24
Use TCA APIs on this branch

stephanenicolas:
On master the anvil plugin already uses TCA, but not on this branch.


vRallev (changes requested):
This already affects master, it's probably only hidden because the APIs for compiler plugins for Kotlin 1.3 are broken. Can you rebase this PR and merge this into master? I'll update the Kotlin 1.4 branches and include this change there, too.

PR #134, opened 2020-09-24
Use TCA apis.

stephanenicolas:
3 changes, 1 in the build system of anvil, the other 2 in the plugin itself.
Replaces https://github.com/square/anvil/pull/133

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
@@ -4,6 +4,7 @@ import com.google.auto.service.AutoService
 import org.gradle.api.Project
 import org.gradle.api.tasks.compile.AbstractCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinCommonOptions
+import org.jetbrains.kotlin.gradle.internal.KaptGenerateStubsTask
 import org.jetbrains.kotlin.gradle.internal.KaptTask
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinGradleSubplugin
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
```diff
@@ -56,13 +57,18 @@ class AnvilSubplugin : KotlinGradleSubplugin<AbstractCompile> {
...
-      project.tasks.withType(KaptTask::class.java) { task ->
-        // We don't apply the compiler plugin for KaptTasks. But for some reason the Kotlin Gradle
-        // plugin copies the inputs with the "kotlinCompile" prefix.
-        task.inputs.property("kotlinCompile.$key", relativePath)
+      project.tasks.withType(KaptTask::class.java).configureEach { task ->
+          // We don't apply the compiler plugin for KaptTasks. But for some reason the Kotlin Gradle
+          // plugin copies the inputs with the "kotlinCompile" prefix.
+          task.inputs.property("kotlinCompile.$key", relativePath)
+      }
+      project.tasks.withType(KaptGenerateStubsTask::class.java).configureEach { task ->
```
vRallev, line 68: That's not needed. `KaptGenerateStubsTask` is a `KotlinCompile` task and handled above. With Kotlin 1.4 there isn't the extra key and there's no need to add the input property
```
task.inputs.property("kotlinCompile.$key", relativePath)
```
I tested exactly this yesterday.

vRallev (changes requested):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
stephanenicolas: It fixed the error of the previous build...

vRallev (approved):

PR #135, opened 2020-09-25
Avoid adding the absolute file path as task input

### LINKED ISSUE #65: Usage of absolute path as input

vRallev:
Avoid adding the absolute file path for the src gen directory as task input by using a FilesSubpluginOption.

Fixes #65

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
@@ -4,13 +4,11 @@ import com.google.auto.service.AutoService
 import org.gradle.api.Project
 import org.gradle.api.tasks.compile.AbstractCompile
 import org.jetbrains.kotlin.gradle.dsl.KotlinCommonOptions
-import org.jetbrains.kotlin.gradle.internal.KaptGenerateStubsTask
-import org.jetbrains.kotlin.gradle.internal.KaptTask
+import org.jetbrains.kotlin.gradle.plugin.FilesSubpluginOption
 import org.jetbrains.kotlin.gradle.plugin.KotlinCompilation
 import org.jetbrains.kotlin.gradle.plugin.KotlinGradleSubplugin
 import org.jetbrains.kotlin.gradle.plugin.SubpluginArtifact
 import org.jetbrains.kotlin.gradle.plugin.SubpluginOption
-import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.io.File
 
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
```diff
@@ -42,40 +40,13 @@ class AnvilSubplugin : KotlinGradleSubplugin<AbstractCompile> {
...
-          // We don't apply the compiler plugin for KaptTasks. But for some reason the Kotlin Gradle
-          // plugin copies the inputs with the "kotlinCompile" prefix.
-          task.inputs.property("kotlinCompile.$key", relativePath)
-      }
-      project.tasks.withType(KaptGenerateStubsTask::class.java).configureEach { task ->
-          // We don't apply the compiler plugin for KaptTasks. But for some reason the Kotlin Gradle
-          // plugin copies the inputs with the "kotlinCompile" prefix.
-          task.inputs.property("kotlinCompile.$key", relativePath)
-      }
-    }
```
vRallev, line 73: The stupid workaround is gone.

integration-tests/tests/build.gradle
```diff
@@ -17,26 +17,18 @@ dependencies {
...
-        }
+  task.inputs.properties.each { key, value ->
+    if (key.contains('com.squareup.anvil.compiler.src-gen-dir')) {
+      boolean isAbsolute = new File(value.toString()).isAbsolute()
+      if (isAbsolute) {
+        throw new GradleException(
+            "The workaround for absolute paths in the Anvil Gradle plugin is broken. Key: $key, " +
+                "value: $value, task: $task. For more context see " +
+                "https://github.com/square/anvil/issues/65"
+        )
```
vRallev, line 31: I considered deleting this entirely, because there is no more workaround. I'll wait for more feedback, then I'll delete it.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
cdeange: Just out of curiosity, why is it no longer necessary?

cdeange (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilSubplugin.kt
vRallev: Because the file path isn't added to the task input anymore in the first place. Instead of using `SubpluginOption` I'm using now `FilesSubpluginOption`, which values aren't added.

PR #136, opened 2020-09-28
Use Kotlin 1.4.20-M1 in CI

vRallev:


compiler/build.gradle
```diff
@@ -17,6 +17,7 @@ dependencies {
   kapt deps.auto.service.processor
 
   testImplementation deps.dagger2.compiler
+  testImplementation deps.kotlin.annotation_processing_embeddable
```
vRallev, line 20: This dependency is needed for a transitive dependency.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -91,42 +91,19 @@ class AnvilComponentRegistrar : ComponentRegistrar {
...
-    // The bug is that they use "extensionClass.canonicalName".
-    val classNameToUnregister = extensionClass.name
-    unregisterExtensions({ className, _ ->
-      classNameToUnregister != className
-    }, true)
+    ExtensionPointImpl::class.java.declaredMethods
+        .first { it.name == "doRegisterExtension" }
+        .use {
+          it.invoke(analysisHandlerExtensionPoint, extension, LoadingOrder.FIRST, project)
+        }
```
vRallev, line 107: I called this out in the commit message. I'm not concerned about this workaround at all, because there is supposed to be a public API for that, but I can't see it yet. This API is actually part of the IntelliJ platform. The Kotlin compiler repackages and bundles the code. But this one API is missing either because it's an older version or the method is stripped. JetBrains actually suggested in the ticket to use the non-existent API https://youtrack.jetbrains.com/issue/KT-42103

tehras (approved):

cdeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
zach-klippenstein: That comment might be helpful to actually have as a comment in the source code.

zach-klippenstein (approved):

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
vRallev: Yep, good idea.

PR #139, opened 2020-09-30
Handle inner classes explicitly when resolving the Kotlin type

vRallev:
When resolving the Kotlin type of inner class from dependencies the compiler might fail. It tries to load `my.package.Class$Inner` and fails whereas is should load `my.package.Class.Inner`. I was only able to reproduce this in our Square repository and confirmed that this workaround fixes the bug.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
@@ -19,6 +19,8 @@ import org.jetbrains.kotlin.codegen.state.KotlinTypeMapper
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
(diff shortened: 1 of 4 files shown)


vRallev:
I'm a little said that I wasn't able to reproduce the issue and couldn't create a test.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -125,29 +127,50 @@ internal fun AnnotationDescriptor.getAnnotationValue(key: String): ConstantValue
...
+  val normalClass = this.value
+  if (normalClass !is NormalClass) return argumentType
+
+  val classId = normalClass.value.classId
+
+  return module
+      .findClassAcrossModuleDependencies(
+          classId = ClassId(
+              classId.packageFqName,
+              FqName(classId.relativeClassName.asString().replace('$', '.')),
```
cdeange, line 158: Does this clash with normal, non-inner classes that have a `$` in them? Considering something like how AutoValue generates class names: `abstract class $$AutoValue_Card extends Card { ... }`. Or would that be handled by some of the code above?

cdeange (approved):
Unsure how much of this is unit testable, but that might go a long way to protect against regressions.

PR #141, opened 2020-10-02
Generated hint for @ContributesBinding and @ContributesTo explicitly defined as internal to support Kotlin 1.4 Explicit API Mode

### LINKED ISSUE #140: Anvil is incompatible with Kotlin 1.4.0 "-Xexplicit-api=strict" flag
When using Anvil in a module that has `"-Xexplicit-api=strict"` the project fails to compile with the following message:

```
Visibility must be specified in explicit API mode
Return type must be specified in explicit API mode
```

This is happening due to the following snippet of code (for both `ContributesBindingGenerator` and `ContributesToGenerator`):

```
         val content = """
              package $generatedPackage
              
              val ${className.replace('.', '_')}$REFERENCE_SUFFIX = $className::class
              val ${className.replace('.', '_')}$SCOPE_SUFFIX = $scope::class
          """.trimIndent()
```

The reason is because the variables do not have visibility modifiers neither return types but are defined inside the module defined as strict mode. Other than that, I didn't experience any other issues with strict mode so far.

To simulate, just turn on the strict mode in a module that uses Anvil. For example:

```
compileKotlin {
    kotlinOptions {
        jvmTarget = JavaVersion.VERSION_1_8.toString()
        freeCompilerArgs = [
                "-Xexplicit-api=strict"
        ]
    }
}
```

marcellogalhardo:
When using Anvil in a module that has strict API Mode enable, `ContributesBindingGenerator` and `ContributesToGenerator` will generate the hint used internally for merging modules and components interface without a visibility modifier and therefore, Kotlin Explicit API will fail the compilation.

Fixes #140.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesBindingGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesBindingGenerator.kt
@@ -52,8 +52,8 @@ internal class ContributesBindingGenerator : CodeGenerator {
           val content = """
             package $generatedPackage
             
-            val ${className.replace('.', '_')}$REFERENCE_SUFFIX = $className::class
-            val ${className.replace('.', '_')}$SCOPE_SUFFIX = $scope::class
+            public val ${className.replace('.', '_')}$REFERENCE_SUFFIX: kotlin.reflect.KClass<$className> = $className::class
+            public val ${className.replace('.', '_')}$SCOPE_SUFFIX: kotlin.reflect.KClass<$scope> = $scope::class
           """.trimIndent()
           file.writeText(content)
 
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesToGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesToGenerator.kt
@@ -64,8 +64,8 @@ internal class ContributesToGenerator : CodeGenerator {
           val content = """
               package $generatedPackage
               
-              val ${className.replace('.', '_')}$REFERENCE_SUFFIX = $className::class
-              val ${className.replace('.', '_')}$SCOPE_SUFFIX = $scope::class
+              public val ${className.replace('.', '_')}$REFERENCE_SUFFIX: kotlin.reflect.KClass<$className> = $className::class
+              public val ${className.replace('.', '_')}$SCOPE_SUFFIX: kotlin.reflect.KClass<$scope> = $scope::class
           """.trimIndent()
           file.writeText(content)
 
--- a/compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
@@ -4,7 +4,6 @@ import com.google.common.truth.Truth.assertThat
 import com.squareup.anvil.annotations.MergeComponent
 import com.squareup.anvil.annotations.MergeSubcomponent
 import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.COMPILATION_ERROR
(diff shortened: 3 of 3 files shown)


marcellogalhardo:
I'm not sure what went wrong, I will investigate it.

marcellogalhardo:
I'm investigating the failing tests. 🤔

marcellogalhardo:
I realised my previous solution was unnecessary as these hints are usually used internally. Therefore, `internal` looks like the right visibility modifier to be used here. I also rebased my branch on top of last main branch. All tests are now green.

I updated the PR description.

vRallev (changes requested):
> I realised these generated hints are used internally only. Therefore, `internal` looks like the right visibility modifier to be used here.

That's not correct. `internal` means this code is only used inside of this module and is not visible outside of this modules. However, Anvil needs to see the generated code in other modules. The generated hints must be public.

marcellogalhardo:
My understand was that these values are loaded directly from the GeneratedHint file and the modifiers didn't affect Anvil but I have missed something.

I changed the code to use `public` visibility modifier but Explicit API Mode demands to set a return type. This made [ModuleMergerTest.'inner modules in a merged component fail'](https://github.com/marcellogalhardo/anvil/blob/main/compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt#L713) fail and while investigating I concluded that:

_Before the changes:_ when using a inner class without a return type `ContributedHint.reference` was a `DeferredType` that would be evaluate only when trying to access it ([ClassScanner.findContributedClasses](https://github.com/marcellogalhardo/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt#L55)) causing an error and an [ExitCode.INTERNAL_ERROR](https://github.com/marcellogalhardo/anvil/blob/main/compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt#L729). From the stack trace:
```
	...
	at org.jetbrains.kotlin.types.DeferredType.getDelegate(DeferredType.java:107)
	at org.jetbrains.kotlin.types.WrappedType.getArguments(KotlinType.kt:129)
	at com.squareup.anvil.compiler.UtilsKt.argumentType(Utils.kt:116)
	at com.squareup.anvil.compiler.ContributedHint.toClassDescriptor(ClassScanner.kt:87)
	at com.squareup.anvil.compiler.ContributedHint.access$toClassDescriptor(ClassScanner.kt:68)
	at com.squareup.anvil.compiler.ContributedHint$reference$2.invoke(ClassScanner.kt:72)
	...
```

_After the changes:_ with a return type declared, the `ContributedHint.reference` will always have a type defined (e.g., `KClass<InnerModule>`) and no `DeferredType` will be used. The inner module validation would still happen as expected causing a `ExitCode.COMPILATION_ERROR`. I didn't find any other difference.

The other details of the test are still the same. And the right exception is still throw:

```
test/Source.kt: (10, 18): com.squareup.anvil.compiler.AnvilCompilationException: Back-end (JVM) Internal error: It seems like you tried to contribute an inner class to its outer class. This is not supported and results in a compiler error.
File being compiled: (10,18) in /var/folders/b4/131t8pg542j3q_1g7c0b62wc0000gp/T/Kotlin-Compilation7934599964874323027/sources/src/main/java/com/squareup/test/Source.kt
The root cause java.lang.IllegalStateException was thrown at: org.jetbrains.kotlin.types.ErrorUtils$1.getPackage(ErrorUtils.java:95)
```

I changed the test to expect an `ExitCode.COMPILATION_ERROR` but I'm not sure if this feels wrong and I don't see a way to keep the same `DeferredType` behaviour while with a return type declared...

What do you think @vRallev? 🤔 

Sorry for the long message.

vRallev:
Yes, that is strange. But I don't see why this behavior changed. The original message is still preserved, so I'm not concerned.

vRallev (approved):

vRallev:
Note that I created #144 as a follow-up.

PR #142, opened 2020-10-12
Use Kotlin 1.4.20-M2-158 in CI

vRallev:
Use the new public API for registering extensions with a specific order. This will allow us to remove the workaround to register our extension first.

See https://youtrack.jetbrains.com/issue/KT-42103


compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -105,10 +106,22 @@ class AnvilComponentRegistrar : ComponentRegistrar {
...
-        }
+    if (USE_IR) {
+      // TODO: Remove this reflection call when upgrading to Kotlin 1.4.20.
+      // project.extensionArea
+      //     .getExtensionPoint(AnalysisHandlerExtension.extensionPointName)
+      //     .registerExtension(extension, LoadingOrder.FIRST, project)
+      ExtensionPoint::class.java.declaredMethods
+          .first { it.name == "registerExtension" && it.parameterCount == 3 }
+          .use {
+            it.invoke(analysisHandlerExtensionPoint, extension, LoadingOrder.FIRST, project)
```
vRallev, line 117: I must use reflection for now.

zach-klippenstein (approved):

PR #145, opened 2020-10-18
Support the explicit API mode when generating Dagger factories

### LINKED ISSUE #144: Support the explicit API mode when generating Dagger factories

marcellogalhardo:
Anvil uses KotlinPoet to generate the Dagger factories in which in the [latest version 1.7.0](https://github.com/square/kotlinpoet/blob/master/docs/changelog.md): _"Generated code is now compatible with the explicit API mode by default."_.

This PR:
- Updates [KotlinPoet to 1.7.2](https://github.com/square/kotlinpoet/releases/tag/1.7.2), as it is required due to the error: [Cannot use KotlinPoet version `1.7.0` with Java 8](https://github.com/square/kotlinpoet/issues/999).
- Updates the sample app to use explicit API mode in the `:sample:scopes` and `:sample:library` modules to show its usage, `:sample:app` module is untouched as it is not a "library" module. If this is not necessary, I can revert the commit.

Fixes #144.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -70,7 +70,7 @@ ext {
           stdlib: "org.jetbrains.kotlin:kotlin-stdlib-jdk8:$kotlinVersion",
       ],
 
-      kotlinpoet: "com.squareup:kotlinpoet:1.6.0",
+      kotlinpoet: "com.squareup:kotlinpoet:1.7.2",
 
       truth: "com.google.truth:truth:1.0.1",
   ]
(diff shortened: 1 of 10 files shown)


marcellogalhardo:
Updated the PR to use KotlinPoet 1.7.2.

vRallev (changes requested):
I'm so sorry for the delay. This PR is great. Could I ask you to enable the explicit API mode in the `:integration-tests:*` module rather than the sample? The sample is supposed to be closer to the real world and many projects won't enable the explicit API mode. On the other side, that's exactly the purpose of the integration tests.

marcellogalhardo:
Hey @vRallev - no worries. Right, I dropped the changes on `:sample` and converted all modules from `:integration_tests` to Explicit API Mode as you requested. Thanks for the review!

vRallev (approved):

PR #151, opened 2020-10-24
[#150] Unconventional package names generate correct factories

### LINKED ISSUE #150: Dagger factory - Package name with a sentence case segment results in the wrong import in the generated factory
A dagger module like this
```
        package com.squareup.test
        
        import dagger.Module
        import dagger.Provides
        import com.uppercase.Package.Thing
        
        @Module
        object DaggerModule1 {
          @Provides fun provideThing(): Thing = Thing()
        }
```

generates a factory with the import to the `Package`, assuming that `Thing` is a nested class:

```
package com.squareup.test

import com.uppercase.Package
import dagger.internal.Factory
import dagger.internal.Preconditions
import kotlin.jvm.JvmStatic

object DaggerModule1_ProvideThingFactory : Factory<Package.Thing> {
  override fun get(): Package.Thing = provideThing()

  @JvmStatic
  fun create(): DaggerModule1_ProvideThingFactory = this

  @JvmStatic
  fun provideThing(): Package.Thing = Preconditions.checkNotNull(DaggerModule1.provideThing(),
      "Cannot return null from a non-@Nullable @Provides method")
}
```

This factory does not compile:
```
v: Configuring the compilation environment
e: /var/folders/34/3hpdfjvj7453gz767yc7b6j40000gn/T/Kotlin-Compilation1676785146484626793/build/anvil/com/squareup/test/DaggerModule1_ProvideThingFactory.kt: (5, 22): Packages cannot be imported
e: /var/folders/34/3hpdfjvj7453gz767yc7b6j40000gn/T/Kotlin-Compilation1676785146484626793/build/anvil/com/squareup/test/DaggerModule1_ProvideThingFactory.kt: (10, 60): Unresolved reference: Thing
e: /var/folders/34/3hpdfjvj7453gz767yc7b6j40000gn/T/Kotlin-Compilation1676785146484626793/build/anvil/com/squareup/test/DaggerModule1_ProvideThingFactory.kt: (11, 31): Unresolved reference: Thing
e: /var/folders/34/3hpdfjvj7453gz767yc7b6j40000gn/T/Kotlin-Compilation1676785146484626793/build/anvil/com/squareup/test/DaggerModule1_ProvideThingFactory.kt: (17, 31): Unresolved reference: Thing
```

Unfortunately the unconventional package is a 3rd party library that in general does not follow java conventions that are not explicitly enforced. 

From the exploration, the root cause is the same as in https://github.com/square/kotlinpoet/issues/978 - `ClassName.bestGuess` does not have enough context to make the correct guess and can only rely on conventions .

vlad-kasatkin:
Fixes #150 

I decided to remove the `ClassName.bestGuess` instead of using it as a fallback because I could not think of a case where this fallback would be required and be reliable enough for it to produce something useful. However, I don't think I understand the full picture and might be missing something important.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
@@ -62,13 +62,15 @@ internal fun ClassDescriptor.asClassName(): ClassName =
   )
 
 internal fun FqName.asClassName(module: ModuleDescriptor): ClassName {
-  try {
+  val segments = pathSegments().map { it.asString() }
+
+  // If the first sentence case is not the last segment of the path it becomes ambiguous,
+  // for example, com.Foo.Bar could be a inner class Bar or an unconventional package com.Foo.
+  val canGuessClassName = segments.indexOfFirst { it[0].isUpperCase() } == segments.size - 1
+  if (canGuessClassName) {
     return ClassName.bestGuess(asString())
-  } catch (ignored: IllegalArgumentException) {
-    // Probably lowercase class. Try to resolve the class below.
   }
 
-  val segments = pathSegments().map { it.asString() }
   for (index in (segments.size - 1) downTo 1) {
     val packageSegments = segments.subList(0, index)
     val classSegments = segments.subList(index, segments.size)
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
@@ -2,6 +2,9 @@ package com.squareup.anvil.compiler.dagger
 
 import com.google.common.truth.Truth.assertThat
 import com.squareup.anvil.compiler.componentInterfaceAnvilModule
+import com.squareup.anvil.compiler.dagger.UppercasePackage.OuterClass.InnerClass
+import com.squareup.anvil.compiler.dagger.UppercasePackage.TestClassInUppercasePackage
+import com.squareup.anvil.compiler.dagger.UppercasePackage.lowerCaseClassInUppercasePackage
 import com.squareup.anvil.compiler.daggerModule1
 import com.squareup.anvil.compiler.innerModule
 import com.squareup.anvil.compiler.isStatic
@@ -2231,7 +2234,7 @@ public final class DaggerComponentInterface implements ComponentInterface {
   @Test
   fun `a factory class is generated for a method returning a class with a named import`() {
     compile(
-      """
+        """
         package com.squareup.test
         
         import dagger.Module
@@ -2254,7 +2257,7 @@ public final class DaggerComponentInterface implements ComponentInterface {
   @Test
   fun `a factory class is generated ignoring the named import original path`() {
     compile(
-      """
+        """
         package com.squareup.test
         
         import dagger.Module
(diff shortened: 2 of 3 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/TestClassInUpperCasePackage.kt
```diff
@@ -0,0 +1,4 @@
+@file:Suppress("PackageDirectoryMismatch", "PackageName")
+
+package com.uppercase.Package
+class TestClassInUpperCasePackage
```
vRallev, line 4: Nit: Add a space between package and class and comment like `// These classes are used in some of the unit tests. Don't touch them.`. Even though it's unconventional, please put the class into the right directory.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -62,12 +62,6 @@ internal fun ClassDescriptor.asClassName(): ClassName =
   )
 
 internal fun FqName.asClassName(module: ModuleDescriptor): ClassName {
-  try {
-    return ClassName.bestGuess(asString())
-  } catch (ignored: IllegalArgumentException) {
-    // Probably lowercase class. Try to resolve the class below.
-  }
```
vRallev, line 69: I need to benchmark this in our repository before merging. The code below tries to resolve the class, which is more overhead and could slow down compile times.

vRallev (changes requested):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2231,7 +2234,7 @@ public final class DaggerComponentInterface implements ComponentInterface {
   @Test
   fun `a factory class is generated for a method returning a class with a named import`() {
     compile(
-      """
+        """
```
vlad-kasatkin, line 2237: was it not per style guide? seems like it is now

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -62,13 +62,15 @@ internal fun ClassDescriptor.asClassName(): ClassName =
...
+  // com.Foo.Bar could be a inner class Bar or an unconventional package com.Foo
+  val canGuessClassName = segments.indexOfFirst { it[0].isUpperCase() } == segments.size - 1
+  if (canGuessClassName) {
     return ClassName.bestGuess(asString())
-  } catch (ignored: IllegalArgumentException) {
-    // Probably lowercase class. Try to resolve the class below.
   }
 
-  val segments = pathSegments().map { it.asString() }
   for (index in (segments.size - 1) downTo 1) {
```
vlad-kasatkin, line 74: `downTo 1` does not work well with the default package nested classes, but dagger compiler does not handle them either, so I did not bother fixing.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoet.kt
```diff
@@ -62,13 +62,15 @@ internal fun ClassDescriptor.asClassName(): ClassName =
   )
 
 internal fun FqName.asClassName(module: ModuleDescriptor): ClassName {
-  try {
+  val segments = pathSegments().map { it.asString() }
+
+  // if the first sentence case is not the last segment it gets ambiguous:
+  // com.Foo.Bar could be a inner class Bar or an unconventional package com.Foo
```
vRallev, line 68: Comments are full sentences with an uppercase letter at the beginning and a dot at the end.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
vRallev: Oh, yes, looks they were wrong. Thanks.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2276,12 +2279,12 @@ public final class DaggerComponentInterface implements ComponentInterface {
...
       val factoryInstance = staticMethods.single { it.name == "create" }
-        .invoke(null) as Factory<Any>
+          .invoke(null) as Factory<Any>
       assertThat(factoryInstance::class.java).isEqualTo(factoryClass)
 
       val providedContributingObject = staticMethods
-        .single { it.name == "provideDate" }
-        .invoke(null)
+          .single { it.name == "provideDate" }
+          .invoke(null)
```
vRallev, line 2287: The indent is off here and above. It should be 2 space. Did you select the Square code style?

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2309,6 +2312,94 @@ public final class DaggerComponentInterface implements ComponentInterface {
...
+
+      val constructor = factoryClass.declaredConstructors.single()
+      assertThat(constructor.parameterTypes.toList()).isEmpty()
+
+      val staticMethods = factoryClass.declaredMethods.filter { it.isStatic }
+
+      val thingProvider = staticMethods.single { it.name == "provideThing" }
+      assertThat(thingProvider.invoke(null)).isInstanceOf(
+          lowerCaseClassInUppercasePackage::class.java
+      )
```
vRallev, line 2371: Super tiny nit: 
```kotlin
assertThat(thingProvider.invoke(null))
  .isInstanceOf(lowerCaseClassInUppercasePackage::class.java)
```

vRallev (changes requested):

vRallev:
I like the new solution. That's a good compromise.

PR #153, opened 2020-10-29
Add multibinding variants of @ContributesBinding

### LINKED ISSUE #152: Multibinding with @ContributesBinding
I'd like to propose a(couple) multibinding version(s) of @ContributesBinding to reduce the need for obvious, hand-written modules.

I often use multibindings in tandem with factories to enable constructor injection for types the Android framework is ultimately responsible for instantiating.

## Map Multibinding
```kotlin
@ContributesBindingIntoMap(AppScope::class, mapKey = PlaylistViewModel::class)
class PlaylistViewModel @Inject constructor(streamer: AudioStreamer): ViewModel

// resulting in

[...]
interface Module {
    @Binds
    @IntoMap
    @Named("com.example.PlaylistViewModel")
    fun providesComExamplePlaylistViewModel(viewModel: PlaylistViewModel): ViewModel
```

## Set Multibinding
```kotlin
@ContributesBindingIntoSet(AppScope::class)
class IndexMigration @Inject constructor(currentUser: User): Migration

// resulting in

interface Module {
    @Binds
    @IntoSet
    fun providesComExampleIndexMigration(migration: IndexMigration): Migration
```

## Alternative: Modify the existing ContributesBinding annotation

```kotlin
@Target(CLASS)
@Retention(RUNTIME)
annotation class ContributesBinding(
  val scope: KClass<*>,
  val boundType: KClass<*> = Unit::class,
  val mapKey: KClass<*> = Unit::class,
  val intoSet: Boolean = false
)
```

This would minimize the Anvil API surface, but it doesn't provide a strong API. For example, there's nothing preventing a person from suggesting they would like map _and_ set multibinding.

I have a rough prototype running as well, with a few issues in the map multibinding scenario.

jzbrooks:
Resolves #152

### DIFF
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBindingToMap.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBindingToMap.kt
@@ -1,0 +1,14 @@
+package com.squareup.anvil.annotations
+
+import kotlin.annotation.AnnotationRetention.RUNTIME
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.reflect.KClass
+
+@Target(CLASS)
+@Retention(RUNTIME)
+annotation class ContributesBindingToMap(
+    val scope: KClass<*>,
+    val key: KClass<*>,
+    val boundType: KClass<*> = Unit::class,
+    val replaces: Array<KClass<*>> = []
+)
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBindingToMap.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBindingToMap.kt
@@ -7,8 +7,8 @@ import kotlin.reflect.KClass
 @Target(CLASS)
 @Retention(RUNTIME)
 annotation class ContributesBindingToMap(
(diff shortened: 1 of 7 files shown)


annotations/src/main/java/com/squareup/anvil/annotations/ContributesBindingToMap.kt
```diff
@@ -0,0 +1,14 @@
...
+
+import kotlin.annotation.AnnotationRetention.RUNTIME
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.reflect.KClass
+
+@Target(CLASS)
+@Retention(RUNTIME)
+annotation class ContributesBindingToMap(
+  val scope: KClass<*>,
+  val key: KClass<*>,
```
jzbrooks, line 11: @vRallev, would it make sense to you to allow an arbitrary annotation here? It might lose a little in the way of type specificity, but validation wouldn't be too hard in the plugin I don't think. The plugin would just need to make sure the annotation meets dagger qualifier requirements (maybe those are as simple as "has at least one property").

vRallev:
I'm closing this in favor of #216

PR #156, opened 2020-11-11
Use correct package for import

### LINKED ISSUE #154: Upper case functions not compiling
Having upper case factory-style function names cause compilation errors.
I created a quick example here:
https://github.com/PaulWoitaschek/anvil/commit/1d443c1724314f5999ed5acdcec8739cb192a875

The stacktrace is as follows:
```console
➜  anvil git:(anvil_name_hickup) ./gradlew sample:library:assemble
Watching the file system is an incubating feature.
Received 15 file system events since last build
Virtual file system retained information about 5291 files, 1326 directories and 38 missing files since last build
Configuration on demand is an incubating feature.

> Configure project :
Versions: [Kotlin:1.3.72, Gradle:6.6.1, Kotlin use IR:false, Kotlin IR supported:false]

> Task :sample:library:compileKotlin FAILED
e: /home/paul/repos/anvil/sample/library/build/anvil/src-gen-compileKotlin/com/squareup/anvil/sample/hello/HelloModule_UserFactory.kt: (10, 49): Unresolved reference: User
e: /home/paul/repos/anvil/sample/library/build/anvil/src-gen-compileKotlin/com/squareup/anvil/sample/hello/HelloModule_UserFactory.kt: (11, 32): Unresolved reference: User
e: /home/paul/repos/anvil/sample/library/build/anvil/src-gen-compileKotlin/com/squareup/anvil/sample/hello/HelloModule_UserFactory.kt: (17, 22): Unresolved reference: User

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sample:library:compileKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.compilerRunner.GradleKotlinCompilerWork
   > Compilation error. See log for more details

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output. Run with --scan to get full insights.

* Get more help at https://help.gradle.org

Deprecated Gradle features were used in this build, making it incompatible with Gradle 7.0.
Use '--warning-mode all' to show the individual deprecation warnings.
See https://docs.gradle.org/6.6.1/userguide/command_line_interface.html#sec:command_line_warnings

BUILD FAILED in 733ms
14 actionable tasks: 1 executed, 13 up-to-date
Received 2 file system events for current build
Virtual file system retains information about 5294 files, 1334 directories and 38 missing files till next build
```

vRallev:
If a classname matches two imports, then try to resolve the type to find the correct one.

Fixes #154

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
@@ -239,18 +239,31 @@ internal fun PsiElement.requireFqName(
   // First look in the imports for the reference name. If the class is imported, then we know the
   // fully qualified name.
   importPaths
-      .filter { it.alias == null }
-      .firstOrNull {
-        it.fqName.shortName().asString() == classReference
+      .filter { it.alias == null && it.fqName.shortName().asString() == classReference }
+      .also { matchingImportPaths ->
+        when {
+          matchingImportPaths.size == 1 ->
+            return matchingImportPaths[0].fqName
+          matchingImportPaths.size > 1 ->
+            return matchingImportPaths.first { importPath ->
+              module.resolveClassByFqName(importPath.fqName, FROM_BACKEND) != null
+            }.fqName
+        }
       }
-      ?.let { return it.fqName }
 
   importPaths
-      .filter { it.alias == null }
-      .firstOrNull {
-        it.fqName.shortName().asString() == classReferenceOuter
+      .filter { it.alias == null && it.fqName.shortName().asString() == classReferenceOuter }
+      .also { matchingImportPaths ->
+        when {
+          matchingImportPaths.size == 1 ->
+            return FqName("${matchingImportPaths[0].fqName.parent()}.$classReference")
+          matchingImportPaths.size > 1 ->
+            return matchingImportPaths.first { importPath ->
+              val fqName = FqName("${importPath.fqName.parent()}.$classReference")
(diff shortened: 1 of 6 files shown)


pforhan (approved):

vlad-kasatkin (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -239,18 +239,31 @@ internal fun PsiElement.requireFqName(
...
-      .filter { it.alias == null }
-      .firstOrNull {
-        it.fqName.shortName().asString() == classReference
+      .filter { it.alias == null && it.fqName.shortName().asString() == classReference }
+      .also { matchingImportPaths ->
+        when {
+          matchingImportPaths.size == 1 ->
+            return matchingImportPaths[0].fqName
+          matchingImportPaths.size > 1 ->
+            return matchingImportPaths.first { importPath ->
```
cdeange, line 248: nit: would `.single` instead of `.first` make sense for the extra assertion that there should be one *and only one* correct import? (same question for line 261 below)

cdeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
vRallev: While true resolving types is expensive. In Kotlin you can't import classes with the same name from different packages (neither in Java). So we can assume that there is only one.

PR #157, opened 2020-11-12
Support property as providers in Dagger modules.

### LINKED ISSUE #149: Connot find symbol exception for properties provided in a dagger module in a seperate gradle module
In gradle module A:
``` kotlin
class Foo
@Module class MyModule {
    @get:Provides val foo = Foo()
}
```
In gradle module B which depends on A:
``` kotlin
@Component(modules = [MyModule::class]) interface MyComponent {
    fun foo(): Foo
}
```
```
build/generated/source/kapt/debug/com/package/package/DaggerMyComponent.java:5: error: cannot find symbol
import com.package.package.MyModule_GetFooFactory;
                          ^

### PR DESCRIPTION
**Pull request by vRallev**
Fixes #149

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -7,10 +7,12 @@ import com.squareup.anvil.compiler.codegen.addGeneratedByComment
 import com.squareup.anvil.compiler.codegen.asArgumentList
 import com.squareup.anvil.compiler.codegen.asClassName
 import com.squareup.anvil.compiler.codegen.classesAndInnerClasses
+import com.squareup.anvil.compiler.codegen.findAnnotation
 import com.squareup.anvil.compiler.codegen.functions
 import com.squareup.anvil.compiler.codegen.hasAnnotation
 import com.squareup.anvil.compiler.codegen.isNullable
 import com.squareup.anvil.compiler.codegen.mapToParameter
+import com.squareup.anvil.compiler.codegen.properties
 import com.squareup.anvil.compiler.codegen.requireFqName
 import com.squareup.anvil.compiler.codegen.requireTypeName
 import com.squareup.anvil.compiler.codegen.requireTypeReference
@@ -32,10 +34,11 @@ import com.squareup.kotlinpoet.jvm.jvmStatic
 import dagger.internal.Factory
 import dagger.internal.Preconditions
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.psi.KtCallableDeclaration
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtFile
-import org.jetbrains.kotlin.psi.KtNamedFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
+import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.psiUtil.parents
 import java.io.File
 import java.util.Locale.US
@@ -73,6 +76,17 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
               .forEach { function ->
                 generateFactoryClass(codeGenDir, module, clazz, function)
               }
+
+          clazz
+              .properties(includeCompanionObjects = true)
+              .asSequence()
+              .filter { property ->
+                // Must be '@get:Provides'.
+                property.findAnnotation(daggerProvidesFqName)?.useSiteTarget?.text == "get"
+              }
+              .forEach { property ->
+                generateFactoryClass(codeGenDir, module, clazz, property)
+              }
         }
   }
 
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt [hunk 1]
```diff
@@ -102,9 +121,81 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
+        isProperty = false
+    )
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  private fun generateFactoryClassForProperty(
+    codeGenDir: File,
+    module: ModuleDescriptor,
+    clazz: KtClassOrObject,
+    property: KtProperty
```
vlad-kasatkin, line 143: I wonder if instead of 2 functions that are nearly identical. you could get away with one that accepts a `functionOrProperty: KtCallableDeclaration` and then type checks if it's a `KtProperty`?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -102,9 +121,81 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
+    )
+  }
+
+  @OptIn(ExperimentalStdlibApi::class)
+  private fun generateFactoryClass(
+    codeGenDir: File,
+    clazz: KtClassOrObject,
+    packageName: String,
+    className: String,
+    functionName: String,
```
vlad-kasatkin, line 183: super-nit: no longer just a function name. Is there a word other than "statement" that describes both?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -71,13 +76,27 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
+
+          clazz
+              .properties(includeCompanionObjects = true)
+              .asSequence()
+              .filter { property ->
+                val annotation = property.findAnnotation(daggerProvidesFqName)
+                    ?: return@filter false
+
+                // Must be '@get:Provides'.
+                annotation.useSiteTarget?.text == "get"
```
cdeange, line 90: nit: could make this filter a single line:

```
// Must be '@get:Provides'.
property.findAnnotation(daggerProvidesFqName)?.useSiteTarget?.text == "get"
```

cdeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: I was looking at `function.valueParameters` and the decompiled sources don't show that this is an overridden method:
```kotlin
  @NotNull
  public List<KtParameter> getValueParameters() {
  }
```
So I assumed they don't share a common type, but they do. I can probably revert most of the changes, which is quite nice :) Good catch!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: Oh, yes.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: Hm, the common name would be `callableDeclaration`. I'll still with `function` for now, because it makes it easier to read and understand the code for me. I'll add a comment, though.

vRallev: I've written that and immediately changed my mind. `declaration` it is.

vlad-kasatkin (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2440,6 +2440,358 @@ public final class DaggerComponentInterface implements ComponentInterface {
...
+    compile(
+        """
+        package com.squareup.test
+        
+        import dagger.Module
+        import dagger.Provides
+        
+        @Module
+        class DaggerModule1 {
+          @get:Provides val string: String = "abc"
```
zach-klippenstein, line 2489: Is the `get:` qualifier necessary? If so, what is the rationale for requiring it? I can't see an unqualified annotation having any other interpretation, so it seems like unnecessary boilerplate.

zach-klippenstein (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
vRallev: Yes, Dagger requires that and I want to stay compatible to Dagger as close as possible.

zach-klippenstein: Not sure I understand. Wouldn't this code not work anyway if you removed Anvil?

vRallev: Correct, with plain Dagger `@Provides` would NOT work and you need to use `@get:Provides`. Using Anvil for this should be a 1 to 1 replacement and I don't want them to diverge.

zach-klippenstein: Sorry, I meant – would code that is using Anvil continue to work if anvil were removed? If not, then I don't understand the reasoning for keeping this limitation that presumably only exists because Dagger isn't aware of Kotlin.

vRallev: Yes, if you only use this feature and you enable the Dagger annotation processor, then the code would still work. https://github.com/square/anvil#experimental-dagger-factory-generation

PR #168, opened 2020-11-30
Dagger Factories Only

vRallev:
Add a new flag to the Anvil extension so that Anvil will only generate Dagger factories and no other code. The benefit of this that the stub generating Kotlin task can be incremental.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
@@ -17,4 +17,12 @@ open class AnvilExtension {
    * By default this feature is disabled.
    */
   var generateDaggerFactories = false
+
+  /**
+   * There are occasions where consumers of Anvil are only interested in generating Dagger
+   * factories to speed up build times and don't want to make use of the other features. With this
+   * flag Anvil will only generate Dagger factories and skip all other steps. If this flag is set
+   * to `true`, then also [generateDaggerFactories] must be set to `true`.
+   */
+  var generateDaggerFactoriesOnly = false
 }
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -68,7 +68,7 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
   }
 
   override fun apply(target: Project) {
-    target.extensions.create("anvil", AnvilExtension::class.java)
+    val extension = target.extensions.create("anvil", AnvilExtension::class.java)
 
     val once = AtomicBoolean()
 
@@ -87,10 +87,10 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
     // issue. Also make sure to apply it only once. A module could accidentally apply the JVM and
     // Android Kotlin plugin.
     target.pluginManager.withPluginOnce("org.jetbrains.kotlin.android") {
-      realApply(target, true)
+      realApply(target, true, extension)
     }
     target.pluginManager.withPluginOnce("org.jetbrains.kotlin.jvm") {
-      realApply(target, false)
+      realApply(target, false, extension)
     }
 
     target.afterEvaluate {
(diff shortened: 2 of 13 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -145,29 +161,47 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
+      project.tasks
+          .withType(KaptGenerateStubsTask::class.java)
+          .configureEach { stubsTask ->
+            // Disable incremental compilation for the stub generating task. Trigger the compiler
+            // plugin if any dependencies in the compile classpath have changed. This will make sure
+            // that we pick up any change from a dependency when merging all the classes. Without
+            // this workaround we could make changes in any library, but these changes wouldn't be
+            // contributed to the Dagger graph, because incremental compilation tricked us.
+            stubsTask.doFirst {
+              stubsTask.incremental = false
```
ZacSweers, line 197: I believe this might also mess with build caching, why not do it in configureEach?

vRallev: That's the main reason I change this at execution time and not config time. I don't want to invalidate the config cache and it doesn't need to. When the task runs, because something has changed in the classpath, then the task should not be incremental. If the task doesn't, because nothing has changed, then it doesn't matter whether this flag is true or false. It's really not a big deal. 

I also want to toggle this flag as late as possible, because I've seen the Kotlin Gradle Plugin to change this flag again. Anvil needs to have the last say.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -17,4 +17,12 @@ open class AnvilExtension {
...
    */
   var generateDaggerFactories = false
+
+  /**
+   * There are occasions where consumers of Anvil are only interested in generating Dagger
+   * factories to speed up build times and don't want to make use of the other features. With this
+   * flag Anvil will only generate Dagger factories and skip all other steps. If this flag is set
+   * to `true`, then also [generateDaggerFactories] must be set to `true`.
+   */
+  var generateDaggerFactoriesOnly = false
```
vRallev, line 27: I'm not intending to document this flag in the README. Most people probably don't want to use this and I'm only aware of one project using it so far. We can always add it to the documentation when a higher demand is there.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -17,4 +17,12 @@ open class AnvilExtension {
    * By default this feature is disabled.
    */
   var generateDaggerFactories = false
+
+  /**
+   * There are occasions where consumers of Anvil are only interested in generating Dagger
+   * factories to speed up build times and don't want to make use of the other features. With this
+   * flag Anvil will only generate Dagger factories and skip all other steps. If this flag is set
```
FrostRocket, line 24: If it's a requirement that generateDaggerFactories has the same value, would it be possible to either add a test or throw some sort of error? Might be easy for someone to overlook this last part of the kdoc.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -100,22 +104,34 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
                 "'${target.path}'. Only Android and Java modules are supported for now."
         )
       }
+
+      if (!extension.generateDaggerFactories && extension.generateDaggerFactoriesOnly) {
```
FrostRocket, line 108: Given the kdoc in AnvilExtension.kt it's good to see the check here. 👍

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -100,22 +104,34 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
+        )
+      }
     }
   }
 
   private fun realApply(
     project: Project,
-    isAndroidProject: Boolean
+    isAndroidProject: Boolean,
+    extension: AnvilExtension
```
FrostRocket, line 120: Would it make sense to change the method signature to use the generateDaggerFactoriesOnly boolean instead? You don't seem to use anything else in the AnvilExtension class.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -145,29 +161,47 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
   @OptIn(ExperimentalStdlibApi::class)
   private fun disableIncrementalKotlinCompilation(
     project: Project,
-    isAndroidProject: Boolean
+    isAndroidProject: Boolean,
+    extension: AnvilExtension
```
FrostRocket, line 165: Here too (using the flag instead of extension directly).

FrostRocket (approved):
LGTM.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: I think it's okay the pass the container (the extension) object to the methods.

PR #176, opened 2020-12-28
Build Anvil with Kotlin 1.4.30-M1 in CI.

vRallev:


kirillzh (approved):

vRallev:
Three unit tests are failing due to https://youtrack.jetbrains.com/issue/KT-44130. Nothing critical, Anvil will still work with Kotlin `1.4.30`.

0legg (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -79,11 +79,11 @@ internal class CodeGenerationExtension(
...
-      bindingTrace.bindingContext,
-      module,
-      emptyList(),
-      listOf(codeGenDir),
-      true
+      bindingContext = bindingTrace.bindingContext,
+      moduleDescriptor = module,
+      additionalJavaRoots = emptyList(),
+      additionalKotlinRoots = listOf(codeGenDir),
+      addToEnvironment = true
```
vRallev, line 86: They've added another parameter with a default argument in between instead of at the end. Therefore I had to use named arguments.

PR #177, opened 2021-01-19
Enable the IR extension now that the JVM IR backend is in beta with Kotlin 1.4.30

vRallev:


vRallev:
This is a draft until Kotlin 1.4.30 stable has been released.

chrisjenx:
Oh just saw this after creating: #190 hopefully this fixes that :partying_face:

compiler/generate_build_properties.gradle
```diff
@@ -1,7 +1,7 @@
 def generatedDirPath = 'generated/sources/build-properties/kotlin/main'
 
 sourceSets {
-  main.java.srcDirs += "$buildDir/$generatedDirPath"
+  test.java.srcDirs += "$buildDir/$generatedDirPath"
```
vRallev, line 4: This generates the boolean flag `USE_IR`. I only need in tests and not the main source code anymore. We still run all tests with and without the new IR backend.

christiandeange (approved):

PR #180, opened 2021-01-20
Allow warnings in code locally.

vRallev:


christiandeange (approved):

zach-klippenstein:
I'm not sure what the reason is for this change, but I've found it useful to predicate this on the IDE, but still run them locally from the command line.

zach-klippenstein (approved):

vRallev:
> I'm not sure what the reason is for this change, but I've found it useful to predicate this on the IDE, but still run them locally from the command line. Eg https://github.com/square/radiography/blob/640dbb4ffc52d0e5baef0b1ec32eb471ed7ce717/build.gradle.kts#L75

I often build in the terminal and rely less often on the IDE.

Issue #165, opened 2020-11-30
Support AssistedInject


vRallev:
Support generating the code that the [AssistedInject](https://github.com/square/AssistedInject) annotation processor would generate.

mos8502:
With dagger recently introducing support for [AssitedInject](https://dagger.dev/dev-guide/assisted-injection) does this still make sense ?

vRallev:
Anvil only needs to support Dagger's assisted inject. The Square library is pretty much obsolete right now.

vRallev:
Created #182

PR #182, opened 2021-01-25
#165 Support Dagger's assisted inject and generate necessary code

vRallev:
Dagger 2.31 has built-in support for assisted injection. Dagger generates code for the classes using the new annotations. Anvil allows to replace the Dagger annotation processor in many modules and generates the necessary code. This PR adds support for assisted injection.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
@@ -227,13 +227,6 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
         .let { addType(it) }
     }
 
-    val directory = File(codeGenDir, packageName.replace('.', File.separatorChar))
-    val file = File(directory, "$className.kt")
-    check(file.parentFile.exists() || file.parentFile.mkdirs()) {
-      "Could not generate package directory: ${file.parentFile}"
-    }
-    file.writeText(content)
-
-    return GeneratedFile(file, content)
+    return createGeneratedFile(codeGenDir, packageName, className, content)
   }
 }
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
@@ -1,0 +1,239 @@
+package com.squareup.anvil.compiler.codegen.dagger
+
+import com.squareup.anvil.compiler.AnvilCompilationException
+import com.squareup.anvil.compiler.assistedFactoryFqName
+import com.squareup.anvil.compiler.assistedFqName
+import com.squareup.anvil.compiler.assistedInjectFqName
+import com.squareup.anvil.compiler.classDescriptorForType
+import com.squareup.anvil.compiler.codegen.CodeGenerator.GeneratedFile
+import com.squareup.anvil.compiler.codegen.PrivateCodeGenerator
+import com.squareup.anvil.compiler.codegen.asClassName
+import com.squareup.anvil.compiler.codegen.asTypeName
+import com.squareup.anvil.compiler.codegen.buildFile
+import com.squareup.anvil.compiler.codegen.classesAndInnerClasses
+import com.squareup.anvil.compiler.codegen.fqNameOrNull
+import com.squareup.anvil.compiler.codegen.hasAnnotation
+import com.squareup.anvil.compiler.codegen.requireFqName
+import com.squareup.anvil.compiler.generateClassName
+import com.squareup.kotlinpoet.ClassName
+import com.squareup.kotlinpoet.FileSpec
+import com.squareup.kotlinpoet.FunSpec
+import com.squareup.kotlinpoet.KModifier.OVERRIDE
+import com.squareup.kotlinpoet.KModifier.PRIVATE
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.PropertySpec
+import com.squareup.kotlinpoet.TypeName
+import com.squareup.kotlinpoet.TypeSpec
+import com.squareup.kotlinpoet.TypeVariableName
+import com.squareup.kotlinpoet.asClassName
+import com.squareup.kotlinpoet.jvm.jvmStatic
+import dagger.internal.InstanceFactory
+import org.jetbrains.kotlin.descriptors.DescriptorVisibilities.PROTECTED
+import org.jetbrains.kotlin.descriptors.DescriptorVisibilities.PUBLIC
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.descriptors.Modality.ABSTRACT
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.descriptors.resolveClassByFqName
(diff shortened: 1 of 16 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -0,0 +1,232 @@
...
+    projectFiles
+      .asSequence()
+      .flatMap { it.classesAndInnerClasses() }
+      .filter { it.hasAnnotation(assistedFactoryFqName) }
+      .forEach { clazz ->
+        generateFactoryClass(codeGenDir, module, clazz)
+      }
+  }
+
+  private fun generateFactoryClass(
```
FrostRocket, line 65: This method does a lot. I feel like it would be beneficial to try to split it up for unit test and readability purposes.

FrostRocket (approved):
There's a lot here to review, I'd get another pair of eyes on this one before you merge. Overall, flow looks good aside from the factory generator method being a bit verbose. To be fair, there's a lot of logic that it needs to do, though. It's up to you if you wish to take another stab at refactoring.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
vRallev: I tried to do that in the past and it didn't work well, because there are many variables that need to be shared. Most of it is also a giant builder. Unit tests also verify that the whole output matches the expectation for a given input. It's not worth to test single steps.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -0,0 +1,232 @@
...
+          FunSpec.builder("create")
+            .addModifiers(OVERRIDE)
+            .returns(returnType.asClassName().parameterized())
+            .apply {
+              functionParameters.forEach { parameter ->
+                addParameter(parameter.name.asString(), parameter.type.asTypeName())
+              }
+
+              val argumentList = functionParameters.joinToString { it.name.asString() }
+              addStatement("return delegateFactory.get($argumentList)")
```
christiandeange, line 202: Change to
```kotlin
addStatement("return %L.get($argumentList)", delegateFactoryName)
```

or simply:
```kotlin
addStatement("return ${delegateFactoryName}.get($argumentList)")
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -0,0 +1,232 @@
...
+                  "return %T.create(%T($delegateFactoryName))",
+                  InstanceFactory::class,
+                  implClass.parameterized()
+                )
+                .build()
+            )
+            .build()
+            .let {
+              addType(it)
+            }
```
christiandeange, line 224: super nit: you have a `.let { addType(it) }` below that is all on a single line, yet this one is multiline. Make them consistent.

_(Same in `AssistedInjectGenerator`)_

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedInjectGenerator.kt
```diff
@@ -0,0 +1,180 @@
...
+import com.squareup.kotlinpoet.TypeVariableName
+import com.squareup.kotlinpoet.jvm.jvmStatic
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.psi.KtConstructor
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.allConstructors
+import java.io.File
+
+internal class AssistedInjectGenerator : PrivateCodeGenerator() {
```
christiandeange, line 30: There's no comments in this file...may want to add some to document what's going on.

integration-tests/library/src/main/java/com/squareup/anvil/test/AssistedService.kt
```diff
@@ -0,0 +1,16 @@
+package com.squareup.anvil.test
+
+import dagger.assisted.Assisted
+import dagger.assisted.AssistedFactory
+import dagger.assisted.AssistedInject
+
+public class AssistedService @AssistedInject constructor(
+  public val integer: Int,
+  @Assisted public val string: String
```
christiandeange, line 9: nit: could you use a separate, custom class as the injected type? Something like even a `data class Name(val fullName: String)` might make the example a bit more clear, as injecting a raw `String` isn't a super-typical use case.

christiandeange (approved):
I have no experience with AssistedInject so I really am not familiar with the implementation much...to be frank I glossed over a large part of the implementation.

Happy to give a ✅ but if you want some more confidence, I would ask for another reviewer that is familiar with the tool you're trying to replace here.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -0,0 +1,232 @@
...
+    // Check for exact number of functions.
+    val function = when (functions.size) {
+      0 -> throw AnvilCompilationException(
+        "The @AssistedFactory-annotated type is missing an abstract, non-default method " +
+          "whose return type matches the assisted injection type.",
+        element = clazz
+      )
+      1 -> functions[0]
+      else -> throw AnvilCompilationException(
+        "The @AssistedFactory-annotated type should contain a single abstract, non-default " +
```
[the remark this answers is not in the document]
vRallev: This error message is coming from Dagger. I kept the same wording so that Anvil can be a one to one replacement.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -0,0 +1,232 @@
...
+    if (assistedParameters.size != functionParameters.size) {
+      throw AnvilCompilationException(
+        "The parameters of the factory method must be assignable to the list of @Assisted " +
+          "parameters in ${returnType.fqNameSafe}.",
+        element = clazz
+      )
+    }
+
+    functionParameters.forEachIndexed { index, parameter ->
+      val assistedType = assistedParameters[index].type
```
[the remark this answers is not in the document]
vRallev: I like that, but will keep it as is to avoid allocating extra objects.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
vRallev: The 2nd, good catch.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedInjectGenerator.kt
vRallev: Added KDoc.

integration-tests/library/src/main/java/com/squareup/anvil/test/AssistedService.kt
vRallev: This really just a tiny integration test and doesn't serve as a sample. An extra class is unnecessary and doesn't make it clearer IMO.

PR #186, opened 2021-02-01
Check more constraints

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -411,3 +415,18 @@ fun KtUserType.findExtendsBound(): List<FqName> {
     .typeParameters
     .mapNotNull { it.fqName }
 }
+
+fun KtClassOrObject.singleInjectConstructorOrNull(fqName: FqName): KtConstructor<*>? {
```
christiandeange, line 419: The naming is a bit odd, since Kotlin stdlib functions that end with `OrNull` aren't expected to throw an exception...perhaps some kdoc that clarifies the "returns null when 0 matching constructors found, throws when more than 1 found" behaviour would be enough

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -411,3 +415,18 @@ fun KtUserType.findExtendsBound(): List<FqName> {
...
+  val constructors = allConstructors.filter {
+    it.hasAnnotation(injectFqName) || it.hasAnnotation(assistedInjectFqName)
+  }
+
+  return when (constructors.size) {
+    0 -> null
+    1 -> if (constructors[0].hasAnnotation(fqName)) constructors[0] else null
```
christiandeange, line 426: Should you just add the `hasAnnotation(fqName)` to the `filter` predicate above?

```kotlin
val constructors = allConstructors.filter {
  it.hasAnnotation(fqName) &&
    (it.hasAnnotation(injectFqName) || it.hasAnnotation(assistedInjectFqName))
}
```

christiandeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
vRallev: That's the 2nd awkward thing about this method. It can either return a constructor for `@Inject` or `@AssistedInject`. A class is only allowed to have one constructor annotated with either of these annotations. Because I didn't want to repeat this method for both annotations, you pass in as argument which one you're looking for. But in the filter method above I always need to check both. 

I think I'll check the inputs that you only use one of these annotations. With the KDoc it should become clearer.

PR #191, opened 2021-02-03
A provider method inside of a companion object is not abstract. 

### LINKED ISSUE #187: False positive check leads to compilation fail
First of all thanks for quick support for dagger's AssistedInject! 

We did a migration to it with `2.0.12-SNAPSHOT` anvil version and everything worked fine, but with update to stable 2.0.12 compilation started failing with `@Provides methods cannot be abstract` error. 

Module failed to compile looks like this: 
```
@Module
@ContributesTo(AppComponent::class)
interface Module {

    @Binds
    @Singleton
    fun bindsDependency(i: DependencyImpl): Dependency

    companion object {
        @Provides
        @Singleton
        fun providesDepB(depA: DepA): DepB = DepB(depA)
    }
}
```
This code is valid with both dagger kapt and anvil 2.0.11. 

It looks like it was introduced in #186  https://github.com/square/anvil/pull/186/files#diff-c59935dc466bb8ab8c74b8e866de744af48eeddaa9340add1833bb09776ff354R63
and `@Provides` function should be filtered if it's inside companion object

vRallev:
This was a false positive.

Fixes #187

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -38,6 +38,7 @@ import org.jetbrains.kotlin.lexer.KtTokens.ABSTRACT_KEYWORD
 import org.jetbrains.kotlin.psi.KtCallableDeclaration
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtFunction
 import org.jetbrains.kotlin.psi.KtObjectDeclaration
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.psi.psiUtil.parents
@@ -61,12 +62,7 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
           .asSequence()
           .filter { it.hasAnnotation(daggerProvidesFqName) }
           .onEach { function ->
-            if (clazz.isInterface() || function.hasModifier(ABSTRACT_KEYWORD)) {
-              throw AnvilCompilationException(
-                "@Provides methods cannot be abstract",
-                element = function
-              )
-            }
+            checkFunctionIsNotAbstract(clazz, function)
           }
           .also { functions ->
             // Check for duplicate function names.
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -293,4 +289,28 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
 
     return createGeneratedFile(codeGenDir, packageName, className, content)
   }
+
+  private fun checkFunctionIsNotAbstract(
+    clazz: KtClassOrObject,
+    function: KtFunction
+  ) {
+    fun fail(): Nothing = throw AnvilCompilationException(
```
FrostRocket, line 297: Nifty usage of internal functions.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -293,4 +289,28 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
+      "@Provides methods cannot be abstract",
+      element = function
+    )
+
+    // If the function is abstract, then it's an error.
+    if (function.hasModifier(ABSTRACT_KEYWORD)) fail()
```
FrostRocket, line 303: This method has a bit of code smell because of how the fail states operate as sort of a state machine. Would it be better to split out each of these checks into it's own method instead? Because checkFunctionIsNotAbstract doesn't really capture all that's going on here.

FrostRocket: Splitting might not be the best either. The flow just seems odd. in it's current form.

FrostRocket (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: It's not a state machine but rather a control flow. I don't think splitting would bring any value. I also believe that the method name describes well what's going on. All functions in interfaces are abstract by default and all functions with the `abstract` keyword are abstract. This method checks for both of these two cases ¯\_(ツ)_/¯

PR #207, opened 2021-02-21
Add support for registering generated dirs to sourceSets

ZacSweers:
This way they're visible for debugging, jumping through generated code, etc. Enabled by default per our prior offline chat

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -64,6 +64,12 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
     val extension = project.extensions.findByType(AnvilExtension::class.java)
       ?: project.objects.newInstance(AnvilExtension::class.java)
 
+    if (extension.addGeneratedDirToSourceSets.get()) {
+      kotlinCompilation.defaultSourceSet {
+        kotlin.srcDir(srcGenDir)
+      }
+    }
+
     return project.provider {
       listOf(
         FilesSubpluginOption(
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
@@ -43,6 +43,14 @@ abstract class AnvilExtension @Inject constructor(objects: ObjectFactory) {
   val disableComponentMerging: Property<Boolean> = objects.property(Boolean::class.java)
     .convention(false)
 
+  /**
+   * Add Anvil generated source directories to sourceSets in Gradle for indexing visibility. This
+   * can be useful in debugging and is enabled by default but can be disabled if you don't
+   * need/want this.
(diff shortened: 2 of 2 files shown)


ZacSweers:
CI failure looks unrelated?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -55,6 +55,12 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
     )
     val extension = project.extensions.findByType(AnvilExtension::class.java) ?: AnvilExtension()
 
+    if (extension.addGeneratedDirToSourceSets) {
+      kotlinCompilation.defaultSourceSet {
+        kotlin.srcDir(srcGenDir)
+      }
+    }
```
vRallev, line 62: Ha, nice

vRallev (approved):

vRallev:
> CI failure looks unrelated?

No, it looks like a legit failure. I can reproduce it locally with `./gradlew :sample:app:assDeb -Psquare.useIR=true`.

vRallev:
I filed: https://youtrack.jetbrains.com/issue/KT-45100

ZacSweers:
I'll plan to rekick this after 1.5.0 is out

ZacSweers:
Updated but will need to wait for #259 to merge first

vRallev:
The failure is legit. I'll have a look next week. I'm also seeing build cache misses. All of these issues are gone when I set the flag to false :-/

vRallev:
Broken incremental builds are unrelated to this change and fixed in #294. 

However, there's still one issue with incremental builds. [This](https://scans.gradle.com/s/ggrszvzqxowmu) is a full build after clean of the sample project. In the src gen directory are no sources before the build. After the build there are sources and therefore an [incremental build after that](https://scans.gradle.com/s/k7p7j3mcv5vgu) is not up-to-date. The [third build](https://scans.gradle.com/s/xljjv644b47du/timeline?details=j62m2xx52wwxy) is fully up-to-date.

This is kinda a deal breaker for me to set the new setting to true by default. I wonder if we could add the directory to the source set, but exclude it from the up-to-date check? @ZacSweers The alternative is to the set the new option to false by default and you explicitly need to opt in.

ZacSweers:
After trying this in our codebase manually, I think we saw similar issues and I think this needs to be addressed in KGP first. I'd be in favor of holding off on this for now and revisiting later I can file an issue on youtrack as well

vRallev:
I tested this more with the tiny sample project and saw random compile errors. I'm unsure why this is exactly. But while debugging I realized that Anvil will process the generated sources during incremental builds twice. During the very first code generation phase it'll pick up the previously generated files, generate new files and process the new files again. This can be dangerous, if the old files are stale and actually will change. This issue can be avoided by filtering files in the src-gen dir, but the internal class scanner might pick them up nonetheless.

Considering all the issues I don't plan to merge this PR even with the default value set to false. It's too easy to shoot yourself into the foot. There should be a better way to add directories as source dirs to the IDE without affecting the compile classpath for Gradle.

PR #208, opened 2021-02-22
Use the mangled function name to generate the factory for a provider method.

vRallev:
If a provider function's name is mangled (usually when the function has internal visibility), then use the mangled name to generate the factory.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -21,6 +21,7 @@ import com.squareup.anvil.compiler.codegen.withJvmSuppressWildcardsIfNeeded
 import com.squareup.anvil.compiler.daggerModuleFqName
 import com.squareup.anvil.compiler.daggerProvidesFqName
 import com.squareup.anvil.compiler.generateClassName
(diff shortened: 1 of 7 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -110,6 +112,9 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
 
     val isProperty = declaration is KtProperty
 
+    val isMangled = !isProperty &&
+      declaration.visibilityModifierTypeOrDefault() == INTERNAL_KEYWORD
```
christiandeange, line 116: If the declaration is also marked as `@PublishedApi` then I _believe_ the name isn't mangled. May want to add a test for that just to make sure.

FrostRocket (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
vRallev: Nice, pretty sure that the test case for that will fail at the moment. I couldn't find an API in the compiler to get the mangled function name, only for inline classes. This solution here is so fragile.

christiandeange (approved):

PR #210, opened 2021-02-25
Use Kotlin Poet for generating the hints. 

vRallev:
This is a clean up PR. I believe this way the generated code is easier to read, because proper imports are used.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -42,7 +42,6 @@ import org.jetbrains.kotlin.incremental.components.NoLookupLocation
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
-import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtFile
 import org.jetbrains.kotlin.psi.KtProperty
 import org.jetbrains.kotlin.resolve.DescriptorUtils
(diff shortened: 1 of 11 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -294,23 +293,21 @@ internal class BindingModuleGenerator(
...
-          .toList()
+        it.findChildrenByClass(KtProperty::class.java).toList()
       }
       .mapNotNull { ktProperty ->
-        (
-          (ktProperty.initializer as? KtClassLiteralExpression)
-            ?.firstChild as? KtDotQualifiedExpression
-          )
-          ?.text
-          ?.let { FqName(it) }
```
vRallev, line 313: This wasn't great, it was pretty much hardcoded that a fully qualified name is used for `val prop = abc.def.Class::class`.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesBindingGenerator.kt
```diff
@@ -42,35 +48,44 @@ internal class ContributesBindingGenerator : CodeGenerator {
...
-            '_'
-          )}$SCOPE_SUFFIX: kotlin.reflect.KClass<$scope> = $scope::class
-          """.trimIndent()
-        file.writeText(content)
+          FileSpec.buildFile(generatedPackage, propertyName) {
+            addProperty(
+              PropertySpec
+                .builder(
+                  name = propertyName + REFERENCE_SUFFIX,
+                  type = KClass::class.asClassName().parameterizedBy(className)
```
christiandeange, line 64: nice

christiandeange (approved):

FrostRocket (approved):
Nice cleanup!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesBindingGenerator.kt
FrostRocket: Chain is a lot easier to follow 👍

PR #212, opened 2021-02-26
Upgrade Dagger 

vRallev:
And add a test case for an assisted injection scenario that previously failed.

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
@@ -4,9 +4,11 @@ import com.google.common.truth.Truth.assertThat
 import com.squareup.anvil.compiler.assistedService
 import com.squareup.anvil.compiler.assistedServiceFactory
 import com.squareup.anvil.compiler.createInstance
+import com.squareup.anvil.compiler.daggerModule1
 import com.squareup.anvil.compiler.factoryClass
 import com.squareup.anvil.compiler.implClass
 import com.squareup.anvil.compiler.isStatic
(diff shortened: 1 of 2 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
```diff
@@ -1061,6 +1063,54 @@ public final class AssistedServiceFactory_Impl implements AssistedServiceFactory
...
+      )
+      
+      @AssistedFactory
+      interface AssistedServiceFactory {
+        fun create(string: String): AssistedService
+      }
+      
+      @Module
+      object DaggerModule1 {
+        @Provides @Named("") fun provideService(): AssistedService = AssistedService(5, "Hello")
```
vRallev, line 1090: That provider method previously triggered an error in Dagger. Without the qualifier the same error happens. Anvil never had this restriction ¯\\_(ツ)_/¯ 

https://github.com/google/dagger/issues/2370

dependencies.gradle
```diff
@@ -5,7 +5,7 @@ ext {
   daggerVersion = '2.33'
   espressoVersion = '3.2.0'
   kotlinVersion = project.hasProperty('square.kotlinVersion') ?
-      project.getProperty('square.kotlinVersion') : '1.4.30'
+      project.getProperty('square.kotlinVersion') : '1.4.31'
```
christiandeange, line 8: nit: commit message says this updates dagger but this commit only updates the version of kotlin.

christiandeange (approved):

FrostRocket (approved):

dependencies.gradle
vRallev: It's actually a 2nd commit that I squeezed in 🙈

PR #215, opened 2021-03-02
Add error message for missing scope

seanfreiburg:
While debugging an issue, I was unable to find the module this scope threw an exception for, so I added an error message to make it easier to find.

I tried to add a test, but I couldn't figure out where to place it, let me know if any changes are needed.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
@@ -136,7 +136,9 @@ internal fun AnnotationDescriptor.getAnnotationValue(key: String): ConstantValue
   allValueArguments[Name.identifier(key)]
 
 internal fun AnnotationDescriptor.scope(module: ModuleDescriptor): ClassDescriptor {
-  val kClassValue = requireNotNull(getAnnotationValue("scope")) as KClassValue
+  val annotationValue = getAnnotationValue("scope")
+    ?: throw AnvilCompilationException(message = "Couldn't find scope for $fqName")
+  val kClassValue = annotationValue as KClassValue
   return kClassValue.argumentType(module).classDescriptorForType()
 }
 
--- a/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
@@ -136,10 +136,13 @@ internal fun AnnotationDescriptor.getAnnotationValue(key: String): ConstantValue
   allValueArguments[Name.identifier(key)]
 
 internal fun AnnotationDescriptor.scope(module: ModuleDescriptor): ClassDescriptor {
-  val annotationValue = getAnnotationValue("scope")
-    ?: throw AnvilCompilationException(message = "Couldn't find scope for $fqName")
-  val kClassValue = annotationValue as KClassValue
-  return kClassValue.argumentType(module).classDescriptorForType()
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -136,7 +136,9 @@ internal fun AnnotationDescriptor.getAnnotationValue(key: String): ConstantValue
   allValueArguments[Name.identifier(key)]
 
 internal fun AnnotationDescriptor.scope(module: ModuleDescriptor): ClassDescriptor {
-  val kClassValue = requireNotNull(getAnnotationValue("scope")) as KClassValue
+  val kClassValue = requireNotNull(
+    getAnnotationValue("scope")
+  ) { "Couldn't find scope for ${module.name}" } as KClassValue
```
vRallev, line 141: The formatting is a little weird. I think 
```kotlin
requireNotNull(getAnnotationValue("scope")) { 
  "Couldn't find scope for ${module.name}" 
} as KClassValue
```
works better with our codestyle.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -136,7 +136,9 @@ internal fun AnnotationDescriptor.getAnnotationValue(key: String): ConstantValue
   allValueArguments[Name.identifier(key)]
 
 internal fun AnnotationDescriptor.scope(module: ModuleDescriptor): ClassDescriptor {
-  val kClassValue = requireNotNull(getAnnotationValue("scope")) as KClassValue
+  val kClassValue = requireNotNull(getAnnotationValue("scope")) {
+    "Couldn't find scope for ${module.name}"
```
vRallev, line 140: I suggest to add a test for that. I don't believe `module.name` is what you expect. It's also weird that this is needed in the first place. Declaring the scope is required, otherwise there's a compile error:

image

```
e: com.squareup.anvil.compiler.AnvilCompilationException: Back-end (JVM) Internal error: The first argument for com.squareup.anvil.annotations.ContributesTo must be a class literal: @Module
@ContributesTo
public object AppModule2 {
  @Provides @Singleton public fun provideFunction(): (String) -> Int = { it.length }
  @Provides internal fun provideType(): CharSequence = "Hello"
}
File being compiled: (12,1) in /Users/ralf/Development/projects/square/anvil/integration-tests/library/src/main/java/com/squareup/anvil/test/Modules.kt
```

I guess Anvil runs before normal Kotlin compilation and then this error is thrown. A test will help to find out where the ambiguous error is thrown. I don't know in which exact scenario, but once you add the test I can point you to the best location.

vRallev (changes requested):

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
seanfreiburg: Yeah, I think that is the scenario I'm seeing, anvil fails before kotlin runs, and it was on a big rebase, so I had no idea where to look. I'll work on a test.

vRallev: Thanks. There are a few examples already. I also suggest to use `AnvilCompilationException` rather than `requireNotNull`, because then you get better error messages. I was lazy in this code snippet in the past 🙃

vRallev:
I made some slight changes to report the source element in the error message. The rest looks good!

vRallev (approved):

PR #216, opened 2021-03-02
Generate methods for contributed multibindings

### LINKED ISSUE #152: Multibinding with @ContributesBinding

vRallev:
This will partially resolve #152. This PR adds support for generating multibindings into sets. Generating multibindings for maps will be a separate PR #219. 

This PR is quite large and broken down into smaller commits. There are plenty of tests, which "contributeTo" (Do you see what I did here, yes, I'm very funny) to the size of the PR. In single commits you my find a TODO that is resolved in a later commit or a follow-up PR in #217 or #218. 

From the consumer side contributing multibindings is very similar to bindings and both annotations can be used in conjunction. Qualifiers are supported as well.
```kotlin
@ContributesBinding(SubScope::class)
@ContributesMultibinding(SubScope::class, ignoreQualifier = true)
@Named("middle")
object SomeType : Parent
```

### DIFF
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesMultibinding.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesMultibinding.kt
@@ -51,7 +51,7 @@ import kotlin.reflect.KClass
  * ```
  *
  * Contributed multibindings can replace other contributed modules and contributed multibindings
- * with the [replaces] parameter. This is especially helpful for different bindings in
+ * with the [replaces] parameter. This is especially helpful for different multibindings in
  * tests.
  * ```
  * @ContributesMultibinding(
@@ -69,10 +69,14 @@ import kotlin.reflect.KClass
  * `@Provides @IntoSet` method returning [boundType].
  *
  * [ContributesMultibinding] can be used in conjunction with [ContributesBinding]. Each annotation
- * will generate the respective binding or multibindings method.
+ * will generate the respective binding or multibindings method. If a qualifier is only meant for
+ * one of the annotations, then you can set [ignoreQualifier] to `true` and the qualifier won't
+ * be added to the generated multibinding method.
  * ```
  * @ContributesBinding(AppScope::class)
- * @ContributesMultibinding(AppScope::class)
+ * @ContributesMultibinding(AppScope::class, ignoreQualifier = true)
+ * @Named("Prod")
+ * object MainListener : Listener
  * object MainListener : Listener
  * ```
  */
@@ -81,5 +85,6 @@ import kotlin.reflect.KClass
 public annotation class ContributesMultibinding(
   val scope: KClass<*>,
   val boundType: KClass<*> = Unit::class,
-  val replaces: Array<KClass<*>> = []
+  val replaces: Array<KClass<*>> = [],
+  val ignoreQualifier: Boolean = false
 )
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesMultibinding.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesMultibinding.kt
@@ -1,0 +1,85 @@
+package com.squareup.anvil.annotations
+
+import kotlin.annotation.AnnotationRetention.RUNTIME
+import kotlin.annotation.AnnotationTarget.CLASS
+import kotlin.reflect.KClass
+
+/**
+ * Generate a Dagger multibinding method for an annotated class and contributes this multibinding
+ * method to the given [scope]. Imagine this example:
+ * ```
+ * interface Listener
+ *
+ * class MainListener @Inject constructor() : Listener
+ *
+ * @Module
+ * @ContributesTo(AppScope::class)
+ * abstract class MainListenerModule {
+ *   @Binds @IntoSet
+ *   abstract fun bindMainListener(listener: MainListener): Listener
+ * }
+ * ```
+ * This is a lot of boilerplate. You can replace this entire module with the
+ * [ContributesMultibinding] annotation. The equivalent would be:
+ * ```
+ * interface Listener
+ *
+ * @ContributesMultibinding(AppScope::class)
+ * class MainListener @Inject constructor() : Listener
+ * ```
+ * Notice that it's optional to specify [boundType], if there is only exactly one super type. If
+ * there are multiple super types, then it's required to specify the parameter:
+ * ```
+ * @ContributesMultibinding(
+ *   scope = AppScope::class,
+ *   boundType = Listener::class
+ * )
+ * class MainListener @Inject constructor() : Activity(), Listener
+ * ```
+ *
+ * [ContributesMultibinding] supports qualifiers. If you annotate the class additionally with a
+ * qualifier, then the generated multibinding method will be annotated with the same qualifier,
+ * e.g.
+ * ```
+ * @ContributesMultibinding(AppScope::class)
+ * @Named("Prod")
+ * class MainListener @Inject constructor() : Listener
+ *
+ * // Will generate this binding method.
+ * @Binds @IntoSet @Named("Prod")
+ * abstract fun bindMainListener(listener: MainListener): Listener
+ * ```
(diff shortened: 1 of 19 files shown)


annotations/src/main/java/com/squareup/anvil/annotations/ContributesMultibinding.kt
```diff
@@ -0,0 +1,90 @@
...
+ * one of the annotations, then you can set [ignoreQualifier] to `true` and the qualifier won't
+ * be added to the generated multibinding method.
+ * ```
+ * @ContributesBinding(AppScope::class)
+ * @ContributesMultibinding(AppScope::class, ignoreQualifier = true)
+ * @Named("Prod")
+ * object MainListener : Listener
+ * object MainListener : Listener
+ * ```
+ */
```
vRallev, line 82: The API is described here. The documentation for map multibindings is part of #219.

marcellogalhardo, line 82: I really enjoyed the API surface, and I like how simple it is to use `Set` or `Map`. I think this feature will help to remove a lot of boilerplate and it is very intuitive in my opinion. Thank you a lot for this work!

There is one concern that I would like to raise around the overall API, when considering #219, that might affect developers that do not have a lot of experience with Dagger: I feel many of them struggle with multi-binding. If I look at vanilla Dagger, it requires a specific annotation (`@IntoMap`, `@IntoSet`) that help developers to instantly identify what type of multi-binding they are using, and it also helps to fail when a wrong combination is found (e.g., `@IntoMap` without a key).

On the other hand, it might be hard for those developers to understand if a `@ContributesMultibinding` is a `Map` or `Set`. Another point is that, correct me if I'm wrong, removing a `@MapKey` from a `@ContributesMultibinding` by mistake would automatically send that contributes to a `Set` without the developer be aware of and could be trick to catch or even result in runtime exceptions when the key is not found in the map.

I personally prefer how you did with `@ContributesMultibinding` and it feels more intuitive to me but I think it is relevant to consider the cases I mentioned above. What do you think? 🤔

vRallev: I thought about that, but I disliked the idea of having two different annotations or adding a parameter to differentiate between a set and map. This parameter felt like boilerplate when it can be implied. The parameter shouldn't use a default value, so you always must specify it. If you set the parameter to a Map and the map key is missing, then it's an error. If you set the parameter to a Set and add a map key, then it's an error. It feels like you need to declare twice whether you want a set or map multibinding. Since Anvil is all about less boilerplate I decided for this way. 

I agree that it can lead to mistakes, but you could say the same thing about multibindings in general, e.g. when you forget to include the module with vanilla Dagger.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -190,67 +188,135 @@ internal class BindingModuleGenerator(
...
-          if (DescriptorUtils.isObject(contributedClass)) {
-            ProviderMethod(
-              FunSpec
-                .builder(
-                  name = concreteType
-                    .asString()
-                    .split(".")
-                    .joinToString(separator = "", prefix = "provide") {
-                      it.capitalize(US)
+      // Note that this is an inner function to share some of the parameters. It computes all
```
FrostRocket, line 193: 👍

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -190,67 +188,135 @@ internal class BindingModuleGenerator(
...
-            )
+                  }
+                  .addAnnotations(qualifiers)
+                  .returns(boundType.asClassName())
+                  .addStatement("return %T", contributedClass.asClassName())
+                  .build()
+              )
+            } else {
+              BindingMethod(
+                FunSpec
```
FrostRocket, line 276: This API is quite nice. You can visualize how the function looks in code easily by following this chain.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesBindingGenerator.kt
```diff
@@ -37,15 +38,8 @@ internal class ContributesBindingGenerator : CodeGenerator {
...
-        if (clazz.visibilityModifierTypeOrDefault().value != KtTokens.PUBLIC_KEYWORD.value) {
-          throw AnvilCompilationException(
-            "${clazz.requireFqName()} is binding a type, but the class is not public. Only " +
-              "public types are supported.",
-            element = clazz.identifyingElement
-          )
-        }
-
-        checkNotMoreThanOneQualifier(module, clazz)
+        clazz.checkClassIsPublic()
```
FrostRocket, line 41: Great use of extensions.

compiler/src/test/java/com/squareup/anvil/compiler/TestUtils.kt
```diff
@@ -134,26 +134,30 @@ internal val Result.anyQualifier: Class<*>
...
+  get() = getHintScope(HINT_BINDING_PACKAGE_PREFIX)
+
+internal val Class<*>.hintMultibinding: KClass<*>?
+  get() = getHint(HINT_MULTIBINDING_PACKAGE_PREFIX)
+
+internal val Class<*>.hintMultibindingScope: KClass<*>?
+  get() = getHintScope(HINT_MULTIBINDING_PACKAGE_PREFIX)
+
+private fun Class<*>.getHint(prefix: String): KClass<*>? = contributedProperties(prefix)
+  ?.filter { it.java == this }
```
FrostRocket, line 155: Is it.java nullable? Just making sure this is a safe comparison to make.

FrostRocket (approved):
You've been busy. Nice work.

compiler/src/test/java/com/squareup/anvil/compiler/TestUtils.kt
vRallev: No, it's not nullable. Even then the check would be safe, it would simply return false.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vRallev: I love Kotlin Poet, it's one of the few libraries that still feels magical to me.

PR #219, opened 2021-03-03
Generate multibinding methods for contributed map multibindings. 

vRallev:
If a class is annotated with @ContributesMultibinding and a map key annotation, then Anvil will generate a map multibinding method.

This PR completes #152.

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleMultibindingMapTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleMultibindingMapTest.kt
@@ -1,0 +1,295 @@
+package com.squareup.anvil.compiler.codegen
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.annotations.MergeComponent
+import com.squareup.anvil.annotations.MergeSubcomponent
+import com.squareup.anvil.annotations.compat.MergeModules
+import com.squareup.anvil.compiler.AnyDaggerComponent
+import com.squareup.anvil.compiler.anyDaggerComponent
+import com.squareup.anvil.compiler.bindingKey
+import com.squareup.anvil.compiler.compile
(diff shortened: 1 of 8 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleMultibindingMapTest.kt
```diff
@@ -0,0 +1,295 @@
...
+
+      val methods = modules.single().java.declaredMethods
+      assertThat(methods).hasLength(1)
+
+      with(methods[0]) {
+        assertThat(returnType).isEqualTo(parentInterface)
+        assertThat(parameterTypes.toList()).isEmpty()
+        assertThat(isAbstract).isFalse()
+        assertThat(isAnnotationPresent(Provides::class.java)).isTrue()
+        assertThat(isAnnotationPresent(IntoMap::class.java)).isTrue()
```
FrostRocket, line 127: Redundant, but add assertThat(isAnnotationPresent(IntoSet::class.java)).isFalse() for consistency?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleMultibindingMapTest.kt
```diff
@@ -0,0 +1,295 @@
+package com.squareup.anvil.compiler.codegen
```
FrostRocket, line 1: Love all the tests. ❤️

FrostRocket (approved):

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleMultibindingMapTest.kt
vRallev: That's the bad thing about the tests, there's a lot of repetition. If one condition fails, then many tests start to fail. Ideally, there would be only one failing :-/

PR #226, opened 2021-03-05
Support different priorities for contributed bindings

### LINKED ISSUE #161: Support replacing contributed classes without a direct reference in tests
Tests (unit tests and android tests) have this unfortunate situation that their classpath consists of the main classpath of this module plus test specific dependencies. This leads to duplicate bindings, if a type is implemented in two different modules and one of them should be used for tests.
```
:lib-a -> @ContriburesBinding class A : Interface
:lib-b -> @ContriburesBinding class B : Interface

:app
 -> implementation(':lib-a')
 -> testImplementation(':lib-b')
```
Anvil provides currently two solutions. Either the annotation for `class B` replaces `class A`, or the component in the test sources excludes `class A`. For the former `:lib-b` must depend on `:lib-a` in order to reference `class A`. That's often not desired. The 2nd option doesn't scale well. This exclusion has to be added for every test target and that's exactly what Anvil tries to solve.

There are multiple options how this can be solved. 
* The annotation could have a new attribute that say to prefer this binding / contributed class in test compilations. If two classes have this attribute set, then fail.
* Have a priority (integer) for contributions. Fail if conflicts use the same priority. 
* Allow replacing contributions with a String rather than a class reference. 
* ...

vRallev:
Add a priority field to @ContributesBinding to replace other contributed bindings. The priority should only be used, if you don't have access to the replaced class at compile time. Otherwise prefer the `replaces` field.

Fixes #161

### DIFF
--- a/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
+++ b/annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
@@ -1,5 +1,6 @@
 package com.squareup.anvil.annotations
 
+import com.squareup.anvil.annotations.ContributesBinding.Priority.NORMAL
 import kotlin.annotation.AnnotationRetention.RUNTIME
 import kotlin.annotation.AnnotationTarget.CLASS
 import kotlin.reflect.KClass
@@ -72,6 +73,10 @@ import kotlin.reflect.KClass
  * )
  * class FakeAuthenticator @Inject constructor() : Authenticator
  * ```
+ * If you don't have access to the class of another contributed binding that you want to replace,
+ * then you can change the [priority] of the bindings to avoid duplicate bindings. The contributed
+ * binding with the higher priority will be used.
+ *
  * [ContributesBinding] supports Kotlin objects, e.g.
  * ```
  * @ContributesBinding(AppScope::class)
@@ -99,10 +104,35 @@ public annotation class ContributesBinding(
    */
   val replaces: Array<KClass<*>> = [],
   /**
+   * The priority of this contributed binding. The priority should be changed only if you don't
+   * have access to the contributed binding class that you want to replace at compile time. If
+   * you have access and can reference the other class, then it's highly suggested to
+   * use [replaces] instead.
+   *
+   * In case of a duplicate binding for multiple contributed bindings the binding with the highest
+   * priority will be used and replace other contributed bindings for the same type with a lower
+   * priority. If duplicate contributed bindings use the same priority, then there will be an
+   * error for duplicate bindings.
+   *
+   * Note that [replaces] takes precedence. If you explicitly replace a binding it won't be
+   * considered no matter what its priority is.
+   *
+   * All contributed bindings have a [NORMAL] priority by default.
+   */
+  val priority: Priority = NORMAL,
+  /**
    * Whether the qualifier for this class should be included in the generated binding method. This
    * parameter is only necessary to use when [ContributesBinding] and [ContributesMultibinding]
    * are used together for the same class. If not, simply remove the qualifier from the class
    * and don't use this parameter.
    */
   val ignoreQualifier: Boolean = false
-)
+) {
+  /**
+   * The priority of a contributed binding.
+   */
+  @Suppress("unused")
+  public enum class Priority {
+    NORMAL, HIGH, HIGHEST
+  }
+}
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -29,6 +29,7 @@ import com.squareup.anvil.compiler.isQualifier
 import com.squareup.anvil.compiler.mergeComponentFqName
 import com.squareup.anvil.compiler.mergeModulesFqName
 import com.squareup.anvil.compiler.mergeSubcomponentFqName
+import com.squareup.anvil.compiler.priority
 import com.squareup.anvil.compiler.replaces
 import com.squareup.anvil.compiler.scope
 import com.squareup.kotlinpoet.AnnotationSpec
@@ -43,6 +44,7 @@ import dagger.multibindings.IntoMap
 import dagger.multibindings.IntoSet
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.descriptors.annotations.AnnotationDescriptor
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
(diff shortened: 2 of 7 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -473,4 +497,34 @@ private sealed class GeneratedMethod {
...
+  val boundType: ClassDescriptor
+)
+
+private fun List<Binding>.findHighestPriorityBinding(): Binding {
+  if (size == 1) return this[0]
+
+  val bindingsByPriority = groupBy { it.annotation.priority() }
+    .filter { it.value.isNotEmpty() }
+
+  val bindings = bindingsByPriority[HIGHEST]
```
0legg, line 512: You could've sort `bindingsByPriority` and take `.firstOrNull { it.isNotEmpty() }` instead of hardcoding priorities.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -473,4 +497,34 @@ private sealed class GeneratedMethod {
...
+    ?: bindingsByPriority[HIGH]
+    ?: bindingsByPriority.getValue(NORMAL)
+
+  if (bindings.size > 1) {
```
0legg, line 516: What is our expected behavior if we have one binding of `HIGH` priority, and two of `NORMAL` one? Right now we'll return the one with `HIGH`, but maybe we want to warn consumer about lower priority clash. Also, it's unclear from the documentation what will be the result.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -237,6 +242,25 @@ internal class BindingModuleGenerator(
...
+              bindings
+                .groupBy {
+                  // It's safe to use the FqName here, because we don't allow generic super types.
+                  // If we wouldn't have this limitation, we'd need to use the Kotlin type.
+                  it.boundType.fqNameSafe
+                }
+                .map {
+                  it.value.findHighestPriorityBinding()
+                }
+                .asSequence()
```
0legg, line 260: Will it fail faster (or use less memory) if we'll first convert to the `Sequence`, and then will do mapping?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModulePriorityTest.kt
```diff
@@ -0,0 +1,158 @@
...
+      import com.squareup.anvil.annotations.ContributesBinding.Priority.HIGH
+      import com.squareup.anvil.annotations.ContributesBinding.Priority.HIGHEST
+      $import
+      
+      interface ParentInterface
+      
+      @ContributesBinding(Any::class, priority = HIGHEST)
+      interface ContributingInterface : ParentInterface
+      
+      @ContributesBinding(Any::class, priority = HIGH, replaces = [ContributingInterface::class])
```
0legg, line 83: We should update documentation with the note that `replaces` has a priority over `priority`

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -98,11 +103,33 @@ public annotation class ContributesBinding(
    * use the same scope.
    */
   val replaces: Array<KClass<*>> = [],
+  /**
+   * The priority of this contributed binding. The priority should be changed only if you don't
+   * have access to the contributed binding class that you want to replace at comppile time. If
```
0legg, line 108: typo: `comppile`

0legg (approved):
LGTM, but I'd give one more run on documentation and edge cases.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vRallev: The highest priority wins. That's the only expectation and what's documented. Lower priorities are simply filtered. I don't see the need to print any warnings (Anvil isn't doing this at other places either).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vRallev: Everything in this longer sequence. But because of the `groupBy` above we get a Map and the `map` operator converts it to a sequence.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModulePriorityTest.kt
vRallev: Done

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vRallev: I considered this first, but didn't do it because it's a map. But I just learnt that Kotlin has `SortedMap` and sorting a map with max 3 items is very quick. I updated that.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -98,11 +103,36 @@ public annotation class ContributesBinding(
...
+   * have access to the contributed binding class that you want to replace at compile time. If
+   * you have access and can reference the other class, then it's highly suggested to
+   * use [replaces] instead.
+   *
+   * In case of a duplicate binding for multiple contributed bindings the binding with the highest
+   * priority will be used and replace other contributed bindings for the same type with a lower
+   * priority. If duplicate contributed bindings use the same priority, then there will be an
+   * error for duplicate bindings.
+   *
+   * Note that [replaces] takes precedence. If you explicitly replace a binding don't it won't be
```
0legg, line 117: nit: remove `don't`

0legg (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -473,4 +497,32 @@ private sealed class GeneratedMethod {
...
+  val contributedClass: ClassDescriptor,
+  val annotation: AnnotationDescriptor,
+  val boundType: ClassDescriptor
+)
+
+private fun List<Binding>.findHighestPriorityBinding(): Binding {
+  if (size == 1) return this[0]
+
+  val bindings = groupBy { it.annotation.priority() }
+    .filter { it.value.isNotEmpty() }
```
tehras, line 510: When would this be empty?

vRallev: It was more like a sanity check, but you're absolutely right it shouldn't happen. With the tests the sanity check shouldn't be necessary. I removed this line.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -98,11 +103,36 @@ public annotation class ContributesBinding(
...
    */
   val ignoreQualifier: Boolean = false
-)
+) {
+  /**
+   * The priority of a contributed binding.
+   */
+  @Suppress("unused")
+  public enum class Priority {
+    NORMAL, HIGH, HIGHEST
```
0xMatthewGroves, line 136: LOW, NORMAL, HIGH would make more sense to me. Normal would be the default. Low would be for fake/mock objects. High would be for those rare situations where you need to override the normal priority.

Why'd you decide to go with enums in the end instead of ints?

0xMatthewGroves (approved):
Just one minor comment about the verbiage of the enums, but in the end is a matter of taste.

annotations/src/main/java/com/squareup/anvil/annotations/ContributesBinding.kt
vRallev: In the end I decided for an enum for better auto-complete. And I really can't think of a scenario where you need more options and can't rely on `replaces` either.

Hm, I preferred these three constants, because my gut thinks that you rather want to override existing bindings than provide one with lower priority. I don't want to use `LOW` as the default, because that is awkward to read.

0xMatthewGroves: I see what you mean. I'm fine with the way it is then.

PR #229, opened 2021-03-15
Avoid duplicate iterations and unnecessary type conversions to improve the performance

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -64,11 +64,13 @@ internal class InterfaceMerger(
           )
         }
       }
+      // Convert the sequence to a list to avoid iterating it twice. We use the result twice
+      // for replaced classes and the final result.
+      .toList()
```
vRallev, line 69: I noticed while debugging that the filter and map lambdas are called twice. That's unnecessary.

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -64,11 +64,13 @@ internal class InterfaceMerger(
...
 
     val replacedClasses = classes
       .flatMap { (classDescriptor, contributeAnnotation) ->
         contributeAnnotation.replaces(classDescriptor.module)
-          .asSequence()
```
vRallev, line 71: Most of the time `replaces(..)` returns an empty list. Using the `replaces` annotation attribute should be rare. It's unnecessary to convert an empty list to a sequence.

christiandeange (approved):

0xMatthewGroves (approved):

PR #230, opened 2021-03-16
If the bound type is specified, then multiple super types are allowed.

vRallev:
That's a regression I noticed while testing the new version in our codebase.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
@@ -120,20 +120,51 @@ internal fun KtClassOrObject.scope(
   annotationFqName: FqName,
   module: ModuleDescriptor
 ): FqName {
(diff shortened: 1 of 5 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -120,20 +120,50 @@ internal fun KtClassOrObject.scope(
...
+      if (classLiteralExpression == null) {
+        throw AnvilCompilationException(
+          "The first argument for $annotationFqName must be a class literal: $text",
+          element = this
+        )
+      }
+
+      val children = classLiteralExpression.children
+      children.singleOrNull() ?: throw AnvilCompilationException(
+        "Expected a single child, but had ${children.size} instead: ${classLiteralExpression.text}",
```
0xMatthewGroves, line 134: Nit: "but had" -> "but there were"

0xMatthewGroves (approved):

PR #231, opened 2021-03-16
Add more tests to verify that exclusions and replacements don't result in duplicate bindings

vRallev:
I thought I detected a bug in Anvil while testing the release in our codebase, but it turns out that our project was configured wrong. I wanted to keep nonetheless for more confidence and to avoid regressions.


0xMatthewGroves (approved):
Extra tests are always nice. 👍

PR #243, opened 2021-04-09
Handle Lazy assisted parameters properly in generated assisted factories

vRallev:
There's always way more test code than actual changes in the compiler plugin.


rjrjr (approved):

rjrjr:
> There's always way more test code than actual changes in the compiler plugin.

Always a good sign.

PR #250, opened 2021-04-13
Build Anvil with Kotlin 1.5.0-RC in CI.

vRallev:


dependencies.gradle
```diff
@@ -12,6 +14,8 @@ ext {
 
   generateDaggerFactoriesWithAnvil = (project.hasProperty('square.generateDaggerFactoriesWithAnvil') ?
       project.getProperty('square.generateDaggerFactoriesWithAnvil') : 'true').toBoolean()
+  allWarningsAsErrors = ci &&
+      VersionNumber.parse(kotlinVersion).baseVersion < VersionNumber.parse("1.5.0").baseVersion
```
vRallev, line 18: There are deprecated APIs in 1.5, so I disabled it.

gradle-plugin/build.gradle
```diff
@@ -62,9 +60,9 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
   kotlinOptions {
     jvmTarget = JavaVersion.VERSION_1_8
 
-    // Because Gradle hardcodes Kotlin versions we have warnings during compilation.
-    allWarningsAsErrors = rootProject.ext.ci &&
-        VersionNumber.parse(rootProject.ext.kotlinVersion).baseVersion < VersionNumber.parse("1.5.0").baseVersion
+    // Because Gradle hardcodes Kotlin versions we have warnings during compilation. We might need
+    // disable this setting when upgrading to 1.5.0 stable.
```
lmchew, line 64: nit: missing the word "to": "We might need to disable this setting..."

lmchew (approved):

gradle-plugin/build.gradle
lmchew: Looking at it again, I wonder if this whole comment should be moved to dependencies.gradle, since it's referencing the version-based logic that you moved there.

vRallev: Uff, I saw your comment after merging. But I don't want to move the comment either, because it only applies to the gradle plugin (this module) and nowhere else.

PR #254, opened 2021-04-17
fix where clauses and generic type bounds in injected classes

### LINKED ISSUE #251: where clauses not handled correctly
```kotlin
class SKTokenSelectDelegateImpl<SCDMBinder> @Inject constructor(
  private val prefsManagerLazy: Lazy<PrefsManager>,
  private val slackConnectDmViewBinderLazy: Lazy<SCDMBinder>,
  private val tokenResultsUserViewBinderLazy: Lazy<TokenResultsUserViewBinder>,
  @AndroidIncreaseMpdmLimitFeatureFlag private val increaseMpdmLimit: Boolean,
) : SKTokenSelectDelegate where SCDMBinder : SKListCustomViewBinder, SCDMBinder : SKTokenSlackConnectDmBinder {
```

Generates

```kotlin
public class SKTokenSelectDelegateImpl_Factory<SCDMBinder>(
  private val param0: Provider<@JvmSuppressWildcards PrefsManager>,
  private val param1: Provider<@JvmSuppressWildcards SCDMBinder>,
  private val param2: Provider<@JvmSuppressWildcards TokenResultsUserViewBinder>,
  private val param3: Provider<Boolean>
) : Factory<SKTokenSelectDelegateImpl<SCDMBinder>> {
```

the generated factory is missing the bounds on `SCDMBinder`

### LINKED ISSUE #253: factory generation fails when the injected class has a type variable bound by a generic type
source:
``` Kotlin
class InjectClass<T : List<String>> @Inject constructor(
  private val t: T
)
```

The generated factory should have a type variable of `T : List<String>`, but it's just `T : List`.

generated:
``` Kotlin
public class InjectClass_Factory<T : List>(
  private val param0: Provider<T>
) : Factory<InjectClass<T>> {
  public override fun `get`(): InjectClass<T> = newInstance(param0.get())

  public companion object {
    @JvmStatic
    public fun <T : List> create(param0: Provider<T>): InjectClass_Factory<T> =
        InjectClass_Factory<T>(param0)

    @JvmStatic
    public fun <T : List> newInstance(param0: T): InjectClass<T> = InjectClass<T>(param0)
  }
}
```

RBusarow:
fixes #253 
fixes #251

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
@@ -8,7 +8,6 @@ import com.squareup.anvil.compiler.injectFqName
 import com.squareup.anvil.compiler.jvmSuppressWildcardsFqName
 import com.squareup.anvil.compiler.publishedApiFqName
 import com.squareup.anvil.compiler.safePackageString
-import com.squareup.kotlinpoet.TypeName
 import com.squareup.kotlinpoet.TypeVariableName
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
@@ -48,7 +47,6 @@ import org.jetbrains.kotlin.psi.allConstructors
 import org.jetbrains.kotlin.psi.psiUtil.parents
 import org.jetbrains.kotlin.psi.psiUtil.parentsWithSelf
 import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
-import org.jetbrains.kotlin.utils.addIfNotNull
 
 private val kotlinAnnotations = listOf(jvmSuppressWildcardsFqName, publishedApiFqName)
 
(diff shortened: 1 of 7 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -402,6 +405,49 @@ internal fun ModuleDescriptor.findClassOrTypeAlias(
...
+  // It's also technically possible to have one constraint in the type parameter spot, like this:
+  // class MyClass<T : Any> where T : Set<*>, T : MutableCollection<*>
+  // Merge both groups of type parameters in order to get the full list of bounds.
+  val boundsByVariableName = mutableMapOf<String, MutableList<TypeName>>()
+
+  typeParameterList
+    ?.parameters
+    ?.forEach { parameter ->
+
+      if (parameter.fqNameOrNull(module) == null) {
```
vRallev, line 421: That if check could be a `.filter {..}`

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
```diff
@@ -402,6 +405,49 @@ internal fun ModuleDescriptor.findClassOrTypeAlias(
...
 }
 
+internal fun KtClassOrObject.typeVariableNames(
+  module: ModuleDescriptor
+): List<TypeVariableName> {
+  // Any type which is constrained in a `where` clause is also defined as a type parameter.
+  // It's also technically possible to have one constraint in the type parameter spot, like this:
+  // class MyClass<T : Any> where T : Set<*>, T : MutableCollection<*>
+  // Merge both groups of type parameters in order to get the full list of bounds.
+  val boundsByVariableName = mutableMapOf<String, MutableList<TypeName>>()
```
vRallev, line 415: Appen `.withDefault { mutableListOf() }`, then you can replace the `getOrPut`s below with `getValue(variableName)`.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedInjectGenerator.kt
```diff
@@ -73,17 +72,7 @@ internal class AssistedInjectGenerator : PrivateCodeGenerator() {
...
-      ?.mapNotNull { parameter ->
-        val extendsBound = parameter.extendsBound?.fqNameOrNull(module)?.asClassName(module)
-        if (parameter.fqNameOrNull(module) == null) {
-          TypeVariableName(parameter.nameAsSafeName.asString(), listOfNotNull(extendsBound))
-        } else {
-          null
-        }
-      }
-      ?: emptyList()
+    val typeParameters = clazz.typeVariableNames(module)
```
vRallev, line 75: Nice cleanup.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
```diff
@@ -409,6 +409,106 @@ public final class AssistedServiceFactory_Impl<T extends CharSequence> implement
...
+      package com.squareup.test
+      
+      import dagger.assisted.Assisted
+      import dagger.assisted.AssistedFactory
+      import dagger.assisted.AssistedInject
+      
+      data class AssistedService<T> @AssistedInject constructor(
+        val int: Int,
+        @Assisted val stringBuilder: T
+      ) where T : Appendable, T : CharSequence {
```
vRallev, line 469: Awesome!

vRallev (changes requested):
Great PR, thank you! Only some minor nits that I'd like to see changed.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
RBusarow: Unfortunately that doesn't work.  `withDefault { ... }` returns a default list, but doesn't actually do a `put` into to the map.  So after adding the `TypeName` to the default list, we'd still need to add a `put`.

Instead, I just used `associateTo` for the parameters.  Since any generic in the constraint list must also be at least named in the `parameters` list, we can still use `getValue` in the second `forEach`.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
RBusarow: This is funny because I'd considered it after I made the PR, then decided I was nitpicking and should just leave it alone.  I should have gone with my first instinct. :)

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/PsiUtils.kt
vRallev: Oh, I wasn't aware. Sounds good!

PR #266, opened 2021-05-14
Copy over qualifiers in member injecttion

### LINKED ISSUE #264: Qualifiers on properties not preserved
Given this example:

```kotlin
class Example {
  @Inject
  @ForScope(UserScope::class)
  lateinit var activityNavigator: ActivityNavigator
}
```

The qualifier is not accounted for in Dagger. The fix is to add a field site target

```kotlin
class Example {
  @Inject
  @field:ForScope(UserScope::class)
  lateinit var activityNavigator: ActivityNavigator
}
```

ZacSweers:
Resolves #264

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
@@ -9,10 +9,13 @@ import com.squareup.anvil.compiler.codegen.createGeneratedFile
 import com.squareup.anvil.compiler.codegen.hasAnnotation
 import com.squareup.anvil.compiler.codegen.isGenericClass
 import com.squareup.anvil.compiler.codegen.mapToParameter
+import com.squareup.anvil.compiler.codegen.requireClassDescriptor
 import com.squareup.anvil.compiler.codegen.requireFqName
+import com.squareup.anvil.compiler.codegen.toAnnotationSpec
 import com.squareup.anvil.compiler.daggerDoubleCheckFqNameString
 import com.squareup.anvil.compiler.generateClassName
 import com.squareup.anvil.compiler.injectFqName
+import com.squareup.anvil.compiler.isQualifier
 import com.squareup.anvil.compiler.safePackageString
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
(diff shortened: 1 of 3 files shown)


ZacSweers:
I don't understand this test failure and how it relates to my change 🤔 

Suggestions welcome, will keep digging

image

ZacSweers:
The failing test passes for me locally 🤔

ZacSweers:
Scratch that, I think I ran a stale intelliJ run rather than gradle. Repro'd now and looking into it

ZacSweers:
Both CI failures are flakes in the validation action, just need to be re-kicked. Filed this for the flakes https://github.com/gradle/wrapper-validation-action/issues/40

compiler/src/test/java/com/squareup/anvil/compiler/dagger/MembersInjectorGeneratorTest.kt
```diff
@@ -122,11 +134,13 @@ public final class InjectClass_MembersInjector implements MembersInjector<Inject
       package com.squareup.test
       
       import javax.inject.Inject
+      import javax.inject.Named
       
       typealias StringList = List<String>
       
       class InjectClass {
         @Inject lateinit var string: String
+        @Named("qualified") @Inject lateinit var qualifiedString: String
```
vRallev, line 143: The updated test passed even without your changes to fix the issue. I've added this check:
```kotlin
val namedAnnotation = membersInjector.staticInjectMethod("qualifiedString").annotations
  .single { it.annotationClass == Named::class }
assertThat(namedAnnotation.getValue<String>()).isEqualTo("qualified")
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
```diff
@@ -172,6 +192,19 @@ internal class MembersInjectorGenerator : PrivateCodeGenerator() {
                           .addMember("%S", property.requireFqName())
                           .build()
                       )
+                      .apply {
+                        val hasNonInjectAnnotation = property.annotationEntries
+                          .filterNot { it.isInject(module) }
+                          .isNotEmpty()
```
vRallev, line 198: This check is a little strange, I've added a `KtAnnotationEntry.isQualifier()` function.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
```diff
@@ -92,6 +103,15 @@ internal class MembersInjectorGenerator : PrivateCodeGenerator() {
 
     val memberInjectorClass = ClassName(packageName, className)
 
+    // Lazily evaluated, only computed if any properties have any non-inject annotations
+    val propertyDescriptors by lazy {
+      clazz.requireClassDescriptor(module)
```
vRallev, line 108: I hoped to avoid resolving the class descriptor, but writing a function and copying all annotation arguments is tedious.

vRallev (approved):
I squashed your commits and wrote a new commit message. None of the previous commit messages described what the issue was about.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
ZacSweers: Yeah I ran into the same and that's why I made this lazy. We could revisit if it's a major problem for anyone in the future.

vRallev: Yes, the lazy option is a good compromise in my opinion.

Issue #265, opened 2021-05-14
Expose `CodeGenerator` interface


vRallev:
This would allow consumers to create there own `CodeGenerator` and generate code at compile time. We have several uses cases for that internally like contributing `ScreenRobot`. But also external users would like this feature (some had to fork Anvil for that).

This would require breaking up the `:compiler` module in an API module. `CodeGenerators` would need to be loaded through a service loader probably. 

Tasks for WIP:
- [x] `GeneratedFile` should not be an inner class.
- [x] Create an `anvil` configuration to add dependencies to the Kotlin compiler plugin classpath.
- [x] Eliminate `CodeGenerator.flush()`.
- [x] Update the README.
- [x] How should errors be propagated? Should `AnvilCompilationException` be exposed?

RBusarow:
This is fantastic! I've been meaning to make a feature request for exactly this.  I'd be happy to take a stab at the implementation over the weekend, if you're open to it.

ZacSweers:
I've been chatting with Ralf about a possible design for it as we have a pretty good use case in mind, let me take a spin on it first

RBusarow:
Sounds great. 😃

ZacSweers:
PR is up! https://github.com/square/anvil/pull/267

PR #267, opened 2021-05-14
Extract CodeGenerator to separate compiler-api module

ZacSweers:
This introduces a new `compiler-api` artifact that hosts the `CodeGenerator` interface. This is loaded via `ServiceLoader` and allows both Anvil's internal generators as well as 3rd party other generators to wire up their functionality via this new API.

The general design is almost identical to the original `CodeGenerator`, only adding a new `isApplicable` function + `AnvilContext` to allow implementations to indicate if they are applicable. This logic was previously handwritten into `AnvilComponentRegistrar`, and allows moving this logic out externally to implementations.

One generator － `BindingModuleGenerator` － could not fit in this new pattern because it requires a `ClassScanner` instance. This isn't really a blocker though, it can just stay internal and is still added manually. We could explore supporting its needs via the compiler API as well.

All the compiler APIs are guarded by a new 	`@ExperimentalAnvilApi` annotation to allow for breaking changes going forward while this API incubates.

---

**TODO**
- [x] Docs pass
- [x] Add Experimental API annotations
- [x] README mention?
- [x] Should we codify `PrivateCodeGenerator`'s behavior in the API at all?
- [x] How should custom generators be added to the classpath? I'm thinking exposing an `anvil` configuration that dependencies can be added to and then loaded into the classpath, and make the compileClasspath extend from it under the hood.
  ```kotlin
  dependencies {
    anvil(project(":my-local-generator"))
  }
  ```

---

**Future considerations**

* One thing I was wondering about is the presence of both a `flush()` and `GeneratedFile`. Would it maybe be worth converging on one or the other?
  * If we keep `GeneratedFile`, should `contents` be moved to a `flush()` function to allow more dynamic/lazy/streamed writes rather than hold all the file contents in memory?
* How can/should errors be propagated?
* Currently all generated files must be Kotlin, but it may be nice in the future to allow generation of resource files (i.e. such as generated proguard files, which Dagger does in some cases too).


compiler-api/gradle.properties
```diff
@@ -0,0 +1,3 @@
+POM_NAME=Anvil Compiler
+POM_ARTIFACT_ID=compiler
```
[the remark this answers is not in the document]
ZacSweers: Thanks, forgot to circle around to this

ZacSweers:
CI failure is due to a remote repo outage, just needs a re-kick https://github.com/square/anvil/pull/267/checks?check_run_id=2588874471

RBusarow:
Thanks for doing this, @ZacSweers.  I spent some time consuming this branch yesterday and it's ridiculous just how much this streamlines the process of creating a new generator.

> * [ ]  Should we codify `PrivateCodeGenerator`'s behavior in the API at all?

I think so, and I'm wondering about taking it a step further.  It would be nice to influence the order of execution for the code generators.  For example, I'm doing this pattern a lot, with the generated dir added to the source set (and plans for IDE integration):

``` Kotlin
class ManuallyWrittenClass @Inject constructor(
)
```
I'm running into an issue where `InjectConstructorFactoryGenerator` intermittently throws an exception at `requireFqName`. This is because `InjectConstructorFactoryGenerator` does its first pass before my generator, and it's trying to resolve a class which doesn't exist yet.  Dagger's generator doesn't have an issue since it's obviously later.

My idea is to add an `executionOrder` function to `CodeGenerator`.

``` Kotlin
interface CodeGenerator {

  fun executionOrder(): List<ExecutionOrder> = listOf(ANY)
  ...
  sealed class ExecutionOrder {
    object ANY : ExecutionOrder()
    class After(generatorFqName: String) : ExecutionOrder()
    class Before(generatorFqName: String) : ExecutionOrder()
  }
}
class MyGenerator: CodeGenerator {
  override fun executionOrder() = listOf(
      Before("com.squareup.anvil.compiler.codegen.dagger.InjectConstructorFactoryGenerator"),
      After("com.example.MyOtherGenerator")
    )
  ...
}
```
Then `AnvilComponentRegistrar` or `CodeGenerationExtension` would apply a sort before they're iterated over.

Naturally I'd be happy to make these changes myself later if you guys like the idea, but not the scope creep.

> * [ ]  How should custom generators be added to the classpath? I'm thinking exposing an `anvil` configuration that dependencies can be added to and then loaded into the classpath, and make the compileClasspath extend from it under the hood.
>   ```kotlin
>   dependencies {
>     anvil(project(":my-local-generator"))
>   }
>   ```

This is what I was thinking as well, except I think we need `kotlinCompilerPluginClasspath`?  It probably makes sense to make different anvil configs for each variation of the compiler configs.

| compiler | anvil |
|:---:|:---:|
| kotlinCompilerPluginClasspath | anvil |
| kotlinCompilerPluginClasspathDebug | anvilDebug |
| kotlinCompilerPluginClasspathAndroidTest | anvilDebugAndroidTest |
| etc. | etc. |

ZacSweers:
I like the idea, but do think it's complex enough that that would be better served as a separate followup after the first pass and getting some community feedback in the next release. What you're describing is effectively supporting multiple rounds, which is tricky to get right and ideally individual generators shouldn't need to know about. I think if they did need to declare some prioritization, a simpler method would be to just make them optionally implement `Comparable` and allow them to sort themselves as-needed against known other generators.

Re: classpath - good point, didn't know about the separate configuration being available for that. I agree with different configs, but not sure if that would take a bit more work as I don't think anvil really thinks about different variants right now.

vRallev:
I didn't have a chance to take a look at the code changes yet, but the overall structure looks great. I'd appreciate if you could squash commits and update the commit message to be more meaningful.

I'm a fan of `isApplicable`. Manually initializing `BindingModuleGenerator` for now is probably fine. I wouldn't want to expose `ClassScanner`. 

I'm not a fan of `GeneratedFile` and `flush()` anymore from an API perspective, especially `flush()` is confusing. It's only overridden by one code generator and it's only needed for `PrivateCodeGenerators`. This whole interaction is weird and was more like a quick fix and small perf improvement. So no, I wouldn't want to expose `PrivateCodeGenerator`. `flush()` also needs to be removed, but this can be a separate PR (I could take care of this as well). 

> How can/should errors be propagated?

Maybe we should expose `AnvilCompilationException`? 

> How should custom generators be added to the classpath? 

I like the idea of the `anvil` configuration, which would then need to be added to the Kotlin compiler classpath (not compile classpath). The Kotlin compiler classpath doesn't support variants, so we don't need to either.

> Currently all generated files must be Kotlin, but it may be nice in the future to allow generation of resource files (i.e. such as generated proguard files, which Dagger does in some cases too).

No, at this point I'd suggest to create your own compiler plugin. This shouldn't be necessary for Anvil extensions. 

I wouldn't want to introduce an order for code generators. @RBusarow  you're only running into this issue, because `InjectConstructorFactoryGenerator` is a private code generator. Otherwise you could react to the multiple rounds. This issue is independent of exposing the `CodeGenerator` interface.

annotations/build.gradle
```diff
@@ -1,6 +1,14 @@
...
 
 kotlin {
-  explicitApi = 'strict'
+  explicitApi()
+}
+
+tasks.withType(KotlinCompile).configureEach {
+  kotlinOptions {
+    freeCompilerArgs += ["-Xopt-in=kotlin.RequiresOptIn"]
+  }
```
vRallev, line 13: This could be added to the root `build.gradle` file. Or would that be a bad idea?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilAnnotationDetectorCheck.kt
```diff
@@ -14,8 +17,11 @@ import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.psi.KtFile
 import java.io.File
 
+@AutoService(CodeGenerator::class)
```
vRallev, line 20: Should the internal code generators use this mechanism? Would that have a performance impact if no external code generator is used (98% of the cases)? Or wouldn't it matter, because we need to check for external generators anyways?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -69,6 +71,11 @@ internal class BindingModuleGenerator(
   private val classScanner: ClassScanner
 ) : CodeGenerator {
 
+  // This is actually not checked as we instantiate this manually
```
vRallev, line 74: I'd throw `NotImplementedError` with this message to make it crystal clear.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerators.kt
```diff
@@ -0,0 +1,9 @@
+package com.squareup.anvil.compiler.codegen
+
+import com.squareup.anvil.compiler.api.CodeGenerator
+import java.util.ServiceLoader
+
+internal fun loadCodeGenerators(): List<CodeGenerator> {
```
vRallev, line 6: Move this function into `CodeGenerationExtension` and delete the file.

compiler/src/main/java/com/squareup/anvil/compiler/RealAnvilContext.kt
```diff
@@ -0,0 +1,8 @@
+package com.squareup.anvil.compiler
```
vRallev, line 1: That file belongs into the `codegen` package.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -37,25 +27,16 @@ class AnvilComponentRegistrar : ComponentRegistrar {
...
+    val context = RealAnvilContext(generateDaggerFactories, generateDaggerFactoriesOnly)
+    val codeGenerators = loadCodeGenerators()
+      .filter { it.isApplicable(context) }
+      .toMutableList()
 
     if (!generateDaggerFactoriesOnly) {
-      codeGenerators += ContributesToGenerator()
-      codeGenerators += ContributesBindingGenerator()
-      codeGenerators += ContributesMultibindingGenerator()
+      // TODO do we just special case this one?
```
vRallev, line 38: Yes, please remove the TODO and add a comment why.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/CodeGenerator.kt
```diff
@@ -28,23 +47,41 @@ internal interface CodeGenerator {
...
   ): Collection<GeneratedFile> = emptyList()
 
-  data class GeneratedFile(
+  /**
+   * Represents a generated file that Anvil should eventually write.
+   *
+   * @property file the [File] to generate.
+   * @property content the file contents to write to [file].
+   */
+  public data class GeneratedFile(
```
vRallev, line 61: This class needs to be moved into a separate file and not be an inner class. But to blow up this PR not any further, this should be a separate PR. I've added a TODO in #265.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/CodeGenerator.kt
```diff
@@ -28,23 +47,41 @@ internal interface CodeGenerator {
...
     val content: String
   )
+
+  /**
+   * Contains context about the given Anvil compiler invocation. Considered a read-only API.
+   *
+   * These loosely correspond to the Anvil Gradle plugin's options, but we may add other information
+   * in the future.
+   */
+  public interface AnvilContext {
```
vRallev, line 72: Please create a separate file and not an inner class.

compiler/build.gradle
```diff
@@ -29,6 +30,6 @@ tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
 
     // The flag is needed because we extend an interface that uses @JvmDefault and the Kotlin
     // compiler requires this flag when doing so.
-    freeCompilerArgs += "-Xjvm-default=enable"
+    freeCompilerArgs += ["-Xjvm-default=enable", "-Xopt-in=com.squareup.anvil.annotations.ExperimentalAnvilApi"]
```
vRallev, line 33: Same here, I probably wouldn't mind adding this to the root file, but I can be convinced otherwise. What do you think?

compiler-api/build.gradle
```diff
@@ -0,0 +1,19 @@
...
+
+dependencies {
+  api project(':annotations')
+  api deps.kotlin.compiler
+}
+
+//noinspection UnnecessaryQualifiedReference
+tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).configureEach {
+  kotlinOptions {
+    jvmTarget = JavaVersion.VERSION_1_8
```
vRallev, line 16: The jvmTarget is set in the root build.gradle file.

vRallev (commented):
Okay, I had time to take a look. This PR is great, I'm really liking the direction. 

I'll start collecting some TODOs that come to my mind in #265  that should be addressed in a separate PR.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
ZacSweers: Yep left this for CR discussion 👍. Will update

annotations/build.gradle
ZacSweers: I tend to default to being explicit but don't have strong feelings about it if you'd rather it be in the root

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilAnnotationDetectorCheck.kt
ZacSweers: Wouldn't matter because you'd need to check for external generators anyway!

compiler/src/main/java/com/squareup/anvil/compiler/RealAnvilContext.kt
ZacSweers: Even though this is only used in the AnvilComponentRegistrar?

vRallev: Yes, the interface is in the same package and it makes sense to me.

annotations/build.gradle
vRallev: Ah, leave it as is if it's easier for you and you didn't make the change yet.

ZacSweers: Done!

compiler/build.gradle
vRallev: Ignore.

compiler/src/main/java/com/squareup/anvil/compiler/RealAnvilContext.kt
ZacSweers: Done!

ZacSweers:
Agreed on all the above!

> flush

I'll leave this one to you for a followup as the one implementation is nontrivial and I'm not too familiar to know where to move it to

> Maybe we should expose AnvilCompilationException?

Yeah that works, or offer a `Messager`-style API with an `error()` function+overloads to allow for collecting errors? We could pass it as a parameter to `generateCode()`

compiler-api/gradle.properties
vRallev, line 2: Oh, I think this got reverted? Shouldn't that be `compiler-api`?

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilContext.kt
```diff
@@ -0,0 +1,12 @@
...
+
+/**
+ * Contains context about the given Anvil compiler invocation. Considered a read-only API.
+ *
+ * These loosely correspond to the Anvil Gradle plugin's options, but we may add other information
+ * in the future.
+ */
+public interface AnvilContext {
+  public val generateFactories: Boolean
+  public val generateFactoriesOnly: Boolean
```
vRallev, line 11: I think it's worth to copy some parts from `AnvilExtension` to make it clearer what they mean. Sorry, I missed that initially.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -99,4 +82,9 @@ class AnvilComponentRegistrar : ComponentRegistrar {
       .getExtensionPoint(AnalysisHandlerExtension.extensionPointName)
       .registerExtension(extension, LoadingOrder.FIRST, project)
   }
+
+  private fun loadCodeGenerators(): List<CodeGenerator> {
+    return ServiceLoader.load(CodeGenerator::class.java, CodeGenerator::class.java.classLoader)
+      .toList()
+  }
```
vRallev, line 89: I suggested a different file and didn't realize where it's used 🤦

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
ZacSweers: I knew what you meant in spirit :)

ZacSweers:
CI flaked on wrapper validation again

vRallev:
Tested an internal build just to confirm that the new artifact doesn't cause any problems.

PR #269, opened 2021-05-17
Add anvil configuration for picking up custom code generators

ZacSweers:
Ref #265

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -13,6 +13,7 @@ import com.squareup.anvil.plugin.ProjectType.MULTIPLATFORM_ANDROID
 import com.squareup.anvil.plugin.ProjectType.MULTIPLATFORM_JVM
 import org.gradle.api.GradleException
 import org.gradle.api.Project
+import org.gradle.api.artifacts.Configuration
 import org.gradle.api.plugins.AppliedPlugin
 import org.gradle.api.plugins.PluginManager
 import org.gradle.api.provider.Provider
@@ -87,6 +88,12 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
   override fun apply(target: Project) {
     val extension = target.extensions.create("anvil", AnvilExtension::class.java)
 
+    // Create a configuration for collecting CodeGenerator dependencies.
+    val anvilConfiguration = target.configurations.maybeCreate("anvil").apply {
+      description = "This configuration is used for dependencies with Anvil CodeGenerator " +
+        "implementations."
+    }
+
     val once = AtomicBoolean()
 
     fun PluginManager.withPluginOnce(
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -134,6 +134,14 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
         project.dependencies.add("implementation", "$GROUP:annotations:$VERSION")
       }
     }
+
+    // Create a configuration for collecting CodeGenerator dependencies
+    val anvilConfiguration = project.configurations.maybeCreate("anvil").apply {
+      description = "This configuration is used for dependencies with Anvil CodeGenerator implementations."
+    }
+    // Make the kotlin compiler classpath extend it to pick up our extra generators
+    project.configurations.getByName("kotlinCompilerPluginClasspath")
```
vRallev, line 143: Use the constant `PLUGIN_CLASSPATH_CONFIGURATION_NAME`, it's already used in this file.

vRallev (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
ZacSweers: Ah nice, fixed

ZacSweers:
CI flaked on wrapper validation again

ZacSweers:
CI failure is a flake

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -163,6 +163,15 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
     }
+
+    // Create a configuration for collecting CodeGenerator dependencies
+    val anvilConfiguration = project.configurations.maybeCreate("anvil").apply {
+      description = "This configuration is used for dependencies with Anvil CodeGenerator " +
+        "implementations."
+    }
+    // Make the kotlin compiler classpath extend it to pick up our extra generators
+    project.configurations.getByName(PLUGIN_CLASSPATH_CONFIGURATION_NAME)
+      .extendsFrom(anvilConfiguration)
```
vRallev, line 174: I didn't catch this earlier. I think it would be better not delaying to add the configuration. I'd move the whole block up to place where the Anvil extension is created.

vRallev (changes requested):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
ZacSweers: I don't think we can because it looks up the kotlin plugin classpath configuration, which means it needs to wait until the kotlin plugin is applied. We could try to swing it with something like `configureEach` but I was intentionally just deferring it to the plugin application callback. Thoughts?

vRallev: The configuration has to be created earlier, but the compiler plugin classpath needs to be extended here. Note that this block is only called in an afterEvaluate for multiplatform projects and that's too late. I'll update the PR. I have the change ready because I was testing and confirming that.

ZacSweers: Ahh gotcha. Sounds good 👍

vRallev (approved):

PR #273, opened 2021-05-23
Implement generation-only mode

### LINKED ISSUE #258: Feature Request: Support generation-only mode
Currently, factory generation can be enabled in isolation but not other types of non-transforming generation like `@Contributes*` APIs. Talked offline a bit, it would be nice if there was a way to enable this mode of anvil while still disabling transformation (which requires disabling incremental stub generation).

ZacSweers:
Resolves #258

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
@@ -8,6 +8,7 @@ import com.squareup.anvil.compiler.codegen.ContributesBindingGenerator
 import com.squareup.anvil.compiler.codegen.ContributesMultibindingGenerator
 import com.squareup.anvil.compiler.codegen.ContributesToGenerator
 import com.squareup.anvil.compiler.codegen.dagger.AnvilAnnotationDetectorCheck
+import com.squareup.anvil.compiler.codegen.dagger.AnvilMergeAnnotationDetectorCheck
 import com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator
 import com.squareup.anvil.compiler.codegen.dagger.AssistedInjectGenerator
 import com.squareup.anvil.compiler.codegen.dagger.ComponentDetectorCheck
@@ -39,14 +40,19 @@ class AnvilComponentRegistrar : ComponentRegistrar {
 
     val codeGenerators = mutableListOf<CodeGenerator>()
     val generateDaggerFactoriesOnly = configuration.get(generateDaggerFactoriesOnlyKey, false)
+    val disableComponentMerging = configuration.get(disableComponentMergingKey, false)
 
     if (!generateDaggerFactoriesOnly) {
       codeGenerators += ContributesToGenerator()
       codeGenerators += ContributesBindingGenerator()
       codeGenerators += ContributesMultibindingGenerator()
       codeGenerators += BindingModuleGenerator(scanner)
-    } else {
+    } else if (!disableComponentMerging) {
       codeGenerators += AnvilAnnotationDetectorCheck()
+    }
+
+    if (disableComponentMerging) {
+      codeGenerators += AnvilMergeAnnotationDetectorCheck()
     }
 
     if (configuration.get(generateDaggerFactoriesKey, false)) {
@@ -64,7 +70,7 @@ class AnvilComponentRegistrar : ComponentRegistrar {
     //
     // It's not dangerous for our extension to run first, because we generate code, restart the
     // analysis phase and then don't return a result anymore. That means the next extension can
-    // take over. If we wouldn't do this and any other extension won't let our's run, then we
+    // take over. If we wouldn't do this and any other extension won't let ours run, then we
     // couldn't generate any code.
     AnalysisHandlerExtension.registerExtensionFirst(
       project,
@@ -74,7 +80,7 @@ class AnvilComponentRegistrar : ComponentRegistrar {
       )
     )
 
-    if (!generateDaggerFactoriesOnly) {
+    if (!generateDaggerFactoriesOnly && !disableComponentMerging) {
       SyntheticResolveExtension.registerExtension(
         project,
         InterfaceMerger(scanner)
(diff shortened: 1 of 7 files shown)


ZacSweers:
ktlint issue if a CI flake

ZacSweers:
I'm kind of wondering if we should flip this to just be something like a `enableTransformations` and have that replace the "factories only" mechanism. I don't know if there's any situation where one would just want factory gen and not contributesto if the annotation is used? Just spitballing ideas here on how to make the behaviors easy to grok

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -25,4 +25,13 @@ open class AnvilExtension {
...
   var generateDaggerFactoriesOnly = false
+
+  /**
+   * Enabling this indicates that only code generation should run and no transformations. This is
+   * useful for cases where you want to use `@ContributesTo`, `@ContributesBinding`, or similar
+   * annotations but _not_ `@MergeComponent` or `@MergeSubcomponent` functionality. This allows
+   * for anvil use in projects with kapt enabled but _not_ require disabling incremental compilation
+   * in kapt stub generation tasks.
+   */
+  var generationOnly = false
```
vRallev, line 36: How about `contributeOnly`? That would fit better with the language Anvil is using.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilComponentRegistrar.kt
```diff
@@ -64,7 +70,7 @@ class AnvilComponentRegistrar : ComponentRegistrar {
     //
     // It's not dangerous for our extension to run first, because we generate code, restart the
     // analysis phase and then don't return a result anymore. That means the next extension can
-    // take over. If we wouldn't do this and any other extension won't let our's run, then we
+    // take over. If we wouldn't do this and any other extension won't let ours run, then we
```
vRallev, line 73: Ha, thanks!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilMergeAnnotationDetectorCheck.kt
```diff
@@ -0,0 +1,43 @@
...
+import com.squareup.anvil.compiler.contributesToFqName
+import com.squareup.anvil.compiler.mergeComponentFqName
+import com.squareup.anvil.compiler.mergeInterfacesFqName
+import com.squareup.anvil.compiler.mergeModulesFqName
+import com.squareup.anvil.compiler.mergeSubcomponentFqName
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.psi.KtFile
+import java.io.File
+
+internal class AnvilMergeAnnotationDetectorCheck : PrivateCodeGenerator() {
```
vRallev, line 17: Great that you've added this check.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AnvilMergeAnnotationDetectorCheckTest.kt
```diff
@@ -0,0 +1,123 @@
...
+  @Test fun `@ContributesTo is allowed`() {
+    compile(
+      """
+      package com.squareup.test
+      
+      import dagger.Module
+      
+      @Module
+      @com.squareup.anvil.annotations.ContributesTo(Any::class)
+      object AnyClass {
```
vRallev, line 20: No curly braces needed.

vRallev:
> I'm kind of wondering if we should flip this to just be something like a `enableTransformations` and have that replace the "factories only" mechanism. I don't know if there's any situation where one would just want factory gen and not contributesto if the annotation is used? Just spitballing ideas here on how to make the behaviors easy to grok

I'd keep them separate for now. I've added "factories only" mainly for you, but I could see how other projects leverage that if they don't want to use Anvil but make use of the compiler plugin for faster compile times. The maintenance burden isn't too high.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
ZacSweers: maybe, but would that be confusing with generateDaggerFactories? Since both could be enabled at the same time. Maybe `disableComponentMerging` instead to be ultra explicit?

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AnvilMergeAnnotationDetectorCheckTest.kt
ZacSweers: 7d93172

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
ZacSweers: Would sort of line up semantically with the `AnvilMergeAnnotationDetectorCheck` too

ZacSweers: Tried that out and thought it read quite well. Let me know what you think 0fbccbb

vRallev: Reads well to me as well and is very explicit.

> Since both could be enabled at the same time.

Really? That doesn't makes sense. If you generate Dagger factories only, then it's redundant to set `disableComponentMerging` to true. I wonder if this even should be a check. Only one of the flags should be set to true.

vRallev: Another point: if you generate dagger factories only, then the Gradle plugin doesn't even add the Anvil annotations to the compile classpath.

vRallev (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
vRallev: Could be a follow up, what do you think?

ZacSweers: Ah no i just generate factories, not generate factories _only_. Does that make sense?

vRallev: Oh, damn. I somehow assumed you meant `generateDaggerFactoriesOnly`, but you clearly said `generateDaggerFactories`. Makes sense now 🤦🏻

PR #274, opened 2021-05-25
Respect qualifiers when checking whether there is a duplicate binding.

### LINKED ISSUE #270: ContributesBinding throwing error when 2 implementations are using 2 different qualifiers
Based on docs the code below should work, however I'm seeing "There are multiple contributed bindings with the same bound type." exception when compiling

```
@QualifierOne
@ContributesBinding(scope = SomeScope::class)
class QualifierOneImpl @Inject constructor(): MyInterface

@QualifierTwo
@ContributesBinding(scope = SomeScope::class)
class QualifierTwoImpl @Inject constructor(): MyInterface
```

Might be related to new checks here:
 https://github.com/square/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt#L502-L521

vRallev:
Fixes #270

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -31,6 +31,7 @@ import com.squareup.anvil.compiler.mergeModulesFqName
 import com.squareup.anvil.compiler.mergeSubcomponentFqName
 import com.squareup.anvil.compiler.priority
 import com.squareup.anvil.compiler.replaces
+import com.squareup.anvil.compiler.requireFqName
 import com.squareup.anvil.compiler.safePackageString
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -245,11 +248,7 @@ internal class BindingModuleGenerator(
               bindings
             } else {
               bindings
-                .groupBy {
-                  // It's safe to use the FqName here, because we don't allow generic super types.
-                  // If we wouldn't have this limitation, we'd need to use the Kotlin type.
-                  it.boundType.fqNameSafe
```
vRallev, line 251: That was the issue, we only looked at the class name and completely forgot about qualifiers.

0xMatthewGroves (approved):
Thank you!

PR #277, opened 2021-05-25
Make the CodeGenerator#flush() method internal to Anvil.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -68,14 +68,20 @@ internal class CodeGenerationExtension(
...
       .filterIsInstance<PrivateCodeGenerator>()
       .forEach { codeGenerator ->
         codeGenerator.generateCode(codeGenDir, module, flushedFiles)
-        codeGenerator.flush(codeGenDir, module)
+      }
+
+    codeGenerators
+      .filterIsInstance<FlushingCodeGenerator>()
+      .forEach {
+        it.flush(codeGenDir, module)
```
christiandeange, line 84: super nit: explicitly use `codeGenerator ->` as lambda parameter for consistency with above.

christiandeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
vRallev: Fair

PR #278, opened 2021-05-26
Move AnvilCompilationException into the compiler-api module.

vRallev:


compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
```diff
@@ -0,0 +1,96 @@
...
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.symbols.IrSymbol
+import org.jetbrains.kotlin.ir.util.render
+import org.jetbrains.kotlin.js.resolve.diagnostics.findPsi
+import org.jetbrains.kotlin.resolve.source.KotlinSourceElement
+import org.jetbrains.kotlin.resolve.source.getPsi
+import org.jetbrains.kotlin.util.getExceptionMessage
+
+/**
+ * `AnvilCompilationException` is thrown when Anvil specific can't be processed or an error
```
0xMatthewGroves, line 18: This reads strangely. What does "when Anvil specific can't be processed" mean?

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
```diff
@@ -0,0 +1,96 @@
...
+ * occurs while generating code.
+ */
+public class AnvilCompilationException(
+  message: String,
+  cause: Throwable? = null,
+  element: PsiElement? = null
+) : CompilationException(message, cause, element) {
+  public companion object {
```
0xMatthewGroves, line 26: Aren't the public visibility operators redundant?

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
```diff
@@ -0,0 +1,96 @@
...
+    ): AnvilCompilationException {
+      return AnvilCompilationException(
+        message = message,
+        cause = cause,
+        element = classDescriptor.identifier
+      )
+    }
+
+    public operator fun invoke(
+      message: String,
```
0xMatthewGroves, line 52: The ordering isn't standardized across these invocations. Could you align these? 

message
cause
object

0xMatthewGroves (approved):
A couple small nits but otherwise LGTM.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
vRallev: I forgot to type the word "code", "when Anvil specific code can't be processed".

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
vRallev: No, they're required with the explicit API mode enabled.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilCompilationException.kt
vRallev: Oh, good catch!

PR #279, opened 2021-05-26
Move utility functions to a new artifact.

vRallev:


compiler-utils/README.md
```diff
@@ -0,0 +1,5 @@
+# Compiler-Utils
+
+Contains several utility functions used within Anvil and external `CodeGenerator`s. There's no 
+API stability guaranteed for these functions. They can be changed or removed in every single
+release. So use at your own risk.
```
vRallev, line 5: That's important.

compiler/src/test/java/com/squareup/anvil/compiler/ReflectionTestUtils.kt
```diff
@@ -14,3 +15,14 @@ val Member.isAbstract: Boolean get() = Modifier.isAbstract(modifiers)
...
+
+inline fun <T, E : Executable> E.use(block: (E) -> T): T {
+  // Deprecated since Java 9, but many projects still use JDK 8 for compilation.
+  @Suppress("DEPRECATION")
+  val original = isAccessible
+
+  isAccessible = true
+  return block(this).also {
+    isAccessible = original
+  }
```
christiandeange, line 27: If `block()` throws an exception, your `also` block won't be run, and the element will remain accessible. Using a try/finally usually works around this:

```kotlin
return try {
  isAccessible = true
  block(this)
} finally {
  isAccessible = original
}
```

christiandeange (approved):

PR #282, opened 2021-05-26
[Draft] FqName resolution for generated code

### LINKED ISSUE #283: Generated types cannot be injected with factory generation
This is a follow-up to my comment here: https://github.com/square/anvil/pull/267#issuecomment-841836847

I would like to inject generated code into manually written constructors, with the generated code directory added to source sets thanks to #207.  The code is generated using a custom `CodeGenerator` implementation.  This doesn't work with factory generation enabled, because Anvil cannot resolve types for the generated code even if it has already been written.

For example, this code:
``` Kotlin
package com.squareup.test
      
import javax.inject.Inject
      
)
```

will generate this exception:
```
e: [...]/sources/src/main/java/com/squareup/test/Source.kt: (6, 23): com.squareup.anvil.compiler.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve fqName.
File being compiled: (6,23) in [...]/sources/src/main/java/com/squareup/test/Source.kt

	at com.squareup.anvil.compiler.codegen.KotlinPoetKt$requireTypeName$2.invoke(KotlinPoet.kt:124)
	at com.squareup.anvil.compiler.codegen.KotlinPoetKt.requireTypeName(KotlinPoet.kt:194)
	at com.squareup.anvil.compiler.codegen.KotlinPoetKt.mapToParameter(KotlinPoet.kt:266)
	at com.squareup.anvil.compiler.codegen.dagger.InjectConstructorFactoryGenerator.generateFactoryClass(InjectConstructorFactoryGenerator.kt:82)
	at com.squareup.anvil.compiler.codegen.dagger.InjectConstructorFactoryGenerator.generateCodePrivate(InjectConstructorFactoryGenerator.kt:58)
	at com.squareup.anvil.compiler.codegen.PrivateCodeGenerator.generateCode(PrivateCodeGenerator.kt:20)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension$analysisCompleted$2.invoke(CodeGenerationExtension.kt:61)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted(CodeGenerationExtension.kt:63)
```

It will always fail on the very first pass of the code generators, whenever `InjectConstructorFactoryGenerator` is invoked.

Because all code generation happens within a single invocation of `CodeGenerationExtension.analysisCompleted(...)`, the `ModuleDescriptor` argument never contains the generated files.

Ultimately, this means that `ModuleDescriptor.findClassOrTypeAlias` will fail to resolve an `FqName` for generated code.  In turn, this means that `PsiElement.requireFqName(module)` and `FqName.asClassName(module)` will fail.  Those two functions are used all over the place.

Because there are so many different paths, it would be really awkward to pass additional parameters (like the codeGenDir or a list of generated files) to those two functions.

The best solution I could come up with would be to introduce an `AnvilModuleDescriptor` type, which is basically just a decorator for the `ModuleDescriptor` provided by `CodeGenerationExtension`.  It would add a single function - `resolveFqNameOrNull` - which would replace the existing `findClassOrTypeAlias` (which is only ever used for the `FqName`).  This new function falls back to checking generated code for matches if the normal PSI methods return nothing.

I've put up a draft for review here: #282

RBusarow:
Fixes #283 

There are a few design choices I made here without strong opinions.

- Instead of a decorator, would it be better if `AnvilModuleDescriptor` was just a standalone context-like object with properties for files and the module?  That object would be passed around instead of the `ModuleDescriptor`.
- Instead of explicitly adding files as they're created, it could just take the `codeGenDir` and `PsiManager` as constructor arguments and then re-parse the files itself (and cache them) as needed.  That's less performant, but even for a monolithic module with a lot of generated code it's probably negligible?
- Instead of using normal `ModuleDescriptor` types everywhere with a type check in `ModuleDescriptor.resolveFqNameOrNull`, would it be better to use the custom type everywhere?  It would mean exposing the type in `compiler-api` and the diff is a whole lot bigger.

I'm not yet sure how to go about introducing an extra code generator for tests, but tests are on their way.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnvilModuleDescriptor.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnvilModuleDescriptor.kt
@@ -1,0 +1,44 @@
+package com.squareup.anvil.compiler.codegen
+
+import com.squareup.anvil.compiler.safePackageString
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.descriptors.findTypeAliasAcrossModuleDependencies
+import org.jetbrains.kotlin.descriptors.resolveClassByFqName
+import org.jetbrains.kotlin.incremental.components.NoLookupLocation.FROM_BACKEND
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+internal class AnvilModuleDescriptor(
+  private val delegate: ModuleDescriptor,
+  generatedKtFiles: List<KtFile>
+) : ModuleDescriptor by delegate {
+
+  private val generatedKtFiles: MutableList<KtFile> = generatedKtFiles.toMutableList()
+
+  fun addFiles(newFiles: Collection<KtFile>) {
+    generatedKtFiles.addAll(newFiles)
+  }
+
+  fun resolveFqNameOrNull(
+    packageName: FqName,
+    className: String
+  ): FqName? {
+
+    val maybeClassFqName = FqName("${packageName.safePackageString()}$className")
+
+    resolveClassByFqName(maybeClassFqName, FROM_BACKEND)
+      ?.let { return it.fqNameSafe }
+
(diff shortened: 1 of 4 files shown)


vRallev:
I took your draft as a base for #290. I really liked the solution. I kept your suggestion instead of the other options because it was the least amount of work, doesn't extend the `CodeGenerator` API and has the least amount of overhead. If I would start all over again with Anvil, I'd probably go with your first alternative.

PR #285, opened 2021-05-27
When trying to find an annotation, check if the annotation is part of the same package

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -111,6 +117,15 @@ public fun KtAnnotated.findAnnotation(fqName: FqName): KtAnnotationEntry? {
...
   if (hasStarImport) return annotationEntryShort
 
+  // At this point we know that the class is annotated with an annotation that has the same simple
+  // name as fqName. We couldn't find any imports, so the annotation is likely part of the same
+  // package or Kotlin namespace. Leverage our existing utility function and to find the FqName
+  // and then compare the result.
+  val fqNameOfShort = annotationEntryShort.fqNameOrNull(module)
+  if (fqName == fqNameOfShort) {
+    return annotationEntryShort
+  }
```
vRallev, line 127: That's the important change. One of my later changes will rely on this. I had to add the module descriptor as argument, what caused the cascading changes.

christiandeange (approved):
Probably needs a unit test to verify the behaviour, otherwise LGTM

vRallev:
> Probably needs a unit test to verify the behaviour, otherwise LGTM

That's difficult. I usually don't test these utilities separately. I test the output of the compiler plugin itself. I'm running into a case where this is needed with the integration tests. There it's indirectly tested.

christiandeange:
> That's difficult. I usually don't test these utilities separately. I test the output of the compiler plugin itself. I'm running into a case where this is needed with the integration tests. There it's indirectly tested.

You can't add a unit/integ test where the class being generated and an annotation on it have the same package name? Or am I misunderstanding what this is needed for?

vRallev:
> > That's difficult. I usually don't test these utilities separately. I test the output of the compiler plugin itself. I'm running into a case where this is needed with the integration tests. There it's indirectly tested.
> 
> You can't add a unit/integ test where the class being generated and an annotation on it have the same package name? Or am I misunderstanding what this is needed for?

That's happening here: https://github.com/square/anvil/pull/286/files#diff-d1fcbb8c18df7aabfa72fec5e0d89ed469b286abb2f3e5f45d74865276f314aa What I meant being difficult is creating a unit test that tests this function in isolation.

christiandeange:
> That's happening here: https://github.com/square/anvil/pull/286/files#diff-d1fcbb8c18df7aabfa72fec5e0d89ed469b286abb2f3e5f45d74865276f314aa What I meant being difficult is creating a unit test that tests this function in isolation.

Ah, I didn't realize #286 was based on this branch. No worries then!

christiandeange (approved):

PR #286, opened 2021-05-27
Setup an integration test module to verify the behavior of code generators

vRallev:


integration-tests/code-generator/src/main/java/com/squareup/anvil/test/TestCodeGenerator.kt
```diff
@@ -0,0 +1,51 @@
...
+
+        sequenceOf(
+          createGeneratedFile(
+            codeGenDir = codeGenDir,
+            packageName = generatedPackage,
+            fileName = "GeneratedClass",
+            content = generatedClass
+          )
+        )
+      }
```
vRallev, line 48: I'll extend this list later. It allows me to test various edge cases.

integration-tests/code-generator-tests/src/test/java/com/squareup/anvil/test/Utils.kt
```diff
@@ -0,0 +1,6 @@
+package com.squareup.anvil.test
+
+import kotlin.reflect.KClass
+
+internal infix fun KClass<*>.extends(other: KClass<*>): Boolean =
+  other.java.isAssignableFrom(this.java)
```
christiandeange, line 6: brb copying this into all of my projects because I can never remember the correct operand order for `isAssignableFrom`

settings.gradle
```diff
@@ -2,6 +2,8 @@ include ':annotations'
 include ':compiler'
 include ':compiler-api'
 include ':compiler-utils'
+include ':integration-tests:code-generator'
+include ':integration-tests:code-generator-tests'
```
christiandeange, line 6: Any particular reason the contents of this module couldn't just go into `/integration-tests/code-generator/src/test/java`?

christiandeange (approved):

settings.gradle
vRallev: Yes, because the main source set would need to be added to the Kotlin compiler plugin classpath. There's no separate classpath for the main and test source set.

integration-tests/code-generator-tests/src/test/java/com/squareup/anvil/test/Utils.kt
vRallev: Same :-D

PR #287, opened 2021-05-27
Support generating Dagger modules and component interfaces in custom code generators

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -182,31 +181,20 @@ internal class BindingModuleGenerator(
...
-      val bindingsReplacedInDaggerModules = contributedModuleAndInterfaceClasses
+      val bindingsReplacedInDaggerModules = contributedModuleClasses
         .asSequence()
-        .mapNotNull {
-          if (it.isAnvilModule()) {
-            // The generated Anvil Dagger module cannot replace bindings. Exclude it, because the
-            // descriptor can't be resolved since it's a generated class.
-            null
-          } else {
-            it.requireClassDescriptor(module)
```
vRallev, line 193: That was causing a bug. If another code generator generated a Dagger module, then we cannot resolve the descriptor for that. The reason is that we simulate multiple rounds of code generation. There is no API to pass to generated code to the compiler after each round. We can do this only once after all code has been generated.

integration-tests/code-generator/src/main/java/com/squareup/anvil/test/TestCodeGenerator.kt
```diff
@@ -37,13 +37,47 @@ class TestCodeGenerator : CodeGenerator {
...
+        @Language("kotlin")
+        val contributedModule = """
+          package $generatedPackage
+          
+          import com.squareup.anvil.annotations.ContributesTo
+          import dagger.Module 
+
+          @ContributesTo(Unit::class)
+          @Module
+          interface ContributedModule
```
vRallev, line 59: Without the fix in the first commit, this generated code would cause failures in the compiler plugin.

0xMatthewGroves (approved):

PR #290, opened 2021-05-28
Improve check if FqName exists

### LINKED ISSUE #283: Generated types cannot be injected with factory generation

vRallev:
Introduce a new API to check if a FqName can be resolved. In the implementation also check for the FqName in generated code, which is critical for some code generators.

Before we only resolved the class descriptor and based upon the result decided if the FqName is valid, which was never true for generated code.

Add an integration test for generating a class that is referenced in non-generated code.

Fixes #283, the solution is based on #282.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/AnvilModuleDescriptor.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/AnvilModuleDescriptor.kt
@@ -1,0 +1,31 @@
+package com.squareup.anvil.compiler.internal
+
+import com.squareup.anvil.annotations.ExperimentalAnvilApi
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.name.Name
+
+@ExperimentalAnvilApi
+public abstract class AnvilModuleDescriptor : ModuleDescriptor {
+  public abstract fun resolveClassIdOrNull(classId: ClassId): FqName?
+}
+
+public fun ModuleDescriptor.resolveFqNameOrNull(
+  fqName: FqName
+): FqName? = (this as AnvilModuleDescriptor).resolveClassIdOrNull(fqName.classIdBestGuess())
+
+public fun ModuleDescriptor.canResolveFqName(
+  fqName: FqName
+): Boolean = resolveFqNameOrNull(fqName) != null
+
+public fun ModuleDescriptor.resolveFqNameOrNull(
+  packageName: FqName,
+  className: String
+): FqName? = (this as AnvilModuleDescriptor)
+  .resolveClassIdOrNull(ClassId(packageName, Name.identifier(className)))
+
+public fun ModuleDescriptor.canResolveFqName(
+  packageName: FqName,
+  className: String
+): Boolean = resolveFqNameOrNull(packageName, className) != null
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
@@ -18,10 +18,22 @@ import java.io.File
 
 internal class CodeGenerationExtension(
   private val codeGenDir: File,
-  private val codeGenerators: List<CodeGenerator>
+  codeGenerators: List<CodeGenerator>
 ) : AnalysisHandlerExtension {
 
   private var didRecompile = false
+
+  private val codeGenerators = codeGenerators
+    .onEach {
+      check(it !is FlushingCodeGenerator || it !is PrivateCodeGenerator) {
+        "A code generator can't be a private code generator and flushing code generator at the " +
+          "same time. Private code generators don't impact other code generators, therefore " +
+          "they shouldn't need to flush files after other code generators generated code."
+      }
+    }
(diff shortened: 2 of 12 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/AnvilModuleDescriptor.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.jetbrains.kotlin.name.Name
+
+@ExperimentalAnvilApi
+public abstract class AnvilModuleDescriptor : ModuleDescriptor {
+  public abstract fun resolveClassIdOrNull(classId: ClassId): FqName?
+}
+
+public fun ModuleDescriptor.resolveFqNameOrNull(
+  fqName: FqName
+): FqName? = (this as AnvilModuleDescriptor).resolveClassIdOrNull(fqName.classIdBestGuess())
```
vRallev, line 16: This cast is the scary thing. But the code lives in the utils module, which gives no guarantees about compatibility and API stability. I suggest to read #283 and #282 to understand why this is needed.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -18,11 +18,23 @@ import java.io.File
...
+    .onEach {
+      check(it !is FlushingCodeGenerator || it !is PrivateCodeGenerator) {
+        "A code generator can't be a private code generator and flushing code generator at the " +
+          "same time. Private code generators don't impact other code generators, therefore " +
+          "they shouldn't need to flush files after other code generators generated code."
+      }
+    }
+    // Use a stable sort in case code generators depend on the order.
+    // At least don't make it random.
+    .sortedWith(compareBy({ it is PrivateCodeGenerator }, { it::class.qualifiedName }))
```
vRallev, line 36: A stable sort is important.

compiler/src/test/java/com/squareup/anvil/compiler/MergeModulesTest.kt
```diff
@@ -148,7 +148,7 @@ class MergeModulesTest {
       import com.squareup.anvil.annotations.compat.MergeModules
 
       @ContributesTo(Any::class)
-      abstract class DaggerModule1
+      abstract class DaggerModule2
```
vRallev, line 151: This was wrong and I don't know why it started to blow up now. Notice how the same class name was used below.

0xMatthewGroves (approved):

lmchew (approved):

PR #291, opened 2021-05-28
Replace some mapNotNull operators that can use new Kotlin 1.5 APIs.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -172,7 +172,7 @@ public inline fun <reified T> KtAnnotationEntry.findAnnotationArgument(
   // First check if the is any named parameter. Named parameters allow a different order of
   // arguments.
   annotationValues
-    .mapNotNull { valueArgument ->
+    .firstNotNullOfOrNull { valueArgument ->
```
0xMatthewGroves, line 175: Didn't realize I'd see this crazy method called in production code so soon 😄

0xMatthewGroves (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
lmchew: This name makes my eyes cross a little.

lmchew (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
vRallev: The language lead indirectly replied to this here: https://twitter.com/relizarov/status/1390691109093920768?s=20

lmchew: True, naming is hard.

PR #298, opened 2021-06-01
Use the correct replacement classes. 

vRallev:
This is a regression that was introduced while working on the new CodeGenerator API.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -190,6 +190,12 @@ internal class BindingModuleGenerator(
               annotation = contributesToFqName,
               scope = scope
             )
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -190,6 +190,12 @@ internal class BindingModuleGenerator(
               annotation = contributesToFqName,
               scope = scope
             )
+            .filter { it.annotationOrNull(daggerModuleFqName) != null }
+            .flatMap {
+              it.annotationOrNull(contributesToFqName, scope)
+                ?.replaces(module)
+                ?: emptyList()
+            }
```
vRallev, line 198: I simply forgot to add these lines back. I tried to write a test for that, but it's tricky. It requires to load descriptors from precompiled artifacts. I could do this in the integration tests, but it's kinda awkward to test the edge case there.

ZacSweers (approved):

PR #302, opened 2021-06-03
Fix replaced modules and bindings by excluded modules

### LINKED ISSUE #301: Excluded modules and contributed bindings should have their "replaces" ignored.
Given this example
```kotlin
@ContributesTo(AppScope::class, replaces = [DaggerModule2::class])
@Module
abstract class DaggerModule1

@Module
abstract class DaggerModule2 

@MergeComponent(
    scope = AppScope::class,
    exclude = [DaggerModule1::class]
)
interface ComponentInterface
```

One would expect `DaggerModule2` to be included in the generated component. However, it is not. Similar issues arise if the excluded module, contributed binding, or contributed multibinding specifies that it replaces another module or binding.

I'm working on a fix so that excluded modules and bindings that specify `replaces` have no effect.

bryanstern:
Fixes #301. Modules and bindings are no longer replaced if the replacement is declared by an excluded module.

Open to better ways of accomplishing this as I found the fq and psi class work a little tricky.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -181,9 +181,7 @@ internal class BindingModuleGenerator(
       val bindingsReplacedInDaggerModules = contributedModuleClasses
         .asSequence()
         .filter { it.scope(contributesToFqName, module) == scope }
-        .filterNot {
-          it.fqName in excludedTypesForScope[scope].orEmpty().map { it.fqNameSafe }
-        }
+        .filter { it.fqName !in excludedTypesForScope[scope].orEmpty().map { it.fqNameSafe } }
         .flatMap { it.replaces(contributesToFqName, module) }
         .plus(
           classScanner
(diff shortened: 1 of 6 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -88,6 +88,7 @@ internal class BindingModuleGenerator(
   private val mergedScopes = mutableMapOf<FqName, MutableList<Pair<File, KtClassOrObject>>>()
     .withDefault { mutableListOf() }
 
+  // excludes DaggerModule1
```
ZacSweers, line 91: nit: leftover

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
```diff
@@ -675,6 +675,36 @@ class ModuleMergerTest(
     }
   }
 
+  @Test fun `contributed modules cannot be replaced by excluded modules`() {
```
vRallev, line 678: Hm, the CI tests seem to be configured wrong. This test fails with the IR backend enable. You still need to update `ModuleMergerIr`. To test that update `TestUtils#compile` for now and set the `useIR` flag to true.

vRallev (changes requested):

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
vRallev: Oh, I see it now 
image

bryanstern:
Thanks for suggestions and speedy review. Let me know what you think.

vRallev (approved):
Nice, great contribution. Thank you!

PR #304, opened 2021-06-03
Add a workingDir param to anvilCompile for test folder access

ZacSweers:
kotlin-compile-testing under the hood will otherwise create tmp dirs that are impossible to get manually without access to the `KotlinCompilation`. This allows for customizing that, such as with the parent test's own `TemporaryFolder`.


ZacSweers:
For exercising this - what would you think of a base test class for anvil compile tests that has a `TemporaryFolder` that's used for these?

vRallev (approved):

PR #305, opened 2021-06-03
Add generatorComment parameter to buildFile + use canonical KotlinPoet API

ZacSweers:
This way custom generators can denote their own comments for tracing where their files came from. This also updates it to use the canonical KotlinPoet API for this. Default value is Anvil's usual comment.


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
```diff
@@ -258,9 +251,9 @@ public fun FileSpec.Companion.buildFile(
 
       block()
     }
+    .addComment(generatorComment)
```
vRallev, line 254: I thought I tested this in the past and here we are 🤷🏻

vRallev (approved):

PR #306, opened 2021-06-03
Add rawTypeFilter to KotlinType.asTypeName()

ZacSweers:
This optimization allows for short-circuiting the type resolution if the desired raw type isn't applicable

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
@@ -185,7 +185,7 @@ public fun KtTypeReference.requireTypeName(
 
 @ExperimentalAnvilApi
 public fun KotlinType.asTypeName(): TypeName {
-  return asTypeName { true }!!
+  return asTypeNameOrNull { true }!!
 }
 
 /**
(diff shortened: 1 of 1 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
```diff
@@ -185,9 +185,22 @@ public fun KtTypeReference.requireTypeName(
...
+  return asTypeName { true }!!
+}
+
+/**
+ * @param rawTypeFilter an optional raw type filter to allow for
+ *                      short-circuiting this before attempting to
+ *                      resolve type arguments.
+ */
+@ExperimentalAnvilApi
+public fun KotlinType.asTypeName(rawTypeFilter: (ClassName) -> Boolean = { true }): TypeName? {
```
vRallev, line 197: Hm, what's the use case for that? I also don't really like how the method is overloaded and returns a nullable value here and the other is non-nullable. That seems inconsistent.

ZacSweers: The use-case is to avoid type resolution of all type parameters if the raw type isn't desired. Would maybe calling this `findTypeName` be better?

vRallev: That's a Gradle idiom. `asTypeNameOrNull` would be better for Kotlin.

ZacSweers: That works for me!

ZacSweers: 5235f09

vRallev (approved):

PR #307, opened 2021-06-05
Refactor the Gradle plugin and create multiple Anvil configurations.

vRallev:
Rework how the Gradle plugin sets up its configuration, Gradle tasks and patches for other Gradle tasks. Instead of querying the tasks manually and checking which plugins are applied, use the callback methods from the Kotlin Gradle Plugin. With this change we can remove a lot from our custom logic.

This also allows us to easily create an Anvil configuration for each build variant separately. This allows to add custom code generators to specific build types only.

This refactoring is needed for #100.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -1,25 +1,16 @@
 package com.squareup.anvil.plugin
 
-import com.android.build.gradle.AppExtension
-import com.android.build.gradle.AppPlugin
-import com.android.build.gradle.LibraryExtension
-import com.android.build.gradle.LibraryPlugin
-import com.android.build.gradle.TestExtension
-import com.android.build.gradle.TestedExtension
 import com.android.build.gradle.api.BaseVariant
-import com.squareup.anvil.plugin.ProjectType.ANDROID
-import com.squareup.anvil.plugin.ProjectType.JVM
-import com.squareup.anvil.plugin.ProjectType.MULTIPLATFORM_ANDROID
-import com.squareup.anvil.plugin.ProjectType.MULTIPLATFORM_JVM
+import com.android.build.gradle.internal.tasks.factory.dependsOn
 import org.gradle.api.Action
 import org.gradle.api.GradleException
 import org.gradle.api.Project
 import org.gradle.api.Task
+import org.gradle.api.UnknownTaskException
 import org.gradle.api.artifacts.Configuration
-import org.gradle.api.plugins.AppliedPlugin
-import org.gradle.api.plugins.PluginManager
 import org.gradle.api.provider.Provider
-import org.jetbrains.kotlin.gradle.dsl.KotlinMultiplatformExtension
+import org.gradle.api.tasks.TaskContainer
+import org.gradle.api.tasks.TaskProvider
 import org.jetbrains.kotlin.gradle.internal.KaptGenerateStubsTask
 import org.jetbrains.kotlin.gradle.plugin.FilesSubpluginOption
 import org.jetbrains.kotlin.gradle.plugin.KaptExtension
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -1,60 +1,114 @@
...
-    return kotlinCompilation.target.project.plugins.hasPlugin(AnvilPlugin::class.java)
-  }
+  override fun apply(target: Project) {
+    val extension = target.extensions.create("anvil", AnvilExtension::class.java)
 
-  override fun getCompilerPluginId(): String = "com.squareup.anvil.compiler"
+    // Create a configuration for collecting CodeGenerator dependencies. We need to create all
+    // configurations eagerly and cannot wait for applyToCompilation(..) below, because this
+    // function is called in an afterEvaluate block by the Kotlin Gradle Plugin. That's too late
+    // to register the configurations.
```
vRallev, line 43: This is so unfortunate.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -1,60 +1,114 @@
...
     kotlinCompilation: KotlinCompilation<*>
   ): Provider<List<SubpluginOption>> {
-    val project = kotlinCompilation.target.project
+    val variant = Variant(kotlinCompilation)
+    val project = variant.project
+
+    // Make the kotlin compiler classpath extend our configurations to pick up our extra
+    // generators.
+    project.configurations.getByName(variant.compilerPluginClasspathName)
+      .extendsFrom(getConfiguration(project, variant.name))
```
vRallev, line 100: Here I'm making the connection between the Anvil configuration and the Kotlin compiler plugin classpath.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -88,233 +142,123 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
-            realApply(target, MULTIPLATFORM_ANDROID, extension, anvilConfiguration)
-          targets?.any { it.platformType == jvm } == true ->
-            realApply(target, MULTIPLATFORM_JVM, extension, anvilConfiguration)
-          else -> throw GradleException(
-            "Multiplatform project $target is not setup for Android or JVM."
-          )
-        }
-      }
-    }
-    target.pluginManager.withPluginOnce("org.jetbrains.kotlin.jvm") {
```
vRallev, line 137: All these workarounds with applying the plugin once or checking which plugin is applied are no longer needed.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -88,233 +142,123 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
-        }
-      }
   }
 
-  private fun disableIncrementalKotlinCompilation(
-    project: Project,
-    projectType: ProjectType,
-    extension: AnvilExtension
-  ) {
+  private fun disableIncrementalKotlinCompilation(variant: Variant) {
```
vRallev, line 188: This method became so much simpler.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -423,3 +307,98 @@ private fun <T : Task> T.doFirstCompat(block: (T) -> Unit) {
     }
   })
 }
+
+private inline fun <reified T : Task> Project.namedLazy(
```
vRallev, line 311: Is there something like this built in?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -88,233 +142,123 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
     // This will trigger a full compilation of a module using Anvil even though its
     // source files might not have changed. This workaround is necessary, otherwise
     // incremental builds are broken. See https://youtrack.jetbrains.com/issue/KT-38570
-    val disableIncrementalCompilationTaskProvider = project.tasks.register(
+    val disableIncrementalCompilationTaskProvider = variant.project.tasks.register(
       compileTaskName + "CheckIncrementalCompilationAnvil",
       DisableIncrementalCompilationTask::class.java
     ) { task ->
       task.pluginClasspath.from(
-        project.configurations.getByName(PLUGIN_CLASSPATH_CONFIGURATION_NAME)
```
vRallev, line 301: Here was a very tricky bug. The code is correct with Kotlin 1.4, because there was only a single Kotlin compiler plugin classpath. What this code does is to disable incremental compilation, if the compiler plugin classpath changed. This is for example important when we bump the Anvil version, because new or different might need to be generated.

With Kotlin 1.5 this classpath is empty and they generate compiler plugin classpaths for every single variant (main, test, android test, debug, release, etc.) and add our compiler plugin to these specific classpaths. But here we kept watching the old empty classpath, but there can never be a change. 

That was tricky to track down :-/

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -423,3 +307,98 @@ private fun <T : Task> T.doFirstCompat(block: (T) -> Unit) {
...
 }
+
+private inline fun <reified T : Task> Project.namedLazy(
+  name: String,
+  crossinline action: (TaskProvider<T>) -> Unit
+) {
+  try {
+    action(tasks.named(name, T::class.java))
+    return
+  } catch (ignored: UnknownTaskException) {
```
0xMatthewGroves, line 318: 🙈

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -423,3 +307,98 @@ private fun <T : Task> T.doFirstCompat(block: (T) -> Unit) {
...
+    if (!didRun) {
+      throw GradleException("Didn't find task $name with type ${T::class}.")
+    }
+  }
+}
+
+/**
+ * Runs the given [action] for each Android variant including androidTest and unit test variants.
+ */
+private fun Project.androidVariantsConfigure(action: (BaseVariant) -> Unit) {
```
0xMatthewGroves, line 340: 👨‍🍳

gradle-plugin/src/main/java/com/squareup/anvil/plugin/CheckMixedSourceSet.kt
```diff
@@ -47,31 +45,18 @@ object CheckMixedSourceSet {
...
-    return if (project.isAndroidProject) {
-      project.androidVariants()
-        .findVariantForCompileTask(compileTask)
-        .sourceSets
-        .flatMap { it.javaDirectories }
-        .let { project.files(it) }
-    } else {
-      null
-    }
+  private fun getAndroidSourceDirs(variant: Variant): FileCollection? {
```
0xMatthewGroves, line 48: Nice extension cleanup.

0xMatthewGroves (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
0xMatthewGroves: Maybe ask trobalik?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -88,233 +142,123 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
+        CheckMixedSourceSet.disablePreciseJavaTrackingIfNeeded(compileTask, result)
 
-      if (!extension.generateDaggerFactories.get() &&
-        extension.generateDaggerFactoriesOnly.get()
-      ) {
-        throw GradleException(
-          "You cannot set generateDaggerFactories to false and generateDaggerFactoriesOnly " +
-            "to true at the same time."
+        compileTask.logger.info(
+          "Anvil: Use precise java tracking: ${compileTask.usePreciseJavaTracking}"
```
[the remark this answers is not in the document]
vRallev: No, slf4j has support for Markers. I'll look into that separately.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -88,233 +142,123 @@ open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
...
-          // this workaround we could make changes in any library, but these changes wouldn't be
-          // contributed to the Dagger graph, because incremental compilation tricked us.
-          stubsTask.doFirstCompat {
-            stubsTask.incremental = false
-            stubsTask.logger.info(
-              "Anvil: Incremental compilation enabled: ${stubsTask.incremental} (stub)"
+    variant.project.pluginManager.withPlugin("org.jetbrains.kotlin.kapt") {
+      variant.project
+        .namedLazy<KaptGenerateStubsTask>(
+          "kaptGenerateStubs${variant.taskSuffix}"
```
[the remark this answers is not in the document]
vRallev: I have to do this, because this
```kotlin
.namedLazy<KaptGenerateStubsTask>("kaptGenerateStubs${variant.taskSuffix}") { stubsTaskProvider ->
```
is too long.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: Wait for the next PR that became a lot simpler because of that.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: Asked him, there's no built-in solution.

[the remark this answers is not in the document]
vRallev: Done, good point.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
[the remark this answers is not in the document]
vRallev: https://github.com/square/anvil/pull/313

PR #312, opened 2021-06-07
Introduce a variant filter

### LINKED ISSUE #100: Factory generation: provide a way to disable it in androidTest

vRallev:
Introduce a new `VariantFilter` for the Gradle extension. This API allows you to enable and disable Anvil for specific variants. 

Allow to override `generateDaggerFactories`, `generateDaggerFactoriesOnly` and `disableComponentMerging` through the variant filter.

This resolves #100.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -31,9 +31,12 @@ import org.jetbrains.kotlin.gradle.plugin.mpp.KotlinJvmAndroidCompilation
 import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
 import java.io.File
 import java.util.Locale.US
+import java.util.concurrent.ConcurrentHashMap
 
 @Suppress("unused")
 internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
+
+  private val variantCache = ConcurrentHashMap<String, Variant>()
 
   override fun apply(target: Project) {
     val extension = target.extensions.create("anvil", AnvilExtension::class.java)
@@ -84,7 +87,8 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
 
   override fun isApplicable(kotlinCompilation: KotlinCompilation<*>): Boolean {
     return when (kotlinCompilation.platformType) {
-      androidJvm, jvm -> true
+      // If the variant is ignored, then don't apply the compiler plugin.
+      androidJvm, jvm -> !getVariant(kotlinCompilation).variantFilter.ignore
       else -> false
     }
   }
@@ -92,7 +96,7 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
   override fun applyToCompilation(
     kotlinCompilation: KotlinCompilation<*>
   ): Provider<List<SubpluginOption>> {
-    val variant = Variant(kotlinCompilation)
+    val variant = getVariant(kotlinCompilation)
     val project = variant.project
 
     // Make the kotlin compiler classpath extend our configurations to pick up our extra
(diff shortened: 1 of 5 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -285,6 +286,22 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
         "implementations."
     }
   }
+
+  private val variantCache = mutableMapOf<String, Variant>()
```
christiandeange, line 286: nit: move this to the top with other `val`s?

christiandeange: Also you might be able to just make the map itself synchronized:

```kotlin
private val variantCache = ConcurrentHashMap<String, Variant>()
```

and then can drop the `@Synchronized` method below?

integration-tests/tests/build.gradle
```diff
@@ -6,6 +6,14 @@ kotlin {
   explicitApi()
 }
 
+anvil {
+  variantFilter {
+    if (name == 'main') {
+      ignore = true
+    }
```
christiandeange, line 13:
```groovy
anvil {
  variantFilter {
    ignore = (name == 'main')
  }
}
```

?

christiandeange (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: I made it specifically so to ensure that the filter action is only executed once. `getOrPut` on a `ConcurrentHashMap` is not threadsafe. The documentation calls this out as well:
```
Note that the operation is not guaranteed to be atomic if the map is being modified concurrently.
```

christiandeange: Ah sorry, forgot to mention that at that point you would also use `computeIfAbsent` instead of `getOrPut`, which is a method of `ConcurrentHashMap` and not an extension.

vRallev: Ah, right. That works!

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -388,23 +401,33 @@ internal class Variant private constructor(
...
+        variantFilter = variantFilter
       ).also {
         // Sanity check.
         check(it.compileTaskProvider.name.startsWith("compile"))
       }
     }
   }
 }
+
+private val KotlinCompilation<*>.identifier: String get() = name
```
0xMatthewGroves, line 433: The Variant constructor takes in a parameter named "name", why'd you change it?

0xMatthewGroves: Actually, CommonFilter too 🙈

gradle-plugin/src/main/java/com/squareup/anvil/plugin/VariantFilter.kt
```diff
@@ -0,0 +1,71 @@
+package com.squareup.anvil.plugin
```
0xMatthewGroves, line 1: Verryyyyy clean! 👨‍🍳

0xMatthewGroves (approved):

integration-tests/tests/build.gradle
0xMatthewGroves: ?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: My thinking was to have a proxy for the getter. But it's not needed and I don't see a big benefit either anymore. Will change.

PR #313, opened 2021-06-07
Add a Marker to logging messages.

vRallev:


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -164,7 +164,7 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
         // and minimize the IO at configuration time.
         CheckMixedSourceSet.disablePreciseJavaTrackingIfNeeded(compileTask, result)
 
-        compileTask.logger.info(
+        compileTask.log(
           "Anvil: Use precise java tracking: ${compileTask.usePreciseJavaTracking}"
```
0xMatthewGroves, line 168: Being able to add the marker means you don't need to prefix these messages with "Anvil:" anymore, correct?

0xMatthewGroves (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: The marker actually doesn't do anything. I'll remove it again, but I'll keep the separate `log` method.

PR #315, opened 2021-06-08
make non-test variants extend common anvil config

### LINKED ISSUE #314: non-test Anvil dependency configurations don't extend the main configuration
This is using the current 2.3.1-SNAPSHOT with new Anvil configurations.

AGP creates a `debug` variant by default, and `AnvilPlugin` is correctly creating an `anvilDebug` configuration.  However, this `anvilDebug` configuration does not extend the main/common configuration, so it doesn't service load any `CodeGenerator` which is added via `anvil(...)`.  Dependencies need to be added explicitly via `anvilDebug(...)`.

RBusarow:
fixes #314


vRallev (approved):
Thanks!

Issue #310, opened 2021-06-07
PsiUtils' requireClassDescriptor fails to resolve generated types


RBusarow:
Given this generated class:

``` Kotlin
@ContributesBinding(AppScope::class)
class GeneratedFooImpl @Inject constructor() : Foo
```

Both versions of `requireClassDescriptor` will fail to resolve `GeneratedFooImpl`.  This happens with any of the contributes/merge generators.

The new `RealAnvilModuleDescriptor` is being used, but it doesn't do any good since we're trying to get the `ClassDescriptor` instead of just an `FqName`.

Some `requireClassDescriptor` calls can be replaced with existing Psi calls, like when getting the scope argument in `BindingGenerator`.  `KtClassOrObject.checkClassExtendsBoundType` seems much more awkward, though, since that's trying to resolve super types.

vRallev:
This one is challenging. Replacing all `requireClassDescriptor` calls in `BindingModuleGenerator` isn't possible, likely. I can make the situation better for `ContributesBindingGenerator` though. As a workaround I suggest you generate a Dagger module with a binding method.

RBusarow:
Thanks for your updates and the suggestion.  I have to admit, I kind of forgot that normal Dagger modules exist!

I'm able to do nearly everything I need now, leaning on a vanilla Dagger module.  I do have some more changes to `BindingModuleGenerator` and how we get the scope and exclude FqName.  I'll try to get the PR out this afternoon.

PR #318, opened 2021-06-09
use Psi APIs to resolve scope and exclude arguments

RBusarow:
partial fix for #310


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -136,7 +136,7 @@ public fun KtClassOrObject.scope(
     .let { classLiteralExpression ->
       if (classLiteralExpression == null) {
         throw AnvilCompilationException(
-          "The first argument for $annotationFqName must be a class literal: $text",
+          "Couldn't find scope for $annotationFqName.",
```
RBusarow, line 139: There's a test for this in `ModuleMergerTest`.  This change makes it so that the two `scope` extension functions use the same message.

vRallev (approved):
This is great. I've added a commit that removes some duplication and avoids resolving types multiple times.

PR #322, opened 2021-06-13
Test merging multiple components with the same scope

### LINKED ISSUE #321: Excluded bindings are excluded from all components with the same scope.
Given the following example
```kotlin
interface ParentInterface
      
@ContributesBinding(Any::class)
interface ContributingInterface : ParentInterface
      
@MergeComponent(Any::class)
interface ComponentInterface
      
@MergeComponent(
  scope = Any::class,
  exclude = [ContributingInterface::class]
)
interface ComponentInterface2
```

We'd expect that the `ContributingInterface` would be present in `ComponentInterface`, but not `ComponentInterface2`. However, it is getting excluded in both. I don't believe this is a regression.

vRallev:
Ensure that excluded bindings and modules are only excluded for the specific component that merges a scope and not all components merging the same scope.

Fixes #321

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -84,10 +84,9 @@ internal class BindingModuleGenerator(
 
   // Keeps track of for which scopes which files were generated. Usually there is only one file,
   // but technically there can be multiple.
-  private val mergedScopes = mutableMapOf<FqName, MutableList<Pair<File, KtClassOrObject>>>()
-    .withDefault { mutableListOf() }
+  private val mergedClasses = mutableMapOf<MergedClassKey, File>()
 
-  private val excludedTypesForScope = mutableMapOf<FqName, List<FqName>>()
+  private val excludedTypesForClass = mutableMapOf<MergedClassKey, List<FqName>>()
 
   private val contributedBindingClasses = mutableListOf<FqName>()
   private val contributedMultibindingClasses = mutableListOf<FqName>()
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -84,10 +84,9 @@ internal class BindingModuleGenerator(
 
   // Keeps track of for which scopes which files were generated. Usually there is only one file,
   // but technically there can be multiple.
-  private val mergedScopes = mutableMapOf<FqName, MutableList<Pair<File, KtClassOrObject>>>()
-    .withDefault { mutableListOf() }
+  private val mergedClasses = mutableMapOf<MergedClassKey, File>()
 
-  private val excludedTypesForScope = mutableMapOf<FqName, List<FqName>>()
+  private val excludedTypesForClass = mutableMapOf<MergedClassKey, List<FqName>>()
```
vRallev, line 89: The bug is here where we collected all excluded types per scope rather than class with a merge annotation.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -565,3 +564,28 @@ private fun Binding.key(module: ModuleDescriptor): String {
 
   return fqName + allArguments
 }
+
+private class MergedClassKey(
```
christiandeange, line 568: `data class` doesn't suffice for your needs?

christiandeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
christiandeange: @vRallev I think you missed this comment

vRallev: Ah, sorry. Notice that I ignore `clazz` in the methods and use its FqName instead. That wouldn't work with a `data class`.

PR #323, opened 2021-06-14
Remove the accidental dependency on the Android Gradle Plugin.

vRallev:


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -7,7 +7,6 @@ import com.android.build.gradle.TestedExtension
 import com.android.build.gradle.api.BaseVariant
 import com.android.build.gradle.api.TestVariant
 import com.android.build.gradle.api.UnitTestVariant
-import com.android.build.gradle.internal.tasks.factory.dependsOn
```
vRallev, line 10: Oopsie.

christiandeange (approved):

Issue #326, opened 2021-06-22
Unable to generate @AssistedFactory from custom compiler


romankivalin:
This is similar to the #310, but happens with the `AssistedFactoryGenerator`.

Custom code generator creates assisted factories for view models like this:

```kotlin
@AssistedFactory
public interface ExampleViewModel_AssistedFactory : ViewModelFactory<ExampleViewModel, ExampleState> {
  public override fun create(initialState: ExampleState): ExampleViewModel
}
```

We have also enabled `generateDaggerFactories`. Then `AssistedFactoryGenerator` fails to process generated file with the following exception:

<details>
<summary>
e: com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve class for com.example.ExampleViewModel_AssistedFactory.
</summary>

```
e: com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve class for com.example.ExampleViewModel_AssistedFactory.
File being compiled: (12,1) in /com/example/ExampleViewModel_AssistedFactory.kt

	at com.squareup.anvil.compiler.internal.PsiUtilsKt.requireClassDescriptor(PsiUtils.kt:499)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator.generateFactoryClass(AssistedFactoryGenerator.kt:79)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator.generateCodePrivate(AssistedFactoryGenerator.kt:69)
	at com.squareup.anvil.compiler.codegen.PrivateCodeGenerator.generateCode(PrivateCodeGenerator.kt:21)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted$generateCode(CodeGenerationExtension.kt:89)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted(CodeGenerationExtension.kt:102)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration$invokeExtensionsOnAnalysisComplete(TopDownAnalyzerFacadeForJVM.kt:111)
	at org.jetbrains.kotlin.cli.jvm.compiler.TopDownAnalyzerFacadeForJVM.analyzeFilesWithJavaIntegration(TopDownAnalyzerFacadeForJVM.kt:121)
[...]

PR #328, opened 2021-06-22
Only call private code generators once after all other code generators are done

vRallev:
This could potentially help with #310 and #326.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
@@ -69,6 +69,9 @@ internal class CodeGenerationExtension(
     val anvilModule = RealAnvilModuleDescriptor(module)
     anvilModule.addFiles(files)
 
+    val (privateCodeGenerators, nonPrivateCodeGenerators) =
+      codeGenerators.partition { it is PrivateCodeGenerator }
+
     fun Collection<GeneratedFile>.toKtFile(): Collection<KtFile> {
       return this
         .mapNotNull { (file, content) ->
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/RealAnvilModuleDescriptor.kt
```diff
@@ -15,11 +15,15 @@ class RealAnvilModuleDescriptor(
...
 ) : AnvilModuleDescriptor, ModuleDescriptor by delegate {
 
+  internal val allFiles = mutableListOf<KtFile>()
+
   private val classesMap = mutableMapOf<String, List<KtClassOrObject>>()
   private val allClasses: Sequence<KtClassOrObject>
     get() = classesMap.values.asSequence().flatMap { it }
 
   fun addFiles(files: Collection<KtFile>) {
+    allFiles.addAll(files)
```
christiandeange, line 25:
```suggestion
    allFiles += files
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -69,6 +69,8 @@ internal class CodeGenerationExtension(
     val anvilModule = RealAnvilModuleDescriptor(module)
     anvilModule.addFiles(files)
 
+    val nonPrivateCodeGenerators = codeGenerators.filter { it !is PrivateCodeGenerator }
```
christiandeange, line 72: You could do something like:

```kotlin
val (privateCodeGenerators, nonPrivateCodeGenerators) =
  codeGenerators.partition { it is PrivateCodeGenerator }
```
to separate out the two kinds of code generators, and then use `privateCodeGenerators` instead of the extra `codeGenerators.filterIsInstance<PrivateCodeGenerator>()` pass below. Up to you.

christiandeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
vRallev: I had no idea, that's gorgeous.

PR #329, opened 2021-06-23
Add a README quote joke

zach-klippenstein:
Some of our open source repos' readmes start with silly fake quotes. I came up with one for Anvil, feel free to reject this PR without cause.


vRallev (approved):
Nice!

PR #330, opened 2021-06-24
Don't use `whenTaskAdded`, which causes all tasks for the Gradle module to be realized

vRallev:


autonomousapps (approved):
If it's green, it looks good to me!

christiandeange (approved):
thank u gradle

PR #338, opened 2021-07-01
Improve the error message when a type must be explicitly declared.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -474,7 +474,11 @@ public fun KtCallableDeclaration.requireTypeReference(module: ModuleDescriptor):
     )
   }
 
-  throw AnvilCompilationException("Couldn't obtain type reference.", element = this)
+  throw AnvilCompilationException(
+    message = "Couldn't obtain type reference. Did you declare a type for ${this.name} " +
+      "explicitly?",
```
ZacSweers, line 479: Just a suggestion that I think is a bit easier to grok at a glance

```suggestion
    message = "Couldn't obtain type reference. Did you declare an explicit type for ${this.name}?",
```

ZacSweers (approved):

PR #370, opened 2021-09-03
Make `anvilMain()` extend the `anvil()` configuration in JVM modules.

### LINKED ISSUE #368: Custom compiler extension not picked up in library module (just app module
I'm trying to implement the custom compiler extension.
If I have my custom annotation in :app module then codegen works as expected.
If I have my custom annotation in some :library module, then it is not picked up, even though `anvil project("custom-compiler")`

Here is repro with `anvil` command commented out in :app/build.gradle, 
https://github.com/ursusursus/AnvilCompilerTest

if I comment it back in, you'll notice debugger now hitting break points like so
image

If you F9 over it, codeGenDir always only talks about  `\app\build...`
If I observe the log from the console after using the `gradlew clean build --no-daemon -Dorg.gradle.debug=true -Dkotlin.compiler.execution.strategy="in-process" -Dkotlin.daemon.jvm.options="-Xdebug,-Xrunjdwp:transport=dt_socket,address=5005,server=y,suspend=n"` command, I notice it executing `:library1:...` without breaking being hit; all pointing to SampleCodeGenerator not being executed in `:library1`

vRallev:
Due to this bug custom code generators weren't picked up in JVM modules if you used `anvil('my-code-generator')'`.

Fixes #368

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -73,6 +73,15 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
         }
       }
     }
+
+    jvmPlugins.forEach { javaPlugin ->
+      target.pluginManager.withPlugin(javaPlugin) {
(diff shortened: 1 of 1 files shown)


0xMatthewGroves (approved):
LGTM.

Issue #379, opened 2021-10-07
2.3.5 treats all constant references in qualifiers as enums


ZacSweers:
This snippet is slightly wrong as qualified references could also be string/int/etc constants imported to there.

https://github.com/square/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/codegen/KotlinPoetUtils.kt#L188

In our repo this is easy to reproduce by using a named string constant in `@Named`

![image](https://user-images.githubusercontent.com/1361086/136454793-0890de27-1b9c-434e-ab78-1215d1d4c469.png)

![image](https://user-images.githubusercontent.com/1361086/136454818-1754d2f3-66b0-4508-b488-45b7c40a849d.png)

PR #384, opened 2021-10-11
Add more test cases + implement support for importing constant members in annotation specs

ZacSweers:
This mostly completes #379, but there's still a known undefined behavior with imports in the same package but different files (I don't know what to do about that case)


ZacSweers:
https://github.com/square/anvil/runs/3863054821 looks like a flake with the validator action

PR #405, opened 2021-10-30
Small clean ups for testing utilities

vRallev:


compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/TestUtils.kt
```diff
@@ -203,7 +212,7 @@ public class AnvilCompilation internal constructor(
  */
 @ExperimentalAnvilApi
 public fun compileAnvil(
-  vararg sources: String,
+  @Language("kotlin") vararg sources: String,
```
RBusarow, line 215: Cool!

I always assumed this annotation needed to be directly above the string literal, like with `//language=kotlin`.  I didn't expect that the IDE would actually be doing deep inspection.

RBusarow (approved):

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/TestUtils.kt
vRallev: Same. @ZacSweers started using this and I just added the missing annotations in this commit.

christiandeange (approved):

PR #406, opened 2021-11-02
Apply IDE suggestions that promise a better performance and fix a typo.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -97,7 +97,7 @@ internal class InterfaceMerger(
             val contributesMultibindingAnnotation = classDescriptorForReplacement
               .annotationOrNull(contributesMultibindingFqName)
 
-            // Verify that the the replaced classes use the same scope.
+            // Verify that the replaced classes use the same scope.
```
vRallev, line 100: Copy pasta

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -159,7 +159,7 @@ internal class InterfaceMerger(
     if (excludedClasses.isNotEmpty()) {
       val intersect = supertypes.getAllSuperTypes()
         .toList()
-        .intersect(excludedClasses)
+        .intersect(excludedClasses.toSet())
```
vRallev, line 162: IDE showed this here (and at other places) and this makes sense, because then the potentially same elements don't need to be checked multiple times.

christiandeange: Woah neat

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -219,11 +219,10 @@ internal class ModuleMergerIr(
     val contributedModules = modules
       .asSequence()
       .map { it.first.owner }
-      .filterIsInstance<IrClass>()
```
christiandeange, line 222: Filter call is no longer needed?

christiandeange (approved):

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
vRallev: No. I think the compiler APIs must have changed at some point.

PR #408, opened 2021-11-02
support generic supers for assisted factories when the assisted factory interface is generated

### LINKED ISSUE #407: Assisted factory generation fails if the factory interface is generated and it implements an interface from another module
This is similar to #395, but only affects generated factory interfaces.  If this code is generated, then the subsequent generation of `AssistedFactoryImpl` will fail.

```kotlin
data class AssistedService @AssistedInject constructor(
  @Assisted val string: String
)
    
@AssistedFactory
interface AssistedFactory : Function1<String, AssistedService>
```

start of the stacktrace:
```
e: java.lang.AssertionError: Classifier descriptor of a type should be of type ClassDescriptor: R
	at org.jetbrains.kotlin.resolve.DescriptorUtils.getClassDescriptorForTypeConstructor(DescriptorUtils.java:390)
	at org.jetbrains.kotlin.resolve.DescriptorUtils.getClassDescriptorForType(DescriptorUtils.java:383)
	at com.squareup.anvil.compiler.internal.CompilerUtilsKt.classDescriptorForType(CompilerUtils.kt:93)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator$AssistedFactoryFunction$Companion$toAssistedFactoryFunction$returnTypeLazy$1.invoke(AssistedFactoryGenerator.kt:293)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator$AssistedFactoryFunction$Companion$toAssistedFactoryFunction$returnTypeLazy$1.invoke(AssistedFactoryGenerator.kt:291)
	at kotlin.UnsafeLazyImpl.getValue(Lazy.kt:81)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator.generateFactoryClass(AssistedFactoryGenerator.kt:106)
	at com.squareup.anvil.compiler.codegen.dagger.AssistedFactoryGenerator.generateCodePrivate(AssistedFactoryGenerator.kt:90)
	at com.squareup.anvil.compiler.codegen.PrivateCodeGenerator.generateCode(PrivateCodeGenerator.kt:21)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted$generateCode(CodeGenerationExtension.kt:92)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted(CodeGenerationExtension.kt:113)
```

This is happening because the `fun invoke(p1: P1): R` function is invisible to PSI, and the `FunctionDescriptor` we get is from `Function1`, before the types are defined.

RBusarow:
fixes #407

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
@@ -1,0 +1,146 @@
+package com.squareup.anvil.compiler.internal
+
+import com.squareup.anvil.annotations.ExperimentalAnvilApi
+import com.squareup.anvil.compiler.api.AnvilCompilationException
+import com.squareup.anvil.compiler.internal.ClassReference.Descriptor
+import com.squareup.anvil.compiler.internal.ClassReference.Psi
+import org.jetbrains.kotlin.descriptors.ClassDescriptor
+import org.jetbrains.kotlin.descriptors.ModuleDescriptor
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.resolve.descriptorUtil.fqNameSafe
+
+/**
+ * Used to create a common type between [KtClassOrObject] class references and [ClassDescriptor]
+ * references, to streamline parsing.
+ *
+ * @see allSuperTypeClassReferences
+ * @see requireClassReference
+ */
+@ExperimentalAnvilApi
+public sealed class ClassReference {
+
+  public abstract val fqName: FqName
+
+  public class Psi internal constructor(
+    public val clazz: KtClassOrObject,
+    override val fqName: FqName
+  ) : ClassReference()
+
+  public class Descriptor internal constructor(
+    public val clazz: ClassDescriptor,
+    override val fqName: FqName
+  ) : ClassReference()
+
+  override fun toString(): String {
+    return "${this::class.qualifiedName}($fqName)"
+  }
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is ClassReference) return false
+
+    if (fqName != other.fqName) return false
+
+    return true
+  }
+
+  override fun hashCode(): Int {
+    return fqName.hashCode()
+  }
+}
+
+/**
+ * Attempts to resolve the [fqName] to a [ClassDescriptor] first, then falls back to a
+ * [KtClassOrObject] if the descriptor resolution fails. This will happen if the code being parsed
+ * was generated as part of the compilation round for this module.
+ */
+@ExperimentalAnvilApi
+public fun ModuleDescriptor.classReferenceOrNull(fqName: FqName): ClassReference? {
+  return fqName.classDescriptorOrNull(this)?.toClassReference()
+    ?: getKtClassOrObjectOrNull(fqName)?.toClassReference()
+}
+
+@ExperimentalAnvilApi
+public fun ClassDescriptor.toClassReference(): Descriptor {
+  return Descriptor(this, fqNameSafe)
+}
+
+@ExperimentalAnvilApi
+public fun KtClassOrObject.toClassReference(): Psi {
+  return Psi(this, requireFqName())
+}
+
+@ExperimentalAnvilApi
+public fun ModuleDescriptor.requireClassReference(fqName: FqName): ClassReference {
+  return classReferenceOrNull(fqName)
+    ?: throw AnvilCompilationException("Couldn't resolve ClassReference for $fqName")
+}
+
+/**
+ * This will return all super types as [ClassReference], whether they're parsed as [KtClassOrObject]
+ * or [ClassDescriptor]. This will include generated code, assuming it has already been generated.
+ * The returned sequence will be distinct by FqName, and Psi types are preferred over Descriptors.
+ *
+ * The first elements in the returned sequence represent the direct superclass to the receiver. The
+ * last elements represent the types which are furthest up-stream.
+ *
+ * @param includeSelf If true, the receiver class is the first element of the sequence
+ */
+@ExperimentalAnvilApi
+public fun ClassReference.allSuperTypeClassReferences(
+  module: ModuleDescriptor,
+  includeSelf: Boolean = false
+): Sequence<ClassReference> {
+
+  fun KtClassOrObject.superTypeEntryNames() = superTypeListEntries
+    .map { it.requireFqName(module) }
+
+  fun List<FqName>.typeRefs(): Sequence<ClassReference> {
+    return asSequence().map { module.requireClassReference(it) }
+  }
+
+  val initial = sequenceOf(this).drop(if (includeSelf) 0 else 1)
+
+  return generateSequence(initial) { superTypes ->
+    superTypes
+      .mapNotNull { classRef ->
+        when (classRef) {
+          is Psi -> {
+            classRef.clazz.superTypeEntryNames()
+              .takeIf { it.isNotEmpty() }
+              ?.typeRefs()
+          }
+          is Descriptor -> {
+            classRef.clazz.directSuperClassAndInterfaces()
+              .takeIf { it.isNotEmpty() }
+              ?.asSequence()
+              ?.map { module.requireClassReference(it.fqNameSafe) }
+          }
+        }
+      }
+      .flatten()
+      .takeIf { it.firstOrNull() != null }
+  }.flatten()
+    .distinctBy { it.fqName }
+}
+
+/**
+ * @param parameterName The name of the parameter to be found, not including any variance modifiers.
+ * @return The 0-based index of a declared generic type.
+ */
+@ExperimentalAnvilApi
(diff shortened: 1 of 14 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * @see allSuperTypeClassReferences
+ * @see requireClassReference
+ */
+@ExperimentalAnvilApi
+public sealed class ClassReference {
+
+  public abstract val fqName: FqName
+
+  public class Psi(
```
vRallev, line 25: I wonder if it's better to have the constructors private.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -0,0 +1,162 @@
...
+        clazz.directSuperClassAndInterfaces()
+          .map { Descriptor(it, it.fqNameSafe) }
+          .asSequence()
+      }
+      is Psi -> {
+        clazz.superTypeEntryNames()
+          .typeRefs()
+      }
+    }
+  }
```
vRallev, line 119: This duplication can be likely avoided if you use `sequenceOf(this).drop(if (includeSelf) 0 else 1)`. Or am I missing something?

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -0,0 +1,162 @@
...
+    .distinctBy { it.fqName }
+}
+
+/**
+ * @param parameterName The name of the parameter to be found, not including any variance modifiers.
+ * @return The 0-based index of a declared generic type.
+ */
+@ExperimentalAnvilApi
+public fun ClassReference.indexOfTypeParameter(parameterName: String): Int {
+  return when (this) {
```
vRallev, line 150: I think instead of the when here, this should be a proper function on the type. This will avoid the `when` check.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
```diff
@@ -482,6 +482,51 @@ public final class AssistedServiceFactory_Impl implements AssistedServiceFactory
     }
   }
 
+  @Test fun `the factory function may be provided by a generic super type with generic parameter`() {
+
+    // This is broken for Dagger, but working with Anvil factory generation
+    if (useDagger) return
```
vRallev, line 488: I'm unsure about this. In the past I didn't support use cases that Dagger didn't support. I found problems with deeply nested `Provider<Provider<Provider<..>>>`. At least we should file a Dagger issue and link it here.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -0,0 +1,162 @@
...
+}
+
+@ExperimentalAnvilApi
+public fun ClassDescriptor.toClassReference(): Descriptor {
+  return ClassReference.Descriptor(this, fqNameSafe)
+}
+
+@ExperimentalAnvilApi
+public fun KtClassOrObject.toClassReference(): Psi {
+  return ClassReference.Psi(this, requireFqName())
```
vRallev, line 73: The IDE show me here and above that `ClassReference.` is a redundant qualifier name and can be dropped.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -285,28 +289,64 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
+              // This can be a generic type like `T` if the function is declared in a super type.
+              // Check that it can resolve before using this type.
+              .takeIf { it.classDescriptorOrNull() != null }
+              ?.asTypeName()
+              ?: factoryClass.resolveGenericKotlinType(
+                module,
+                containingClass.toClassReference(),
+                param.type
+              )
+                ?.requireTypeName(module)
```
vRallev, line 333: I find the formatting a little weird here. I think it's easier to read if `?.requireTypeName()` is on the previous line.
```suggestion
              )?.requireTypeName(module)
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -285,28 +289,64 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
+              ?: run {
+                throw AnvilCompilationException(
+                  "Unable to resolve assisted function's argument type: ${param.type}.",
+                  element = factoryClass
+                )
+              }
```
vRallev, line 339: The `run` is redundant here.
```suggestion
              ?: throw AnvilCompilationException(
                "Unable to resolve assisted function's argument type: ${param.type}.",
                element = factoryClass
              )
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -426,10 +467,26 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
-          val type = descriptor.type.asTypeName()
+          val type = descriptor.type
+            .takeIf { it.classDescriptorOrNull() != null }
+            ?.asTypeNameOrNull()
+            ?: factoryClass.resolveGenericKotlinType(
+              module = module,
+              declaringClass = containingClass.toClassReference(),
+              typeToResolve = descriptor.type
+            )
+              ?.requireTypeName(module)
```
vRallev, line 485:
```suggestion
            )?.requireTypeName(module)
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -285,28 +289,64 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
+          // This can be a generic type like `T` if the function is declared in a super type.
+          // Check that it can resolve before using this type.
+          returnType?.classDescriptorOrNull()
+            ?.fqNameSafe
+            ?: factoryClass.resolveGenericKotlinType(
+              module,
+              containingClass.toClassReference(),
+              returnType ?: fail()
+            )
+              ?.fqNameOrNull(module)
```
vRallev, line 317:
```suggestion
            )?.fqNameOrNull(module)
```

vRallev (approved):
I was always "afraid" of starting with a class like `ClassReference` and having two implementations. In hindsight, it's long overdue and many of the utilities should be moved into this class long term.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
RBusarow: Heh...  Thanks!  This was a standalone "fail" function which I inlined via the IDE.  I didn't notice the addition of the `run`.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
RBusarow: My thinking is that if it works, then someone might be using it, so it's best to catch any regressions.  But I'm happy to remove the test if you'd prefer.

I'd intended to log an issue, but hadn't had a chance yet.  Here we are: https://github.com/google/dagger/issues/2984

I'll add a link to the comment.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
RBusarow: Sure, since there are also the `.to____` extension functions.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
vRallev: Keep the test. The link is good enough for me.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
RBusarow: I can see how this looks a little weird in isolation.  

The idea is that `ClassReference` is the unifying facade, so that the `when` checks can be avoided elsewhere.  Those pending `BindingModuleGenerator` changes have more extension functions like this.  For instance, `ClassReference.scope(...)`, which just does a `when` and then delegates to the existing functions for the two different APIs.

PR #409, opened 2021-11-03
Support contributing subcomponents

### LINKED ISSUE #160: Support delaying generating subcomponents
Imagine this module dependency tree:
```
:app -> :lib-a
    \-> :lib-b
```
`:lib-a` creates a subcomponent with `@MergeSubcomponent`. Note that Dagger won't generate any code in `:lib-a`. `:app` creates the component with `@MergeComponent`. That's where Dagger generates the code for the subcomponent. 

If `:lib-b` contributes to the scope of the subcomponent in `:lib-a`, then the contribution won't be picked up without adding the dependency `:lib-a -> :lib-b`. This can lead to confusing situations. 

Anvil should generate the code for the subcomponent in the `:app` module and then pick up all contributed bindings. This would also solve the problem that Dagger components need to be declared twice in the main source set and the source set for tests, if the test components merge different classes.

vRallev:
This resolves #160. 

I acknowledge that the size of the PR isn't ideal. But a lot of it is are coming from tests. It might be helpful to review commits separately. I thought about creating separate stacked PRs, but it didn't feel right to merge this feature half-finished. One good way to review this PR is by reading the documentation of the new `@ContributesSubcomponent` annotation and then looking at the tests whether they make sense. That would be a big help. I also suggest reading the issue #160 and my proposal in the issue first.


chrisjenx:
Are you able to throw a snapshot up so we can try some real world tests?
Read through it last night seems good, just want to try some cases to check some assumptions

vRallev:
> Are you able to throw a snapshot up so we can try some real world tests? Read through it last night seems good, just want to try some cases to check some assumptions

Absolutely: [2.3.8-CONTRIBUTE-SNAPSHOT](https://oss.sonatype.org/content/repositories/snapshots/com/squareup/anvil/annotations/2.3.8-CONTRIBUTE-SNAPSHOT/). I appreciate the feedback and testing this.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
```diff
@@ -175,6 +175,225 @@ class ContributesSubcomponentGeneratorTest {
     }
   }
 
+  @Test fun `two ore more parent component interfaces aren't allowed`() {
```
0xMatthewGroves, line 178: Spelling mistake: ore -> or

0xMatthewGroves (approved):
Approval on this one is a bit of a rubberstamp based on the size of the PR BUT tests look complete and I didn't see any glaring issues verifying logic.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
vRallev: No, typo :)

chrisjenx:
Need to do some more testing to nail down the issue, but I've only gotten kapt errors with all my testing (as soon as I change to use `ContributesSubcomponent`) probably something about our setup, I'll try an nail down a simpler example

chrisjenx:
Oh looks like might throw better errors in those latest commits, if you are able to push a newer snap, wondering if that works better.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -0,0 +1,492 @@
...
+                  addClassArrayMember("exclude", contribution.exclude)
+                }
+                .build()
+            )
+            .apply {
+              if (factoryClass != null) {
+                addType(generateFactory(factoryClass.originalDescriptor))
+              }
+            }
+            .addType(generateParentComponent(contribution, module, factoryClass))
```
kingsleyadio, line 232: Prior to this change, we had something like this in **application** code:
```kotlin
@MergeSubcomponent(FragmentScope::class)
@SingleIn(FragmentScope::class)
public interface GeneratedFragmentComponent : FragmentComponent {

    @Subcomponent.Factory
    public interface Factory : FragmentComponent.Factory

    @ContributesTo(ActivityScope::class)
    @dagger.Module(subcomponents = [GeneratedFragmentComponent::class])
    public interface Module {

        @Binds
        public fun bindComponent(target: Factory): FragmentComponent.Factory // 1
    }
}
```
which enabled us inject a `FragmentComponent.Factory` directly when provisioning a fragment. After trying out this PR, I modified the **library** code to define the relationship directly. So it became something like this
```kotlin
@ContributesSubcomponent(
    scope = FragmentScope::class,
    parentScope = ActivityScope::class
)
@SingleIn(FragmentScope::class) // 2
public interface FragmentComponent {
    public fun getFragmentMap(): FragmentMap

    @ContributesSubcomponent.Factory
    public interface Factory {
        public fun create(@BindsInstance someParam: SomeParam): FragmentComponent
    }

    @ContributesTo(ActivityScope::class)
    public interface ParentComponent {
        public fun createFragmentFactory(): Factory
    }
}
```
which generated code that looks like this:
```kotlin
@MergeSubcomponent(scope = FragmentScope::class)
public interface FragmentComponentAnvilSubcomponent : FragmentComponent {
  @Subcomponent.Factory
  public interface SubcomponentFactory : FragmentComponent.Factory

  @ContributesTo(
    ActivityScope::class,
    replaces = [FragmentComponent.ParentComponent::class]
  )
  public interface ParentComponent : FragmentComponent.ParentComponent {
    public override fun createFragmentFactory(): SubcomponentFactory
  }
}
```

A few things stand out here:
1. We had a module that was contributed into the parent scope, where we could bind the generated Factory with the defined one. This was how we were able to simply inject the defined Factory. With this PR, we couldn't do that anymore even if we wanted to, since library code doesn't have access to the generated Factory from application code. I think it will be desirable to generate this auto-wiring module here as well
2. Since ideally, we can now totally avoid writing a subcomponent, I tried adding the `SingleIn` scope annotation on the FragmentComponent, and hoped that it would also be copied over to the generated FragmentComponentAnvilSubcomponent, but that wasn't the case. I wonder if this is something that can/should happen

kingsleyadio (commented):
Hi
First off: This is amazing! Thanks a lot for working on it. It came just in time for some improvements we're trying to add to our library (to be potentially open sourced soon)
Background: the library is a thin layer on top of Anvil that aims to offer all the features from Hilt, but with much less restrictions

I took some time to play with this PR and it is really refreshing. I found a little quirk during the Dagger compilation phase that I think can be smoothened out with minimal effort

Kindly let me know your thoughts

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
vRallev: I appreciate the very detailed feedback a lot 🙏🏻 

For 1) I haven't tested it yet, but I assumed that you'd be able to inject the factory. I'll double check, thanks for pointing out!

2) Is an oversight, I'll fix that.

vRallev: 1) is addressed here https://github.com/square/anvil/pull/427 It was quite easy to solve.

vRallev: 2) is addressed here https://github.com/square/anvil/pull/428

kingsleyadio: Beautiful!
That was really quick. Thank you
Went through both PRs, and they look solid to me ✅ 

If you get a chance to publish a new snapshot once they get merged, that'd be nice. I can do another quick test and let you know how it goes

vRallev: There was actually another big issue that I had to fix: https://github.com/square/anvil/pull/430

I published a new snapshot ([`2.3.10-SNAPSHOT`](https://oss.sonatype.org/content/repositories/snapshots/com/squareup/anvil/compiler/2.3.10-SNAPSHOT/)) and tested the feature finally internally myself and our first use case can be solved with this new feature. More testing next week.

Please let me know how the snapshot works for you.

kingsleyadio: Oh. Great find! Probably would have run into that eventually. Haha
I played with the new snapshot, and it worked great

I'll give it a spin on a much larger codebase and report any findings. Really looking forward to having this feature merged in 🚀

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
```diff
@@ -0,0 +1,399 @@
...
+      assertThat(messages).contains("Source0.kt: (9, 3)")
+      assertThat(messages).contains(
+        "A factory must have exactly one abstract function returning the subcomponent " +
+          "com.squareup.test.SubcomponentInterface."
+      )
+    }
+  }
+
+  @Test
+  fun `a factory must have a single abstract method returning the subcomponent - non-abstract function`() {
```
chrisjenx, line 370: I think we're missing one or two tests here?
- Must supply `@ContributesSubcomponent.Factory`  not `@Subcomponent.Factory`
- Fail on `@Subcomponent.Builder` <- I kept on forgetting and kept on using the builder and wondered why I got dagger errors.

chrisjenx:
Awesome work with this @vRallev, I thought or two:

## First thought

I'm wondering if this is out of scope for this specific PR, but semi-related to the overall feature. This now makes it much easier to replace Modules/Bindings/Impl at the point the component is actually compiled. However in the name of keeping boilerplate down we ideally do this:

```
private val yourDayApi: YourDayApi = mock {
    on(it.shouldShowYourDay(anyOrNull(), anyOrNull(), anyOrNull())).thenReturn(Single.just(true))
    on(it.fetchYourDayMetrics(any(), anyOrNull(), anyOrNull())).thenReturn(Single.error(NotImplementedError()))
}

@Module
@ContributesTo(YourDayScope::class, replaces = [YourDayBindings::class])
inner class TestYourDayBinding {
    @Provides fun yourDayApi(): YourDayApi = yourDayApi
}
```
This way for each test run we'd get a clean module and mock each time the graph is created.

```
error: @Subcomponent.Factory method is missing parameters for required modules or subcomponents: [com.package.YourDayActivityTest.TestYourDayBinding]
```

I'm debating if this is something Dagger should fix TBH, as a No-arg Module really shouldn't need to be defined as it can be init with `newInstance()`.

Maybe I'm missing something obvious here?

## Thought 2

At this point, scopes are getting generated when the `parentScope` gets generated, in a App build scenario this is pretty straight forward use case, however for testing however we have to define `@MergeComponent(AppScope::class) TestAppComponent` over and over in every `androidTest`.

I'm still fleshing out how I think this would work, something like
`:test-utils / @ContributesComponent(AppScope::class) TestAppComponent`
`:feature/androidTest/ @GenerateComponent(AppScope::class)`

Dunno I'll think about the second one some more...

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
vRallev: Good suggestion: https://github.com/square/anvil/pull/435

vRallev:
@chrisjenx Your first point is really out of scope. Would that be different if you wouldn't use Anvil?

Number 2 won't work. There must be something triggering the generation of all components. We're in a very similar situations where we have hundreds of development applications and they all must define a component like this:
```kotlin
@SingleIn(AppScope::class)
@MergeComponent(AppScope::class)
interface DemoLoginAppComponent : DevelopmentApplicationComponent
```
And then the same component for tests:
```kotlin
@SingleIn(AppScope::class)
@MergeComponent(AppScope::class)
interface DemoLoginTestAppComponent : DevelopmentApplicationWiringTestComponent
```
Quite annoying and that's a lot of "boilerplate". But Anvil provides you all tools to solve this specific issue for your codebase. Take a look at custom code generators: https://github.com/square/anvil/tree/main/compiler-api That's likely what I'll leverage internally.

vRallev:
I tested this feature now a lot internally and it covers all of our use cases. I'm going to merge this new feature. The annotation is still experimental until I'm fully confident it works as expected.

vRallev:
Thanks for all the feedback!

PR #410, opened 2021-11-03
fall back to PSI parsing for BindingModuleGenerator

### LINKED ISSUE #310: PsiUtils' requireClassDescriptor fails to resolve generated types

RBusarow:
fixes #310

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
@@ -4,6 +4,7 @@ import com.squareup.anvil.annotations.ExperimentalAnvilApi
 import com.squareup.anvil.compiler.api.AnvilCompilationException
 import com.squareup.anvil.compiler.internal.ClassReference.Descriptor
 import com.squareup.anvil.compiler.internal.ClassReference.Psi
+import com.squareup.kotlinpoet.AnnotationSpec
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
 import org.jetbrains.kotlin.name.FqName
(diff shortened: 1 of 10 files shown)


integration-tests/code-generator/src/main/java/com/squareup/anvil/test/TestCodeGenerator.kt
```diff
@@ -66,18 +66,18 @@ class TestCodeGenerator : CodeGenerator {
           interface ContributedModule 
         """.trimIndent()
 
-        // This snippet is helpful for testing #310, which is still failing.
+        // This snippet is helpful for testing #310.
```
vRallev, line 69: 🎉

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -93,38 +94,41 @@ public fun ClassReference.allSuperTypeClassReferences(
...
-              ?.asSequence()
-              ?.map { module.requireClassReference(it.fqNameSafe) }
-          }
-        }
-      }
+      .mapNotNull { classRef -> classRef.directSuperClassReferences(module) }
       .flatten()
       .takeIf { it.firstOrNull() != null }
   }.flatten()
     .distinctBy { it.fqName }
```
vRallev, line 109: What I meant earlier was the suggestion below and always use the `sequenceOf(this)` as initial value.
```suggestion
  }
    .drop(if (includeSelf) 0 else 1)
    .flatten()
    .distinctBy { it.fqName }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -93,38 +94,41 @@ public fun ClassReference.allSuperTypeClassReferences(
...
 }
 
+@ExperimentalAnvilApi
+public fun ClassReference.directSuperClassReferences(
+  module: ModuleDescriptor
+): Sequence<ClassReference> {
+
```
vRallev, line 116: Nit: unnecessary empty line.
```suggestion
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -93,38 +94,41 @@ public fun ClassReference.allSuperTypeClassReferences(
...
+  return when (this) {
+    is Descriptor -> {
+      clazz.directSuperClassAndInterfaces()
+        .asSequence()
+        .map { module.requireClassReference(it.fqNameSafe) }
+    }
+    is Psi -> {
+      clazz.superTypeListEntries
+        .map { it.requireFqName(module) }
+        .asSequence()
```
vRallev, line 126: Move the `asSequence()` operator one line up.
```suggestion
        .asSequence()
        .map { it.requireFqName(module) }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -144,3 +148,38 @@ public fun ClassReference.indexOfTypeParameter(parameterName: String): Int {
...
+@ExperimentalAnvilApi
+public fun ClassReference.qualifiers(
+  module: ModuleDescriptor
+): List<AnnotationSpec> {
+  return when (this) {
+    is Descriptor -> {
+      clazz.annotations
+        .filter { it.isQualifier() }
+        .map { it.toAnnotationSpec(module) }
+    }
```
vRallev, line 178: The curly braces here and at many other places are redundant. We usually prefer not to have them, if they aren't required. It's not dangerous to omit them, because you use `when` as an expression. I'd be in favor of removing them everywhere where not needed. But it's not a hill I would die on.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -483,6 +487,11 @@ public fun KtClassOrObject.properties(
   includeCompanionObjects: Boolean
 ): List<KtProperty> = classBodies(includeCompanionObjects).flatMap { it.properties }
 
+@ExperimentalAnvilApi
+public fun KtClassOrObject.isObject(): Boolean {
+  return this is KtObjectDeclaration
+}
```
vRallev, line 493: Nit: expression?
```suggestion
public fun KtClassOrObject.isObject(): Boolean = this is KtObjectDeclaration
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -93,38 +94,41 @@ public fun ClassReference.allSuperTypeClassReferences(
...
-          }
-          is Descriptor -> {
-            classRef.clazz.directSuperClassAndInterfaces()
-              .takeIf { it.isNotEmpty() }
-              ?.asSequence()
-              ?.map { module.requireClassReference(it.fqNameSafe) }
-          }
-        }
-      }
+      .mapNotNull { classRef -> classRef.directSuperClassReferences(module) }
```
vRallev, line 105: This can be null.
```suggestion
      .map { classRef -> classRef.directSuperClassReferences(module) }
```

vRallev (commented):
This PR is awesome. There are mostly nits I'd address.

vRallev (approved):

PR #413, opened 2021-11-05
Support the AGP and Android projects properly when syncing sources in the IDE

vRallev:
https://user-images.githubusercontent.com/1558045/140458356-c110dcc6-7ebb-4ce3-9b6a-97effdd1b49e.mov

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -1,6 +1,10 @@
+@file:Suppress("UnstableApiUsage")
+
 package com.squareup.anvil.plugin
 
+import com.android.build.api.dsl.AndroidSourceSet
 import com.android.build.gradle.AppExtension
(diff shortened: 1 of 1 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -427,6 +436,7 @@ internal class Variant private constructor(
   val project: Project,
   val compileTaskProvider: TaskProvider<KotlinCompile>,
   val androidVariant: BaseVariant?,
+  val androidSourceSets: List<AndroidSourceSet>?,
```
christiandeange, line 439: Is null treated differently than an empty list?

christiandeange (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
vRallev: I really want to represent absence here like with the variant above. I feel better this way.

PR #418, opened 2021-11-05
Check for the correct reported function names

vRallev:
When there are multiple abstract methods in an assisted factory.

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
@@ -856,6 +856,7 @@ public final class AssistedServiceFactory_Impl<T extends CharSequence> implement
       import dagger.assisted.AssistedFactory
       import dagger.assisted.AssistedInject
       
+      @Suppress("EqualsOrHashCode")
       data class AssistedService<T> @AssistedInject constructor(
         val int: Int,
         @Assisted val stringBuilder: T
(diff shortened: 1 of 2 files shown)


ZacSweers (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
```diff
@@ -1968,4 +1969,25 @@ public final class AssistedServiceFactory_Impl implements AssistedServiceFactory
...
+   * Removes parameters of the functions in a String like
+   * ```
+   * [create([string]), get([])]
+   * ```
+   * That's necessary, because the output of the parameters is slightly different between Anvil
+   * and Dagger. Dagger also doesn't guarantee any order of functions.
+   */
+  private fun String.removeParameters(): String {
+    val start = 1 + (indexOf('[').takeIf { it >= 0 } ?: return this)
+    val end = indexOfLast { it == ']' }.takeIf { it >= 0 } ?: return this
```
christiandeange, line 1983:
```suggestion
    val end = indexOf(']').takeIf { it >= 0 } ?: return this
```

?

christiandeange (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedFactoryGeneratorTest.kt
christiandeange: Oh whoops, `indexOfLast`. Never mind.

vRallev: Notice the sample string: `[create([string]), get([])]`. It has to be the index of the last match.

PR #426, opened 2021-11-17
Fix DefinitelyNotNullTye

ZacSweers:
This is another special type that I _believe_ is used for cases where Kotlinc knows a given descriptor is not null due to some external indication like `@NonNull` or a JSR 305/jSpecify annotation. I'm unable to repro it in a test but I can repro it in our codebase (which has a setup that leads me to believe the JSR 305 case, where it trips up on a java base class from an upstream module sitting in a jsr305-dictated package).

I've at least been able to confirm that our code compiles with this fix and also found some prior art in kotlinc that suggests this is the right path to unwrapping [here](https://github.com/JetBrains/kotlin/blob/9ee0d6b60ac4f0ea0ccc5dd01146bab92fabcdf2/core/descriptors/src/org/jetbrains/kotlin/types/TypeUtils.java#L448-L458.


ZacSweers:
I... don't know why github thinks there are three commits here since one was already PR'd and merged 🤔. File diff is the same

vRallev:
> I... don't know why github thinks there are three commits here since one was already PR'd and merged 🤔. File diff is the same

Different hash? But not sure what exactly is different. Do you mind fixing that?

vRallev (approved):

PR #427, opened 2021-11-18
Generate a binding for the generated subcomponent so that the factory can be injected

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -245,21 +246,33 @@ internal class ContributesSubcomponentHandlerGenerator(
...
 
     return builder
       .addAnnotation(Subcomponent.Factory::class)
+      .addAnnotation(
+        // This will allow injecting the factory instance.
+        AnnotationSpec
+          .builder(ContributesBinding::class)
+          .addMember("scope = %T::class", contribution.parentScope.asClassName(module))
+          .addMember("boundType = %T::class", superclass)
+          .build()
```
vRallev, line 274: I know it's the equivalent of the Obama medal meme, but I love how Anvil annotations can be reused within generated code to avoid even more boilerplate and it just works.

chrisjenx: Oh thats really interesting, didn't even know you could do that!

vRallev: Actually, I should give the medal to @RBusarow, because he fixed an issue for `@ContributesBinding`.

0xMatthewGroves (approved):
Nice!

kingsleyadio (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
RBusarow: But then as the author of the frameworks I used to fix it, I think you can claim at least some of that medal as well. 😀

PR #429, opened 2021-11-19
fix IDE auto-format's handling of annotation line breaks for functions

RBusarow:
I hit my auto-format keybind all the time.  I don't even know I'm doing it...  In test files, it tends to create extra diffs like this one:

image

This is because KtLint is configured to ignore length issues when there are back-ticks.

https://github.com/square/anvil/blob/d659af039b28d09ec708d0205410282bd1dedf3d/.editorconfig#L33-L35


vRallev (changes requested):
This change needs to be contributed to https://github.com/square/java-code-styles

vRallev:
I'll close the PR for now given the reason.

PR #430, opened 2021-11-20
Avoid duplicate generated subcomponents

vRallev:
This PR solves an issue that the same subcomponent is generated, what results in very surprising results. We now make sure that each generated subcomponent is unique.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
@@ -36,7 +36,7 @@ import com.squareup.anvil.compiler.internal.scope
 import com.squareup.anvil.compiler.internal.toClassId
 import com.squareup.anvil.compiler.internal.toClassReference
 import com.squareup.anvil.compiler.mergeComponentFqName
-import com.squareup.anvil.compiler.mergeModulesFqName
+import com.squareup.anvil.compiler.mergeInterfacesFqName
 import com.squareup.anvil.compiler.mergeSubcomponentFqName
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
@@ -434,10 +434,10 @@ internal class ContributesSubcomponentHandlerGenerator(
     val generationTrigger = setOf(
       mergeComponentFqName,
       mergeSubcomponentFqName,
-      // Note that we don't include @MergeInterfaces, because we would potentially generate
+      // Note that we don't include @MergeModules, because we would potentially generate
       // components twice. @MergeInterfaces and @MergeModules are doing separately what
       // @MergeComponent is doing at once.
-      mergeModulesFqName,
+      mergeInterfacesFqName,
     )
   }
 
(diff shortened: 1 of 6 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -295,26 +305,13 @@ internal class ContributesSubcomponentHandlerGenerator(
...
-          .addMember("%T::class", contribution.parentScope.asClassName(module))
-          .apply {
-            if (parentComponentInterface != null) {
-              addMember(
-                "replaces = [%T::class]", parentComponentInterface.componentInterface
-              )
-            }
-          }
-          .build()
-      )
```
vRallev, line 310: This no longer worked, unfortunately.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1060,9 +1083,128 @@ class ContributesSubcomponentHandlerGeneratorTest {
     }
   }
 
-  private val Class<*>.anvilComponent: Class<*>
-    get() = classLoader
-      .loadClass("$COMPONENT_PACKAGE_PREFIX.${generatedClassesString()}$ANVIL_SUBCOMPONENT_SUFFIX")
+  @Test fun `subcomponent can be contributed and bindings replaced in a 2nd compilation`() {
+    // This test simulates a compilation in the main source set and test/androidTest source set,
+    // where contributed subcomponents are generated a second time for the test components. This
+    // test ensures that there are no duplicate generated classes.
```
vRallev, line 1089: Without these changes this test failed.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1060,9 +1083,128 @@ class ContributesSubcomponentHandlerGeneratorTest {
...
+      import com.squareup.anvil.annotations.MergeComponent
+
+      @ContributesSubcomponent(Unit::class, parentScope = Any::class)
+      interface SubcomponentInterface
+
+      @MergeComponent(Any::class)
+      interface ComponentInterface1
+
+      @MergeComponent(Any::class)
+      interface ComponentInterface2
```
vRallev, line 1184: This test highlights the issue as well that had to be solved. Originally, we would have generated only one subcomponent, but have two parent components. That would be fine, Dagger can handle this. But this becomes problematic if there are different bindings in the parent component (that's what the test above is verifying).

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1060,9 +1083,128 @@ class ContributesSubcomponentHandlerGeneratorTest {
...
+        import com.squareup.anvil.annotations.ContributesSubcomponent
+        import com.squareup.anvil.annotations.ContributesTo
+        import com.squareup.anvil.annotations.MergeComponent
+        import dagger.Module
+        import dagger.Provides
+
+        @ContributesTo(Any::class, replaces = [DaggerModule1::class])
+        @Module
+        object DaggerModule2 {
+          @Provides fun provideIntegerSix(): Int = 6
```
vRallev, line 1144: Notice that this Dagger modules replaces the other Dagger modules and the new generated subcomponent picks up the replaced module. That's quite common in tests.

0xMatthewGroves (approved):
Changes make sense to me. 👍

PR #443, opened 2021-12-11
fix member injection for multiple supers from a different module

### LINKED ISSUE #442: Still MemberInjectors issue for super classes in different Gradle project
@RBusarow 
I just tried latest SNAPSHOT, and we are closer!

I now see this when I try to build here:
https://github.com/JohnBuhanan/anvil-factory-generation-members-injector-issue

```
public class FeatureActivity1_MembersInjector(
  private val b1: Provider<B1>,
  private val b2: Provider<B2>,
  private val f1: Provider<F1>
) : MembersInjector<FeatureActivity1> {
  public override fun injectMembers(instance: FeatureActivity1): Unit {
    BaseActivity2_MembersInjector.injectB1(instance, b1.get())  <--- Should be BaseActivity1 I think?
    BaseActivity2_MembersInjector.injectB2(instance, b2.get())
    FeatureActivity1_MembersInjector.injectF1(instance, f1.get())
  }
...
```

RBusarow:
fixes #442

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
@@ -9,6 +9,7 @@ import org.jetbrains.kotlin.name.FqName
 
 sealed interface Parameter {
   val name: String
+  val originalName: String
   val typeName: TypeName
   val providerTypeName: ParameterizedTypeName
   val lazyTypeName: ParameterizedTypeName
@@ -35,8 +36,35 @@ sealed interface Parameter {
     }
 }
 
+/**
+ * Returns a name which is unique when compared to the [Parameter.originalName] of the
+ * [superParameters] argument.
+ *
+ * This is necessary for member-injected parameters, because a subclass may override a parameter
+ * which is member-injected in the super.  The `MembersInjector` corresponding to the subclass must
+ * have unique constructor parameters for each declaration, so their names must be unique.
+ *
+ * This mimics Dagger's method of unique naming.  If there are three parameters named "foo", the
+ * unique parameter names will be [foo, foo2, foo3].
+ */
+internal fun String.uniqueParameterName(
+  vararg superParameters: List<Parameter>
+): String {
+
+  val numDuplicates = superParameters.sumOf { list ->
+    list.count { it.originalName == this }
+  }
+
+  return if (numDuplicates == 0) {
+    this
+  } else {
+    this + (numDuplicates + 1)
+  }
+}
+
 internal data class ConstructorParameter(
   override val name: String,
+  override val originalName: String,
   override val typeName: TypeName,
   override val providerTypeName: ParameterizedTypeName,
   override val lazyTypeName: ParameterizedTypeName,
@@ -53,6 +81,7 @@ internal data class ConstructorParameter(
 
 internal data class MemberInjectParameter(
   override val name: String,
+  override val originalName: String,
   override val typeName: TypeName,
   override val providerTypeName: ParameterizedTypeName,
   override val lazyTypeName: ParameterizedTypeName,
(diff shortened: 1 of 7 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/MembersInjectorGeneratorTest.kt
```diff
@@ -1627,6 +1627,276 @@ public final class InjectClass_MembersInjector<T, U, V> implements MembersInject
...
+      """
+    ) {
+      assertThat(exitCode).isEqualTo(OK)
+    }
+
+    compile(
+      """
+      package com.squareup.test
+
+      import com.squareup.test.Base1
```
vRallev, line 1655: This import is not used. I think even `Base2` doesn't need to be imported because the files use the same package.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/MembersInjectorGeneratorTest.kt
```diff
@@ -1627,6 +1627,276 @@ public final class InjectClass_MembersInjector<T, U, V> implements MembersInject
...
+      import com.squareup.test.Base2
```
vRallev, line 1754: Same here.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
```diff
@@ -35,8 +36,31 @@ sealed interface Parameter {
...
 }
 
+/**
+ * Returns a name which is unique when compared to the [Parameter.originalName] of the
+ * [superParameters] argument.
+ *
+ * This mimics Dagger's method of unique naming.  If there are three parameters named "foo", the
+ * unique parameter names will be [foo, foo2, foo3].
+ */
+internal fun String.uniqueName(
```
vRallev, line 46: Would `uniqueParameterName` a better function name?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
```diff
@@ -35,8 +36,31 @@ sealed interface Parameter {
     }
 }
 
+/**
+ * Returns a name which is unique when compared to the [Parameter.originalName] of the
+ * [superParameters] argument.
+ *
+ * This mimics Dagger's method of unique naming.  If there are three parameters named "foo", the
+ * unique parameter names will be [foo, foo2, foo3].
```
vRallev, line 44: I struggled with the description first and couldn't understand why it's needed without looking at the tests added in this PR. First, when I read "parameter" I was thinking of constructor injection, but there can't be multiple parameters with the same name. It would help to mention that this is related to member injection and the same parameter names come from injected members in super types.

RBusarow: Great feedback, thanks!

compiler/src/test/java/com/squareup/anvil/compiler/dagger/MembersInjectorGeneratorTest.kt
RBusarow: Yep, you're right.  Nice catch!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
RBusarow: Definitely!  Especially considering that the receiver is just `String`, and it's top-level.

Thanks!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Parameter.kt
vRallev: Perfect 👍🏻

vRallev (approved):

PR #446, opened 2021-12-14
Always build Anvil with the IR backend and only use the old backend within tests.

vRallev:


.github/workflows/ci.yml
```diff
@@ -40,15 +39,14 @@ jobs:
           name: test-results-old-backend-${{ matrix.kotlin-version }}
           path: ./**/build/reports/tests/
 
-  test-ubuntu-ir:
+  test-ubuntu:
     runs-on: ubuntu-latest
     timeout-minutes: 25
-    continue-on-error: true
```
vRallev, line 46: That shouldn't have been here.

christiandeange (approved):

PR #449, opened 2021-12-15
Use different Anvil annotations for Kotlin 1.6+. 

vRallev:
This allows us to start using new language features and staying backwards compatible to Kotlin 1.5.


annotations/src/main/kotlin-1-6/com/squareup/anvil/annotations/ContributesBinding.kt
```diff
@@ -0,0 +1,138 @@
...
+ * ```
+ * @ContributesBinding(AppScope::class)
+ * object RealAuthenticator : Authenticator
+ * ```
+ * In this scenario instead of generating a `@Binds` method Anvil will generate a `@Provides`
+ * method returning [boundType].
+ */
+@Target(CLASS)
+@Retention(RUNTIME)
+public annotation class ContributesBinding(
```
vRallev, line 90: All these annotations are pure copies. So there's nothing really to review here.

christiandeange (approved):

0xMatthewGroves (approved):

JoelWilcox (approved):

annotations/src/main/kotlin-1-6/com/squareup/anvil/annotations/ContributesBinding.kt
0xMatthewGroves: I get suspicious whenever I see a comment like this 🕵️ 😅 🔍

vRallev:
Closing in favor of #476.

PR #450, opened 2021-12-15
Make `@ContributesTo` repeatable.

vRallev:
While you now can have multiple `@ContributesTo` annotations for Dagger modules and component interfaces, they all must have a different scope (contributing to the same scope twice is redundant).

This is a significant change and many of the internals of Anvil have to be updated.

There was a bug in Kotlin `1.6.0` [KT-49651](https://youtrack.jetbrains.com/issue/KT-49651) that has been fixed in `1.6.10`. I was able to remove the workarounds again. 

I apologize for the size of the PR, but there was simply no way to break this down. In fact, I had separate commits initially, but decided to squash them, because several tests were broken with the intermediate commits. I suggest looking for weird Kotlin idioms and looking at the tests. 

I don't plan to merge this PR soon. Probably next year when I'll try to make the other annotations repeatable as well.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
@@ -78,10 +78,11 @@ public fun KtAnnotated.hasAnnotation(
   fqName: FqName,
   module: ModuleDescriptor
 ): Boolean {
-  return findAnnotation(fqName, module) != null
+  return findAnnotations(fqName, module).isNotEmpty()
 }
 
 @ExperimentalAnvilApi
+@Deprecated("Repeatable")
 public fun KtAnnotated.requireAnnotation(
   fqName: FqName,
   module: ModuleDescriptor
@@ -93,40 +94,63 @@ public fun KtAnnotated.requireAnnotation(
 }
 
 @ExperimentalAnvilApi
+public fun KtAnnotated.requireAnnotations(
+  fqName: FqName,
+  module: ModuleDescriptor
+): List<KtAnnotationEntry> {
+  return findAnnotations(fqName, module)
+    .takeIfNotEmpty()
+    ?: throw AnvilCompilationException(
+      element = this,
+      message = "Couldn't find the annotation $fqName."
+    )
+}
+
+@ExperimentalAnvilApi
+@Deprecated("Repeatable")
 public fun KtAnnotated.findAnnotation(
   fqName: FqName,
   module: ModuleDescriptor
-): KtAnnotationEntry? {
-  val annotationEntries = annotationEntries
-  if (annotationEntries.isEmpty()) return null
+): KtAnnotationEntry? = findAnnotations(fqName, module).singleOrEmpty()
+
+@ExperimentalAnvilApi
+public fun KtAnnotated.findAnnotations(
+  fqName: FqName,
+  module: ModuleDescriptor
+): List<KtAnnotationEntry> {
+  val annotationEntries = annotationEntries.takeIfNotEmpty()
+    ?: return emptyList()
 
   // Look first if it's a Kotlin annotation. These annotations are usually not imported and the
-  // remaining checks would fail.
+  // remaining checks would fail. Return the result eagerly and assume that Kotlin annotations
+  // aren't repeatable.
   if (fqName in kotlinAnnotations) {
-    annotationEntries.firstOrNull { annotation ->
-      val text = annotation.text
-      text.startsWith("@${fqName.shortName()}") || text.startsWith("@$fqName")
-    }?.let { return it }
+    annotationEntries
+      .filter { annotation ->
+        val text = annotation.text
+        text.startsWith("@${fqName.shortName()}") || text.startsWith("@$fqName")
+      }
+      .takeIf { it.isNotEmpty() }
+      ?.let { return it }
   }
 
   // Check if the fully qualified name is used, e.g. `@dagger.Module`.
-  val annotationEntry = annotationEntries.firstOrNull {
-    it.text.startsWith("@${fqName.asString()}")
-  }
-  if (annotationEntry != null) return annotationEntry
+  val fullyQualifiedAnnotations = annotationEntries
+    .filter {
+      it.text.startsWith("@${fqName.asString()}")
+    }
 
   // Check if the simple name is used, e.g. `@Module`.
-  val annotationEntryShort = annotationEntries
-    .firstOrNull {
-      it.shortName == fqName.shortName()
-    }
-    ?: return null
+  val shortNameAnnotations = annotationEntries
+    .filter { it.shortName == fqName.shortName() && it !in fullyQualifiedAnnotations }
+    .takeIf { it.isNotEmpty() }
+    ?: return fullyQualifiedAnnotations
 
   val importPaths = containingKtFile.importDirectives.mapNotNull { it.importPath }
 
   // If the simple name is used, check that the annotation is imported.
   val hasImport = importPaths.any { it.fqName == fqName }
-  if (hasImport) return annotationEntryShort
+  if (hasImport) return shortNameAnnotations + fullyQualifiedAnnotations
 
   // Look for star imports and make a guess.
   val hasStarImport = importPaths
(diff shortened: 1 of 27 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -91,62 +92,85 @@ public fun KtAnnotated.requireAnnotation(
 }
 
 @ExperimentalAnvilApi
+public fun KtAnnotated.requireAnnotations(
+  fqName: FqName,
+  module: ModuleDescriptor
+): List<KtAnnotationEntry> {
+  return findAnnotations(fqName, module)
+    .ifEmpty { null }
```
0xMatthewGroves, line 100: takeIfNotEmpty()? You just created the extension.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -91,62 +92,85 @@ public fun KtAnnotated.requireAnnotation(
...
   module: ModuleDescriptor
-): KtAnnotationEntry? {
+): KtAnnotationEntry? = findAnnotations(fqName, module).singleOrEmpty()
+
+@ExperimentalAnvilApi
+public fun KtAnnotated.findAnnotations(
+  fqName: FqName,
+  module: ModuleDescriptor
+): List<KtAnnotationEntry> {
   val annotationEntries = annotationEntries
```
0xMatthewGroves, line 119: Doesn't look like it.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -91,62 +92,85 @@ public fun KtAnnotated.requireAnnotation(
...
-  if (annotationEntries.isEmpty()) return null
+  if (annotationEntries.isEmpty()) return emptyList()
```
0xMatthewGroves, line 120: Similar comment, but the logic below will return an emptyList() automatically if it's empty, so why do the initial check first? It's my understanding that the collection extensions have early bail out mechanisms baked in. i.e. filter {}

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -159,16 +183,57 @@ public fun KtClassOrObject.scope(
...
 
 @ExperimentalAnvilApi
+public fun KtClassOrObject.scopes(
+  annotationFqName: FqName,
+  module: ModuleDescriptor
+): List<FqName> {
+  return requireAnnotations(annotationFqName, module)
+    .map { it.scope(module) }
+    .also { scopes ->
+      // Exit early to avoid allocating additional collections.
```
0xMatthewGroves, line 216: 👍

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -159,16 +183,57 @@ public fun KtClassOrObject.scope(
...
+            "${duplicates.keys.joinToString()}. Contributing multiple times to the same scope " +
+            "is forbidden and all scopes must be distinct.",
+          element = this
+        )
+      }
+    }
+}
+
+@ExperimentalAnvilApi
+// TODO Repeatable: deprecate and add other function.
```
0xMatthewGroves, line 236: 🤔

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/TestUtils.kt
```diff
@@ -46,6 +47,9 @@ public class AnvilCompilation internal constructor(
     enableExperimentalAnvilApis: Boolean = true,
   ) = apply {
     checkNotCompiled()
+    check(!anvilConfigured) { "Anvil should not be configured twice." }
```
0xMatthewGroves, line 50: Is this related to the @ContributesTo changes or should this be a separate commit/PR? Seems unrelated to me...

compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt
```diff
@@ -1,98 +1,177 @@
...
         // Check that the annotation really is present. It should always be the case, but it's
         // a safetynet in case the generated properties are out of sync.
-        it.annotationOrNull(annotation, scope) != null
+        it.annotations(annotation, scope).isNotEmpty()
+      }
+  }
+
+  /**
+   * Returns a sequence of contributed hints from the dependency graph. Note that the result
+   * includes inner classes already.
```
0xMatthewGroves, line 54: Mega nit: Note that the result already includes inner classes.

compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt
```diff
@@ -1,98 +1,177 @@
...
+            null
+          }
+        }
+        else -> null
+      }
+    }
   }
+}
+
+// TODO Repeatable: Cleanup this class with what's actually required.
```
0xMatthewGroves, line 130: Not sure if you use github to track your issues, but worthwhile to note your ticket number here if you do.

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -36,45 +38,58 @@ internal class InterfaceMerger(
...
     if (!DescriptorUtils.isInterface(thisDescriptor)) {
       throw AnvilCompilationException(thisDescriptor, "Dagger components must be interfaces.")
     }
 
+    val module = thisDescriptor.module
+
+    val scope = try {
+      mergeAnnotation.scope(module)
+    } catch (e: AssertionError) {
+      // In some scenarios this error is thrown. Throw a new exception with a better explanation.
```
0xMatthewGroves, line 66: 👍

0xMatthewGroves (approved):
Whew. Big PR 😅

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
vRallev: For that I had to change the signature slightly. It's working now.
```kotlin
public inline fun <C, T> C.takeIfNotEmpty(): C? where C : Collection<T> = ifEmpty { null }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
vRallev: It's an improvement, because I don't want to call the getter function multiple times. That's unnecessary overhead, because it might create a new collection every time.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
vRallev: The operations below are quite heavy and returning early here happens quite often. It's a quick, faster exit.

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/TestUtils.kt
vRallev: It was necessary for one specific test that verifies contributions from the old format are picked up. This is just a sanity check.

compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt
vRallev: Done

compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt
vRallev: I have to clean up all deprecated methods and TODOs later. This PR would otherwise explode even further.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -618,4 +618,226 @@ class InterfaceMergerTest(
...
+      """
+    ) {
+      assertThat(componentInterface extends contributingInterface).isFalse()
+      assertThat(componentInterface extends secondContributingInterface).isTrue()
+      assertThat(subcomponentInterface extends contributingInterface).isTrue()
+      assertThat(subcomponentInterface extends secondContributingInterface).isFalse()
+    }
+  }
+
+  @Test fun `contributed interfaces in the old format are picked up`() {
```
JoelWilcox, line 800: What does the old format mean here?

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -618,4 +618,226 @@ class InterfaceMergerTest(
...
+    val result = AnvilCompilation()
+      .useIR(USE_IR)
+      .enableAnvil(false)
```
JoelWilcox, line 803: Also curious if maybe `enableAnvil` should be called something else; at least, when I read this test and see that we are disabling Anvil for AnvilCompilation, but the test is for contributed interfaces being picked up, I would expect nothing to be picked up since Anvil was disabled. So perhaps this is a misunderstanding on my part

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
```diff
@@ -1119,6 +1110,244 @@ class ModuleMergerTest(
...
+    ) {
+      assertThat(componentInterface.anyDaggerComponent.modules.withoutAnvilModule())
+        .containsExactly(daggerModule2.kotlin)
+
+      assertThat(subcomponentInterface.anyDaggerComponent.modules.withoutAnvilModule())
+        .containsExactly(daggerModule1.kotlin)
+    }
+  }
+
+  @Test fun `contributed modules in the old format are picked up`() {
```
JoelWilcox, line 1305: The ample test coverage in this PR is awesome 🔥

JoelWilcox (commented):
Overall looks great! A number of my comments are related to me not having dug into the anvil implementation previously and just looking to better understand things :)

vRallev:
Closing this PR in favor of #551.

Issue #445, opened 2021-12-13
Couldn't resolve ClassReference for kotlin.IllegalStateException File is unknown


JohnBuhanan:
I'm seeing this right now after latest Snapshot.

```
e: com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve ClassReference for kotlin.IllegalStateException
File is unknown
```

Struggling to understand it and reproduce it in my sample project.

Will update this issue when I get it.

JohnBuhanan:
```
com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve ClassReference for kotlin.RuntimeException
File is unknown

	at com.squareup.anvil.compiler.internal.ClassReferenceKt.requireClassReference(ClassReference.kt:86)
	at com.squareup.anvil.compiler.internal.ClassReferenceKt$directSuperClassReferences$3.invoke(ClassReference.kt:129)
	at com.squareup.anvil.compiler.internal.ClassReferenceKt$directSuperClassReferences$3.invoke(ClassReference.kt:129)
	at kotlin.sequences.TransformingSequence$iterator$1.next(Sequences.kt:210)
	at kotlin.sequences.FlatteningSequence$iterator$1.next(Sequences.kt:299)
	at kotlin.sequences.SequencesKt___SequencesKt.firstOrNull(_Sequences.kt:169)
	at com.squareup.anvil.compiler.internal.ClassReferenceKt$allSuperTypeClassReferences$1.invoke(ClassReference.kt:108)
	at com.squareup.anvil.compiler.internal.ClassReferenceKt$allSuperTypeClassReferences$1.invoke(ClassReference.kt:104)
	at kotlin.sequences.GeneratorSequence$iterator$1.calcNext(Sequences.kt:591)
	at kotlin.sequences.GeneratorSequence$iterator$1.hasNext(Sequences.kt:609)
	at kotlin.sequences.FlatteningSequence$iterator$1.ensureItemIterator(Sequences.kt:311)
	at kotlin.sequences.FlatteningSequence$iterator$1.hasNext(Sequences.kt:303)
	at kotlin.sequences.DistinctIterator.computeNext(Sequences.kt:570)
	at kotlin.collections.AbstractIterator.tryToComputeNext(AbstractIterator.kt:42)
	at kotlin.collections.AbstractIterator.hasNext(AbstractIterator.kt:29)
[...]

PR #451, opened 2021-12-15
fix ClassDescriptor resolution for typealias

RBusarow:
partial fix of #445


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -768,6 +769,8 @@ public fun FqName.requireClassDescriptor(module: ModuleDescriptor): ClassDescrip
 @ExperimentalAnvilApi
 public fun FqName.classDescriptorOrNull(module: ModuleDescriptor): ClassDescriptor? {
   return module.resolveClassByFqName(this, FROM_BACKEND)
+    // returns the original reference being pointed to by a typealias
```
vRallev, line 772: That's okay for now, but we prefer full sentences as comments starting with an uppercase letter and ending with punctuation mark.

vRallev (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
RBusarow: Thanks for the explanation.  I'd noticed that the project leaned that way, but didn't realize that it'd been formalized.

PR #452, opened 2021-12-15
fix PSI resolution of local references to nested types

RBusarow:
partial fix of #445


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -444,13 +445,36 @@ public fun PsiElement.requireFqName(
...
+ *
+ * In practice, the returned sequence represents all declarations which can be referenced by the
+ * receiver without needing an import statement or qualified name.
+ */
+@ExperimentalAnvilApi
+public fun PsiElement.declarationsWithinScope(): Sequence<KtNamedDeclaration> {
+  return parents.filterIsInstance<KtNamedDeclaration>()
+    .flatMap { it.children.filterIsInstance<KtNamedDeclaration>() + it }
+    .distinct()
+}
```
vRallev, line 476: Nice!

vRallev (approved):

PR #454, opened 2021-12-15
avoid parsing full class hierarchies for injected members when possible

### LINKED ISSUE #453: Unnecessary MembersInjectorGenerator parsing/generation when the target class doesn't have its own injected properties
Right now, `MembersInjectGenerator` inspects the full hierarchy for every`KtClassOrObject`.  If it finds any member-injected properties anywhere in that hierarchy, then it generates a `MembersInjector<T>` for the target class.  This is doing more work than it needs to.  

If a project has an abstract `BaseFragment` or something with 100 different implementations, and none of them have any member injection, the current implementation will parse `BaseFragment` 101 times looking for properties.  

If that `BaseFragment` *does* have a member-injected property but none of the subclasses do, the current implementation will generate 100 injectors (1 for each subclass).  Dagger won't know they exist and will never use them.

Instead, the generator can start by just inspecting the target class without supers.  If the target has any *directly* declared injected fields, *then* the full hierarchy needs to be parsed and a new `MembersInjector<T>` should be generated.

RBusarow:
fixes #453

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
@@ -7,6 +7,7 @@ import com.squareup.anvil.compiler.internal.ClassReference.Psi
 import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.ClassName
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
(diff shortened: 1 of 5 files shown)


vRallev (approved):

vRallev:
@RBusarow There was a conflict with your other PRs.

RBusarow:
> @RBusarow There was a conflict with your other PRs.

Yeah...  I thought that might happen...

Fixed!  Thanks!

PR #455, opened 2021-12-16
Use Gradle 7.3.2. Log4shell mitigation.

autonomousapps:
Gradle 7.3.2 adds dependency constraints to the _build_ classpath to reject known-bad versions of log4j.

See also https://blog.gradle.org/log4j-vulnerability.


vRallev (changes requested):
There are actually two gradle wrappers in this project. I'll follow up.

vRallev:
Follow up: #457

PR #469, opened 2022-01-27
Add a function to get the `replaces` field for a `ClassReference` and use this new utility everywhere

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -81,6 +83,15 @@ public fun KtClassOrObject.toClassReference(): Psi {
   return Psi(this, toClassId())
 }
 
+@ExperimentalAnvilApi
+public fun FqName.toClassReference(module: ModuleDescriptor): ClassReference? {
```
RBusarow, line 87: This is very similar to the `ModuleDescriptor.classReferenceOrNull()` function above.  The key difference would seem to be that one gives priority to `ClassDescriptor` and the other `KtClassOrObject`.

Would it be a good idea to delegate one of these functions to the other?

RBusarow (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
vRallev: It spotted that literally this moment. I find the extension on `ModuleDescriptor` much harder to find (that's why I created the new extension function). I'll merge them.

vRallev: I've created a separate commit for this. I think this change makes a lot of sense.

RBusarow: > I find the extension on `ModuleDescriptor` much harder to find

Yeah, I'd say this is a pretty good indicator that it wasn't very discoverable. 😁   This is a good change. 👍

Issue #462, opened 2022-01-19
[compiler-api] Compilation Error in generated class when using custom code generator


MohiteshAeron:
I wrote a custom code generator using the `compiler-api` to generate subcomponents for Activities with ContributeSubComponent annotation, but I keep getting the following compilation error

```
e: com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: Couldn't resolve class for com.sample.anvil.FailingClass_AnvilSubComponent
File is unknown

	at com.squareup.anvil.compiler.internal.PsiUtilsKt.requireClassDescriptor(PsiUtils.kt:765)
	at com.squareup.anvil.compiler.codegen.ContributesSubcomponentHandlerGenerator.findFactoryClass(ContributesSubcomponentHandlerGenerator.kt:387)
	at com.squareup.anvil.compiler.codegen.ContributesSubcomponentHandlerGenerator.generateCode(ContributesSubcomponentHandlerGenerator.kt:206)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted$generateCode(CodeGenerationExtension.kt:93)
	at com.squareup.anvil.compiler.codegen.CodeGenerationExtension.analysisCompleted(CodeGenerationExtension.kt:107)
```

I noticed that interface is being generated properly with all the annotations and methods in place. Also, if I add the class manually then it works perfectly and compilation doesn't fail. 

I am attaching a sample project with the above use case. Thanks in advance :)
[FailingAnvil.zip](https://github.com/square/anvil/files/7894865/FailingAnvil.zip)

vRallev:
Thanks for the sample. Looks like bug to me. I'll take a look.

vRallev:
This issue was "accidentally" fixed by this commit: https://github.com/square/anvil/commit/e6d391f7a8c655d2e0ac2903c2f3de661f9f76a2 

I noticed this bug while working on something else and it had the side-effect of fixing this issue, too. The fix will be part of the next release.

vRallev:
I need to reopen the issue. While this commit fixes this particular issue, it was wrong and introduced a regression. A proper fix is in the works.

PR #470, opened 2022-01-27
Support custom code generators through the `AnvilCompilation` API in unit tests.

vRallev:
I suggest only reviewing the 2nd commit. The first commit is only moving code into a separate file.

This is pretty helpful for #462. Sample usage will look like that:
```kotlin
  @Test fun `an @ContributeSubcomponent class can be generated`() {
    val codeGenerator = object : CodeGenerator {
      override fun isApplicable(context: AnvilContext): Boolean = true

      override fun generateCode(
        codeGenDir: File,
        module: ModuleDescriptor,
        projectFiles: Collection<KtFile>
      ): Collection<GeneratedFile> {
        return ...
      }
    }

    compile(
      """
        code
      """,
      codeGenerators = listOf(codeGenerator)
    ) {
       ...
    }
  }
```


compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
```diff
@@ -36,11 +37,14 @@ public class AnvilCompilation internal constructor(
...
     disableComponentMerging: Boolean = false,
     enableExperimentalAnvilApis: Boolean = true,
+    codeGenerators: List<CodeGenerator> = emptyList(),
   ) = apply {
     checkNotCompiled()
     anvilConfigured = true
     kotlinCompilation.apply {
-      compilerPlugins = listOf(AnvilComponentRegistrar())
+      compilerPlugins = listOf(
+        AnvilComponentRegistrar().also { it.addCodeGeneratorsManually(codeGenerators) }
```
0xMatthewGroves, line 46: Super nit, but it feels implied that the code generators are included manually, so I would drop the "manually" nomenclature. Just keep it as `addCodeGenerators(codeGens)`.

0xMatthewGroves (approved):

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
vRallev: Done

RBusarow:
This is great!

PR #471, opened 2022-01-27
Extend `ClassReference` and introduce more common APIs …

### LINKED ISSUE #462: [compiler-api] Compilation Error in generated class when using custom code generator

vRallev:
…for PSI and descriptor compiler APIs. This is required to make parsing and generating code easier.

Use this new functionality in contributed subcomponent generator and avoid hardcoding descriptor APIs.

Fixes #462


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -350,41 +349,40 @@ internal class ContributesSubcomponentHandlerGenerator(
...
-    val contributionDescriptor = contributionFqName.requireClassDescriptor(module)
 
-    val contributedInnerComponentInterfaces = contributionDescriptor.unsubstitutedMemberScope
-      .getContributedDescriptors(kindFilter = CLASSIFIERS)
-      .filterIsInstance<ClassDescriptor>()
-      .filter { DescriptorUtils.isInterface(it) }
-      .filter {
-        val annotation = it.annotationOrNull(contributesToFqName, contribution.parentScope)
-        annotation != null
+    val contributedInnerComponentInterfaces = contribution.classReference
```
vRallev, line 353: Note that we used descriptor APIs before, and this was failing when a subcomponent was generated through a code generator. With the abstract API this isn't an issue anymore.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -186,7 +186,7 @@ internal class ContributesSubcomponentHandlerGenerator(
 
     val newContributions = contributions
       .flatMap { contribution ->
-        newTriggers
+        triggers
```
vRallev, line 189: I actually made this change to fix the referenced bug, but it introduced regressions and wasn't the proper fix.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1547,6 +1558,76 @@ class ContributesSubcomponentHandlerGeneratorTest {
     }
   }
 
+  @Test fun `an @ContributeSubcomponent class can be generated`() {
+    val codeGenerator = object : CodeGenerator {
```
vRallev, line 1562: Custom code generator in unit tests!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/AnnotationReference.kt
```diff
@@ -0,0 +1,70 @@
...
+
+@ExperimentalAnvilApi
+public fun AnnotationDescriptor.toAnnotationReference(): Descriptor {
+  return Descriptor(this, requireClass().toClassReference())
+}
+
+@ExperimentalAnvilApi
+public fun AnnotationReference.scope(
+  module: ModuleDescriptor,
+  parameterIndex: Int = 0
```
0xMatthewGroves, line 60: In what scenario would you want to directly call this method with an overridden parameterIndex? This seems a bit clunky, because the caller doesn't know that the parameter does nothing if it's being called on a Descriptor object.

0xMatthewGroves: Splitting them to two separate extensions might be a bit cleaner in this case. That way the Psi specific extension can have extra kdoc explaining what that's for.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -225,3 +248,48 @@ public fun ClassReference.isInterface(): Boolean = when (this) {
...
+    is Psi -> clazz.annotationEntries.map { it.toAnnotationReference(module) }
+    is Descriptor -> clazz.annotations.map { it.toAnnotationReference() }
+  }
+
+@ExperimentalAnvilApi
+public fun ClassReference.functions(): List<FunctionReference> =
+  when (this) {
+    is Psi ->
+      clazz
+        .functions(includeCompanionObjects = false)
```
0xMatthewGroves, line 270: Wow, cool.

0xMatthewGroves (approved):

RBusarow (approved):

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
RBusarow: Between this (which forces Psi resolution) and passing in the previous compilation result (which forces Descriptor resolution), there's nothing we can't _easily_ test.  I love it!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/AnnotationReference.kt
vRallev: Well, that's the whole point that there aren't supposed to be two functions. You don't need to specify the parameter with the default, but I wanted to leave the door open for external code generators and custom annotations. For example, when we read the `replaces` attribute, then the index will be different depending on the annotation.

PR #472, opened 2022-01-27
Allow replacing subcomponents

### LINKED ISSUE #466: Allow replacing `@ContributesSubcomponent`
I didn't implement this feature, because it seemed unnecessary. But there's actually at least one reasonable use case: contributing a new subcomponent with different scope annotations, e.g.
```kotlin
@ContributesSubcompnent(scope = Unit::class, parentScope = Any::class)
interface Subcomponent

@ContributesSubcompnent(scope = Unit::class, parentScope = Any::class, replaces = [Subcomponent::class])
@Singleton
interface OtherSubcomponent
```

vRallev:
Fixes #466

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
@@ -37,6 +37,8 @@ import com.squareup.anvil.compiler.internal.isAbstractClass
 import com.squareup.anvil.compiler.internal.isInterface
 import com.squareup.anvil.compiler.internal.isPublic
 import com.squareup.anvil.compiler.internal.parentScope
+import com.squareup.anvil.compiler.internal.replaces
+import com.squareup.anvil.compiler.internal.requireFqName
 import com.squareup.anvil.compiler.internal.returnType
 import com.squareup.anvil.compiler.internal.safePackageString
 import com.squareup.anvil.compiler.internal.scope
@@ -89,7 +91,10 @@ internal class ContributesSubcomponentHandlerGenerator(
 ) : CodeGenerator {
 
   private val triggers = mutableListOf<Trigger>()
-  private val contributions = mutableListOf<Contribution>()
+  private val contributions = mutableSetOf<Contribution>()
+  private val replacedContributions = mutableSetOf<ClassReference>()
+
+  private var isFirstRound = true
 
   override fun isApplicable(context: AnvilContext): Boolean =
     throw NotImplementedError(
@@ -101,9 +106,51 @@ internal class ContributesSubcomponentHandlerGenerator(
     module: ModuleDescriptor,
     projectFiles: Collection<KtFile>
   ): Collection<GeneratedFile> {
+    if (isFirstRound) {
+      isFirstRound = false
+
+      // Find all contributed subcomponents from precompiled dependencies and generate the
+      // necessary code eventually if there's a trigger.
+      contributions += classScanner
+        .findContributedClasses(
+          module = module,
+          packageName = HINT_SUBCOMPONENTS_PACKAGE_PREFIX,
+          annotation = contributesSubcomponentFqName,
+          // Don't use trigger.scope, because it refers to the parent scope of the
+          // @ContributesSubcomponent annotation.
+          scope = null
+        )
+        .map { descriptor ->
+          val annotation = descriptor.annotation(contributesSubcomponentFqName)
+
+          val modules = (annotation.getAnnotationValue("modules") as? ArrayValue)
+            ?.value
+            ?.map { it.argumentType(module).fqName() }
+            ?: emptyList()
+
+          val exclude = (annotation.getAnnotationValue("exclude") as? ArrayValue)
+            ?.value
+            ?.map { it.argumentType(module).fqName() }
+            ?: emptyList()
+
+          Contribution(
+            classReference = descriptor.toClassReference(),
+            scope = annotation.scope(module).fqNameSafe,
+            parentScope = annotation.parentScope(module).fqNameSafe,
+            modules = modules,
+            exclude = exclude
+          )
+        }
+
+      // Find all replaced subcomponents from precompiled dependencies.
+      replacedContributions += contributions
+        .flatMap {
+          it.classReference.replaces(module, contributesSubcomponentFqName)
+        }
+    }
 
     // Find new @MergeComponent (and similar triggers) that would cause generating new code.
-    val newTriggers = projectFiles
+    triggers += projectFiles
       .classesAndInnerClass(module)
       .mapNotNull { clazz ->
         val annotation = generationTrigger.firstOrNull { trigger ->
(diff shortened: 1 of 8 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -101,9 +106,51 @@ internal class ContributesSubcomponentHandlerGenerator(
     module: ModuleDescriptor,
     projectFiles: Collection<KtFile>
   ): Collection<GeneratedFile> {
+    if (isFirstRound) {
```
vRallev, line 109: It was unnecessary to do this work over and over again.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1628,6 +1628,223 @@ class ContributesSubcomponentHandlerGeneratorTest {
...
+          .toList()
+      ).containsExactly(subcomponentInterface2)
+    }
+  }
+
+  @Test fun `a contributed subcomponent can be replaced by a generated subcomponent`() {
+    // This test verifies an edge case where we generate the code for a contributed subcomponent,
+    // but then later in a new code generation round another code generator generates code that
+    // is supposed to replace the already generated subcomponent. We can't revert the code and
+    // don't want to support that use case.
```
vRallev, line 1780: Kinda proud that I was thinking of this 😅

vRallev:
Tested it internally and it works 🎉

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -101,9 +105,51 @@ internal class ContributesSubcomponentHandlerGenerator(
...
```
JoelWilcox, line 109: Nit: This `generateCode` method seems to have gotten quite large and it makes what's happening a bit harder to follow. It would be helpful to break some of its logic out to new helper methods. e.g.
```
if (isFirstRound) {
  isFirstRound = false
  populateInitialContributions(module)
  populateReplacedContributions(module)
}
```

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
```diff
@@ -463,6 +464,72 @@ class ContributesSubcomponentGeneratorTest {
...
+        replaces = [SubcomponentInterface1::class]
+      )
+      interface SubcomponentInterface2
+      """
+    ) {
+      assertThat(subcomponentInterface1.hintSubcomponent?.java).isEqualTo(subcomponentInterface1)
+      assertThat(subcomponentInterface1.hintSubcomponentParentScope).isEqualTo(Unit::class)
+
+      assertThat(subcomponentInterface2.hintSubcomponent?.java).isEqualTo(subcomponentInterface2)
+      assertThat(subcomponentInterface2.hintSubcomponentParentScope).isEqualTo(Int::class)
```
JoelWilcox, line 492: At first glance, this test makes it look like both subcomponents are contributed successfully, rather than the first being replaced. I was expecting to see some sort of `assertFalse` style assert confirming that `SubcomponentInterface1` was not contributed. Is the reasoning that these second two asserts would fail if `SubcomponentInterface1` hadn't been replaced?

edit: I see that the test I was expecting is included in ContributesSubcomponentHandlerGeneratorTest.kt. It might be helpful to name this one as something like `a contributed subcomponent can be configured to replace another subcomponent`, if the intention is to make sure including `replaces = []` doesn't break anything.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1628,6 +1628,151 @@ class ContributesSubcomponentHandlerGeneratorTest {
...
+        interface SubcomponentInterface2
+        
+        @MergeComponent(Unit::class)
+        interface ComponentInterface
+      """
+    ) {
+      val annotation = subcomponentInterface2.anvilComponent(componentInterface)
+        .getAnnotation(MergeSubcomponent::class.java)
+
+      assertThat(annotation.exclude).isEmpty()
```
JoelWilcox, line 1715: I could use a little clarification on this test as well. The test name makes me think that `SubcomponentInterface1` cannot exclude Dagger modules, component interfaces and bindings (and doing so would result in a failure), but the assert appears to be making sure that `SubcomponentInterface2` doesn't exclude anything. Is this basically saying that all excluded items from the replaced subcomponent will be lost?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
```diff
@@ -1773,6 +1773,78 @@ class ContributesSubcomponentHandlerGeneratorTest {
     }
   }
 
+  @Test fun `a contributed subcomponent can be replaced by a generated subcomponent`() {
```
JoelWilcox, line 1776: This name sounds more like the use-case is supported to me. What about `a previously generated contributed subcomponent cannot be replaced in a later round of generations`? Or at least `can` -> `can't`

JoelWilcox (commented):
Overall lgtm! Mostly interested in some test clarifications

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -90,6 +91,9 @@ internal class ContributesSubcomponentHandlerGenerator(
 
   private val triggers = mutableListOf<Trigger>()
   private val contributions = mutableListOf<Contribution>()
+  private val replacedContributions = mutableSetOf<ClassReference>()
```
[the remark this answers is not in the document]
vRallev: Agree, I changed it to `replacedReferences`. That's good enough.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -101,9 +105,51 @@ internal class ContributesSubcomponentHandlerGenerator(
...
+      isFirstRound = false
+
+      // Find all contributed subcomponents from precompiled dependencies and generate the
+      // necessary code eventually if there's a trigger.
+      contributions += classScanner
+        .findContributedClasses(
+          module = module,
+          packageName = HINT_SUBCOMPONENTS_PACKAGE_PREFIX,
+          annotation = contributesSubcomponentFqName,
+          // Don't use trigger.scope, because it refers to the parent scope of the
```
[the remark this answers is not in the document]
vRallev: Good catch

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
vRallev: Good idea, done.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentGeneratorTest.kt
vRallev: Very good suggestion!

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
vRallev: I changed the method name to `a replaced subcomponent's exclusions for Dagger modules, component interfaces and bindings are ignored`.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGeneratorTest.kt
vRallev: I forgot to update the name. I did TDD and initially planned to implement the functionality, but it was quite tricky and I stopped. Since this is such an edge case I added an explicit check to ensure it isn't supported and then forgot to update the method name.

JoelWilcox (approved):
Looks great!

PR #476, opened 2022-01-28
Upgrade the repository to Kotlin 1.6.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/ClassReference.kt
```diff
@@ -278,6 +278,7 @@ public fun ClassReference.functions(): List<FunctionReference> =
   }
 
 @ExperimentalAnvilApi
+@Suppress("FunctionName")
```
vRallev, line 281: Oh, I didn't mean to commit this. But it's not wrong either, so I'll keep it.

0xMatthewGroves (approved):

RBusarow (approved):

PR #485, opened 2022-02-04
Move `AnvilModuleDescriptor` into a new, more specific package.

vRallev:
Combine some of the utility functions. Always make sure that the receiver of extension function is the type the utility is made for and not the module descriptor. This improves discoverability.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt
@@ -1,11 +1,11 @@
-package com.squareup.anvil.compiler.internal
+package com.squareup.anvil.compiler.internal.reference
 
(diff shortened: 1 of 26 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt
```diff
@@ -62,7 +44,7 @@ public fun KtFile.classesAndInnerClasses(
 }
 
 @ExperimentalAnvilApi
-public fun Collection<KtFile>.classesAndInnerClass(
+public fun Collection<KtFile>.classesAndInnerClasses(
```
RBusarow, line 47: 😄

vRallev: 👋🏻

RBusarow (approved):

PR #493, opened 2022-02-08
Introduce a flag to skip several tests in the :compiler module to improve local development.

vRallev:
This helps me during development.

### DIFF
--- a/compiler/generate_build_properties.gradle
+++ b/compiler/generate_build_properties.gradle
@@ -8,6 +8,7 @@ def generateBuildProperties = project.tasks.register('generateBuildProperties') 
   File buildPropertiesFile = new File(new File(project.buildDir, generatedDirPath), 'BuildProperties.kt')
 
   inputs.property 'useIR', rootProject.ext.kotlinUseIR
+  inputs.property 'useIR', rootProject.ext.fullTestRun
 
   outputs.file buildPropertiesFile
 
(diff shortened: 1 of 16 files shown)


compiler/generate_build_properties.gradle
```diff
@@ -21,6 +22,8 @@ def generateBuildProperties = project.tasks.register('generateBuildProperties')
 
       // With Kotlin 1.6 the old backend triggers deprecation warnings.
       internal const val WARNINGS_AS_ERRORS = ${rootProject.ext.kotlinUseIR && rootProject.ext.kotlinVersion.startsWith("1.6")}
+     
+      internal const val FULL_TEST_RUN = ${rootProject.ext.fullTestRun}
```
JoelWilcox, line 26: Based on the usages, it looks like when this flag is false, we only run tests focused on @MergeComponent. Am I understanding that correctly?

I think this could use a rename and/or a comment indicating what the difference is between a full test run and a non-full test run. Right now even with looking through the usages, that part is not 100% clear to me. Quantifying the need for this would be helpful too, like a rough estimation of average test runtime with and without the flag at this point in time.

JoelWilcox: Additionally, indicating when other contributors would want to use this flag :)

vRallev: Did you see this comment? https://github.com/square/anvil/pull/493/files#diff-d979b641bd0ea7c9da3fe113f9657636df1002652c75042cfe3b5203da064215R22-R24

vRallev: If the flag is false, then we run each test with one annotation instead of all options. We also skip tests that run the Dagger annotation processor (KAPT is slow).

JoelWilcox: 🤦  apparently my brain completely skipped over that comment and straight to the code. 

I'd say it's worth adding "If the flag is false, then we run each test with one annotation instead of all options. We also skip tests that run the Dagger annotation processor (KAPT is slow)." to the comment as well.

JoelWilcox (approved):

compiler/generate_build_properties.gradle
vRallev: Done

PR #499, opened 2022-02-09
Extend `AnvilContext` with a property for the `ModuleDescriptor` to make it more flexible

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -47,7 +49,7 @@ internal class CodeGenerationExtension(
   ): AnalysisResult? {
     // Tell the compiler that we have something to do in the analysisCompleted() method if
     // necessary.
-    return if (!didRecompile && codeGenerators.isNotEmpty()) AnalysisResult.EMPTY else null
```
vRallev, line 50: We know that `codeGenerators` is never empty. That check was unnecessary.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -75,7 +79,9 @@ internal class CodeGenerationExtension(
     val generatedFiles = mutableMapOf<String, GeneratedFile>()
 
     val (privateCodeGenerators, nonPrivateCodeGenerators) =
-      codeGenerators.partition { it is PrivateCodeGenerator }
+      codeGenerators
+        .filter { it.isApplicable(anvilContext) }
```
vRallev, line 83: I had to move the check here in order to get the context. It also feels cleaner. It shouldn't be the responsibility of the registrar.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -99,10 +99,7 @@ internal class ContributesSubcomponentHandlerGenerator(
 
   private var isFirstRound = true
 
-  override fun isApplicable(context: AnvilContext): Boolean =
-    throw NotImplementedError(
-      "This should not actually be checked as we instantiate this class manually."
-    )
+  override fun isApplicable(context: AnvilContext): Boolean = !context.generateFactoriesOnly
```
vRallev, line 102: I also feel better about this.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtensionTest.kt
```diff
@@ -122,4 +122,38 @@ class CodeGenerationExtensionTest {
       assertThat(exitCode).isEqualTo(OK)
     }
   }
+
+  @Test fun `a code generator that is not applicable is never called`() {
```
vRallev, line 126: Aaand we can test this now :)

compiler/src/main/java/com/squareup/anvil/compiler/AnvilCommandLineProcessor.kt
```diff
@@ -80,3 +80,18 @@ class AnvilCommandLineProcessor : CommandLineProcessor {
     }
   }
 }
+
+class CommandLineOptions private constructor(
```
JoelWilcox, line 84: What do you think about moving this to its own file? Alternatively, if we bundle it in the file with another class, it feels like bundling it with `CompilerConfiguration` might be better if we're enforcing that it can only be instantiated from that object.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
JoelWilcox: > We know that codeGenerators is never empty. That check was unnecessary.

Is this actually guaranteed? From looking at the constructor and field initialization, it doesn't seem like this class knows that (even if we might know it outside of this class for its usage).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/RealAnvilContext.kt
```diff
@@ -1,9 +1,21 @@
...
-  override val disableComponentMerging: Boolean
+  override val disableComponentMerging: Boolean,
+  override val module: ModuleDescriptor
 ) : AnvilContext
+
+internal fun CommandLineOptions.toAnvilContext(
+  module: ModuleDescriptor
+): RealAnvilContext = RealAnvilContext(
+  generateFactories = generateFactories,
+  generateFactoriesOnly = generateFactoriesOnly,
```
JoelWilcox, line 18: General question: When do we only generate factories? Curious to understand the use-case

JoelWilcox, line 18: There might also be an opportunity for an enum here to simplify all the spots that know about both of these values, since in theory if `generateFactoriesOnly` is true, `generateFactories` should also be true. e.g.
```
enum CodeGenerationSetting {
  STANDARD,
  STANDARD_WITH_DAGGER_FACTORIES,
  DAGGER_FACTORIES_ONLY
}
```

JoelWilcox (approved):
Left some suggestions but overall lgtm

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
vRallev: We know that, because within Anvil itself are plenty of code generators. It's okay for this class to have some assumptions about the internals of the compiler plugin.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/RealAnvilContext.kt
vRallev: That was actually a request from someone else. Maybe Slack? I don't remember. But we must support it now 🙃 

The idea is to generate all factories for Dagger to get rid of Kapt, but remove all other functionality from Anvil. It helps to convince teams to adopt Anvil.

vRallev: That would require changing `AnvilContext` and that's part of our stable API I'd like to avoid touching.

compiler/src/main/java/com/squareup/anvil/compiler/AnvilCommandLineProcessor.kt
vRallev: That's fair.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/RealAnvilContext.kt
JoelWilcox: Gotcha, makes sense!

PR #501, opened 2022-02-10
Partially migrate `ContributesMultibindingGenerator` to solely rely on `ClassReference` (1/x)

JoelWilcox:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/CompilerUtils.kt
```diff
@@ -168,6 +168,8 @@ public fun List<KotlinType>.getAllSuperTypes(): Sequence<FqName> =
     .map { it.classDescriptor().fqNameSafe }
 
 @ExperimentalAnvilApi
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated("Don't rely on PSI and make the code agnostic to the underlying implementation. See [AnnotationReference#isQualifier]")
```
JoelWilcox, line 172: Fix this comment, PSI -> descriptor

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -41,12 +51,46 @@ public sealed class AnnotationReference {
...
+  }
 
   public class Descriptor internal constructor(
     public val annotation: AnnotationDescriptor,
     override val classReference: ClassReference
-  ) : AnnotationReference()
+  ) : AnnotationReference() {
+
+    override fun isQualifier(): Boolean {
+      return annotation.annotationClass?.annotations?.hasAnnotation(qualifierFqName) ?: false
```
vRallev, line 87: This can be simplified!
```suggestion
      return classReference.isAnnotatedWith(qualifierFqName)
```
`classReference` is a property in `AnnotationReference` and refers to the class of the annotation. It's the same as calling `annotation.annotationClass!!.toClassReference(..)`. 

Prefer using `ClassReference` even in the functions, because many objects are cached like querying annotations.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -41,12 +51,46 @@ public sealed class AnnotationReference {
...
+    }
+
+    override fun isMapKey(): Boolean {
+      return annotation.annotationClass?.annotations?.hasAnnotation(MapKey::class.fqName) ?: false
```
vRallev, line 91: Same here. Also use `mapKeyFqName` instead of `MapKey::class.fqName` to avoid creating unnecessary instances.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -41,12 +51,46 @@ public sealed class AnnotationReference {
...
+      return annotation
+        .typeReference
+        ?.requireFqName(module)
+        // Often entries are annotated with @Inject, in this case we know it's not a qualifier and
+        // we can stop early.
+        ?.takeIf { it != injectFqName }
+        ?.classDescriptor(module)
+        ?.annotations
+        ?.hasAnnotation(qualifierFqName)
+        ?: false
```
vRallev, line 66: Same as for descriptors, you should be able to simplify that:
```suggestion
      return classReference.isAnnotatedWith(qualifierFqName)
```
This should work. If it does, then you don't even need an abstract function. The implementation for descriptors and PSI is the same thanks to `ClassReference` 😄

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -84,6 +84,16 @@ public sealed class ClassReference {
    */
   public abstract fun enclosingClassesWithSelf(): List<ClassReference>
 
+  public fun checkHasVisibility(visibility: Visibility) {
+    if (visibility() != visibility) {
+      throw AnvilCompilationExceptionClassReference(
+        classReference = this,
+        message = "$fqName is binding a type, but the class is not ${visibility.name}. Only " +
+          "public types are supported."
```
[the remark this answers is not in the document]
vRallev: I'd even pass in a lambda argument for the message similar to `checkNotNull` and `requireNotNull`.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingUtils.kt
```diff
@@ -0,0 +1,23 @@
+package com.squareup.anvil.compiler.codegen
```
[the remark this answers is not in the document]
JoelWilcox: Good question. I agree but was also conflicted because some extensions are specific to certain annotations, like `ClassReference.checkNotMoreThanOneQualifier` is specific to ContributesBinding/ContributesMultibinding. I suppose for that one we could remove the comment about those two annotations or rename it to make it more obvious? or maybe it's just fine if some extensions aren't generic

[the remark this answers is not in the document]
JoelWilcox: I made a small adjustment to the comment in that method which I think makes it feel better, and went ahead with renaming the file to `ClassReferenceExtensions.kt`

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -84,6 +84,16 @@ public sealed class ClassReference {
    */
   public abstract fun enclosingClassesWithSelf(): List<ClassReference>
 
+  /**
+   * @param failedMessage The message included in the thrown Exception when this [ClassReference]
+   * doesn't have the expected visibility.
+   */
+  public fun checkHasVisibility(visibility: Visibility, failedMessage: String) {
```
JoelWilcox, line 91: Initially I was going to include the error message here which made it feel worth having a dedicated method for this, but now this feels like not enough code to be worth it. Thoughts?

vRallev: I agree with your conclusion and I would probably inline this method again. You'll find several places in the codebase that use very similar error messages (especially when PSI or descriptors are mixed), but it was never reasonable to extract a common function. Sometimes having more flexibility and a little more duplication is the right thing. The gut feeling should decide :)

BTW, I was suggesting: 
```kotlin 
inline fun checkHasVisbility(visibility: Visibility: message: () -> String) { ... }
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
```diff
@@ -0,0 +1,52 @@
+package com.squareup.anvil.compiler.codegen
+
+import com.squareup.anvil.compiler.contributesMultibindingFqName
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionClassReference
+import com.squareup.anvil.compiler.internal.reference.ClassReference
+import com.squareup.anvil.compiler.internal.reference.Visibility
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.util.capitalizeDecapitalize.toLowerCaseAsciiOnly
```
vRallev, line 8: I've never seen this function before 😅  I think `"".lowercase()` is the new way in the std function (the other methods are deprecated).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGenerator.kt
```diff
@@ -47,12 +43,16 @@ internal class ContributesMultibindingGenerator : CodeGenerator {
...
+      .classAndInnerClassReferences(module)
+      .filter { it.isAnnotatedWith(contributesMultibindingFqName) }
       .onEach { clazz ->
         clazz.checkClassIsPublic()
-        clazz.checkNotMoreThanOneQualifier(module, contributesMultibindingFqName)
-        clazz.checkNotMoreThanOneMapKey(module)
+        clazz.checkNotMoreThanOneQualifier(contributesMultibindingFqName)
+        clazz.checkNotMoreThanOneMapKey()
+      }
+      // TODO: Continue migrating the rest of operations to use ClassReference instead of PSI/Descriptors
```
vRallev, line 53: Sweet!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -25,6 +27,10 @@ public sealed class AnnotationReference {
   public val fqName: FqName get() = classReference.fqName
   public val module: AnvilModuleDescriptor get() = classReference.module
 
+  public fun isQualifier(): Boolean = classReference.isAnnotatedWith(qualifierFqName)
+
+  public fun isMapKey(): Boolean = classReference.isAnnotatedWith(mapKeyFqName)
```
vRallev, line 32: It's so simple, I love it.

vRallev (approved):
Looks great! This doesn't need to be a draft anymore :-)

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
JoelWilcox: Lol same. Ahh nice, didn't even see that since I started `.toL` and it only showed the deprecated ones + the Ascii one. Updated now.

PR #502, opened 2022-02-10
Suppress errors for experimental APIs in generated code.

### LINKED ISSUE #488: Suppress/OptIn Lint warnings/errors are not ignored in Anvil/src-gen
We have some `@OptIn` Errors/Warnings.  We add to certain classes to make sure Engineers are aware of the fallout from using certain classes, however the Kotlin Compiler will flag/fail on these classes when the anvil/src-gen is compiled.

Not sure the best approach, but I'm guessing either Anvil add's the Suppression/OptIn, or has a compiler flag set to ignore lint warnings? Looks like Dagger/KSP/KAPT might ignore these as they don't warn/fail on these lint checks.

(If theres already a plugin option appologies!)

vRallev:
It's impossible to suppress these errors in generated code on the file level for the user, so we must opt-out of all these checks. It's not a bad thing, because the user already must opt-into these features on the usage side.

Fixes #488.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
@@ -315,7 +315,23 @@ private fun FileSpec.writeToString(): String {
 }
 
 private fun FileSpec.Builder.suppressWarnings() {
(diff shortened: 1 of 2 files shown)


0xMatthewGroves (approved):

vRallev:
Damn, Kotlin 1.6.20-M1 fails. I filed: https://youtrack.jetbrains.com/issue/KT-51351

JoelWilcox (approved):

chrisjenx:
Awesome!

PR #504, opened 2022-02-12
Use ClassReference when checking for a single super type in `ContributesMultibindingGenerator`

JoelWilcox:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/CompilerUtils.kt
```diff
@@ -102,6 +102,13 @@ public fun KotlinType.classDescriptor(): ClassDescriptor {
     )
 }
 
+@ExperimentalAnvilApi
+public fun AnnotationDescriptor.boundTypeOrNull(module: ModuleDescriptor): ClassDescriptor? {
```
vRallev, line 106: As far as I can tell this is ONLY used in the ClassDescriptor. Then we don't need this extra utility and it should be inlined. If it's used elsewhere, then this utility should be deprecated.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -859,6 +859,16 @@ public fun KtClassOrObject.generateClassName(
         .asString()
     }
 
+@ExperimentalAnvilApi
+public fun KtClassOrObject.boundTypeOrNull(
```
vRallev, line 863: Same here.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -77,6 +79,8 @@ public sealed class ClassReference {
   public abstract fun isInterface(): Boolean
   public abstract fun isAbstract(): Boolean
   public abstract fun visibility(): Visibility
+  public abstract fun boundTypeOrNull(annotationFqName: FqName): FqName?
```
vRallev, line 82: This function needs to be moved into `AnnotationReference`. An annotation has a bound type, not a class. This is one of the utilities that needs to be refactored more.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -191,6 +208,18 @@ public sealed class ClassReference {
...
     }
 
+    override fun boundTypeOrNull(annotationFqName: FqName): FqName? {
+      return clazz.annotation(annotationFqName)
+        .boundTypeOrNull(module)
+        ?.fqNameSafe
+    }
+
+    override fun directSuperClassReferences(): Sequence<ClassReference> {
+      return clazz.directSuperClassAndInterfaces()
```
vRallev, line 218: `directSuperClassAndInterfaces` is only used here. This utility can be inlined.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -191,6 +208,18 @@ public sealed class ClassReference {
...
+        .asSequence()
+        .map { it.fqNameSafe.toClassReference(module) }
```
vRallev, line 220:
```suggestion
        .map { it.toClassReference(module) }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -144,6 +148,19 @@ public sealed class ClassReference {
       }
     }
 
+    override fun boundTypeOrNull(annotationFqName: FqName): FqName? =
+      clazz.boundTypeOrNull(annotationFqName, module)
+
+    override fun directSuperClassReferences(): Sequence<ClassReference> {
```
vRallev, line 154: Similar to `enclosingClassesWithSelf` I think we should simply cache a list and only query that information once.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
```diff
@@ -51,3 +51,22 @@ internal fun ClassReference.checkNotMoreThanOneMapKey() {
...
 }
+
+internal fun ClassReference.checkSingleSuperType(
+  annotationFqName: FqName
+) {
+  // If the bound type exists, then you're allowed to have multiple super types. Without the bound
+  // type there must be exactly one super type.
+  if (boundTypeOrNull(annotationFqName) != null) return
+
+  val superTypes = directSuperClassReferences()
```
vRallev, line 62: Nit: I think it's okay to inline this variable.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
JoelWilcox: Not sure what level of additional refactoring you were thinking, I went ahead and moved this to `AnnotationReference` and updated usages to find the annotation they want first, then call `boundTypeOrNull` on it. Let me know if you had additional refactoring in mind.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -27,6 +32,8 @@ public sealed class AnnotationReference {
   public val fqName: FqName get() = classReference.fqName
   public val module: AnvilModuleDescriptor get() = classReference.module
 
+  public abstract fun boundTypeOrNull(): FqName?
```
vRallev, line 35: I'm sorry, but I didn't catch this the first time. This method should now return `ClassReference?`.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -78,6 +79,12 @@ public sealed class ClassReference {
   public abstract fun isAbstract(): Boolean
   public abstract fun visibility(): Visibility
 
+  /**
+   * Returns only the super class (excluding [Any]) and implemented interfaces declared directly by
+   * this class.
+   */
+  public abstract fun directSuperClassReferences(): Sequence<ClassReference>
```
vRallev, line 86: That should become a `List`. Sequences are evaluated over and over again and the lazy below becomes meaningless.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -118,6 +125,16 @@ public sealed class ClassReference {
       clazz.annotationEntries.map { it.toAnnotationReference(module) }
     }
 
+    private val directSuperClassReferences: Sequence<ClassReference> by lazy(NONE) {
+      if (clazz is KtEnumEntry) {
+        emptySequence()
```
vRallev, line 130: Hm, I know this if-check is coming from the original utility. But I believe it's wrong. An enum can implement other interfaces.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
vRallev: No, I meant exactly what you did 👍🏻

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -55,23 +55,24 @@ public sealed class AnnotationReference {
...
     public val annotation: AnnotationDescriptor,
     override val classReference: ClassReference
   ) : AnnotationReference() {
-    override fun boundTypeOrNull(): FqName? {
+    override fun boundTypeOrNull(): ClassReference? {
       return (annotation.getAnnotationValue("boundType") as? KClassValue)
         ?.argumentType(module)
         ?.classDescriptor()
         ?.fqNameSafe
+        ?.toClassReference(module)
```
vRallev, line 75: That's cheaper to go from the descriptor to the class reference.
```suggestion
        ?.classDescriptor()
        ?.toClassReference(module)
```

vRallev (approved):
One last change, otherwise it looks great 👍🏻

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
JoelWilcox: Just curious, why is this cheaper? Does FqName.toClassReference need to figure out psi vs descriptor at some point?

vRallev: If we only have an `FqName`, then we try to resolve the class descriptor first. We need to know that the `FqName` can be mapped to a class . This is the expensive operation. Since you already have a descriptor here, you can skip this step.

PR #506, opened 2022-02-14
Move `scope()`, `parentScope()`, `replaces()` and `innerClasses()`

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -86,15 +137,31 @@ public fun AnnotationDescriptor.toAnnotationReference(module: ModuleDescriptor):
...
+    message = message,
+    cause = cause
+  )
+}
+
+private fun replacesIndex(annotationFqName: FqName): Int {
+  return when (annotationFqName) {
+    contributesToFqName -> 1
+    contributesBindingFqName, contributesMultibindingFqName -> 2
+    contributesSubcomponentFqName -> 4
```
JoelWilcox, line 162: If these indexes are used anywhere else, I feel like we should make constants for them as well.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnnotationReferenceExtensions.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.squareup.anvil.compiler.internal.reference.AnnotationReference.Psi
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionAnnotationReference
+import com.squareup.anvil.compiler.internal.reference.ClassReference
+import com.squareup.anvil.compiler.internal.reference.toClassReference
+import com.squareup.anvil.compiler.internal.requireFqName
+import org.jetbrains.kotlin.psi.KtClassLiteralExpression
+import org.jetbrains.kotlin.resolve.constants.KClassValue
+
+internal fun AnnotationReference.parentScope(): ClassReference {
+  return when (this) {
```
JoelWilcox, line 18: What are your thoughts on defining this method in `AnnotationReference` so we can avoid the `when`?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnnotationReferenceExtensions.kt
```diff
@@ -0,0 +1,37 @@
...
+        )
+        ?.requireFqName(module)
+        ?.toClassReference(module)
+
+    is Descriptor ->
+      (annotation.getAnnotationValue("parentScope") as? KClassValue)
+        ?.argumentType(module)
+        ?.classDescriptor()
+        ?.toClassReference(module)
+  } ?: throw AnvilCompilationExceptionAnnotationReference(
```
JoelWilcox, line 33: Is there a use-case where the annotation may not have a parent scope and we'd want to return null instead? Or is the assumption that this will only be called on annotations where one should exist?

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -47,12 +66,49 @@ public sealed class AnnotationReference {
...
     override val classReference: ClassReference
-  ) : AnnotationReference()
+  ) : AnnotationReference() {
+
+    private val scope by lazy(NONE) { computeScope(0) }
+
+    // We need the scope so often that it's better to cache the value. Since the index could be
+    // potentially different, only cache the value for the default index.
+    override fun scopeOrNull(parameterIndex: Int): ClassReference? =
+      if (parameterIndex == 0) scope else computeScope(parameterIndex)
```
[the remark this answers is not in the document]
vRallev: Good point, done.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
vRallev: They aren't used anywhere else. They're only hardcoded here and the values are implied by the order of the parameters in the annotation.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnnotationReferenceExtensions.kt
vRallev: At the moment the utilities are defined in the compiler-utils module, but they're only used in the compiler module. I don't really want to expose this function in the compiler-utils modules and decided therefore to move it into the compiler module. In the compiler module we can only use extension functions.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/AnnotationReferenceExtensions.kt
vRallev: Yes, the assumption right now is that it's only called for the correct annotations. I thought about introducing subclasses for `AnnotationReference`, but was afraid that it could be much more complicated (another use case is `boundType`). For now I'd leave it as is and if we need an `..orNull()` function, then we can easily add it.

JoelWilcox (approved):

PR #511, opened 2022-02-16
Complete migration to use ClassReference in ContributesMultibindingGenerator

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGenerator.kt
```diff
@@ -52,16 +51,21 @@ internal class ContributesMultibindingGenerator : CodeGenerator {
...
         val fileName = clazz.generateClassName()
         val generatedPackage = HINT_MULTIBINDING_PACKAGE_PREFIX +
-          clazz.containingKtFile.packageFqName.safePackageString(dotPrefix = true)
+          clazz.packageFqName.safePackageString(dotPrefix = true)
         val className = clazz.asClassName()
-        val classFqName = clazz.requireFqName().toString()
+        val classFqName = clazz.fqName.toString()
         val propertyName = classFqName.replace('.', '_')
-        val scope = clazz.scope(contributesMultibindingFqName, module).asClassName(module)
+        val scope = clazz.annotations.find { it.fqName == contributesMultibindingFqName }
```
vRallev, line 61: Please use `single`. I think `find` returns the first instance and ignores other annotations. For now we assume there is exactly one annotation and `single` ensures that. As soon as we have repeated annotations `single` will blow up and throw errors as expected.

vRallev (approved):

PR #513, opened 2022-02-16
Migrate most usages of PSI to ClassReference in BindingModuleGenerator

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -461,25 +453,26 @@ private fun ContributedBinding.toGeneratedMethod(
 private val Collection<GeneratedMethod>.specs get() = map { it.spec }
 
 private class MergedClassKey(
-  val scope: FqName,
-  val clazz: KtClassOrObject
+  val scope: ClassReference,
+  val clazz: ClassReference
```
vRallev, line 457: I like this change.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -461,25 +453,26 @@ private fun ContributedBinding.toGeneratedMethod(
...
 
     return true
   }
 
   override fun hashCode(): Int {
-    var result = scope.hashCode()
+    var result = scopeFqName.hashCode()
     result = 31 * result + classFqName.hashCode()
     return result
   }
```
vRallev, line 478: These functions are redundant now. You can make `MergedClassKey` a data class, because `ClassReference` itself implements equals() and hashcode(). That wasn't possible before, because `KtClassOrObject` didn't implement these methods.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -73,7 +71,7 @@ internal class BindingModuleGenerator(
 
   private val contributedBindingClasses = mutableListOf<FqName>()
   private val contributedMultibindingClasses = mutableListOf<FqName>()
```
vRallev, line 73: Could these be converted to `ClassReference` too? Now that we have `ClassReference` working with `FqName` is bad practice, it's a very weak type.

vRallev (approved):
Awesome migration!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -297,37 +295,28 @@ internal class BindingModuleGenerator(
         (ktProperty.initializer as? KtClassLiteralExpression)
           ?.firstChild
           ?.requireFqName(module)
+          ?.toClassReference(module)
```
JoelWilcox, line 298: @vRallev let me know how you feel about this; I initially tried converting the whole method body to use `ClassReference`s and `PropertyReference`s but I did not see a clear way to replicate the 
```
(ktProperty.initializer as? KtClassLiteralExpression)
          ?.firstChild
          ?.requireFqName(module)
```

logic for descriptors.

vRallev: Add a TODO and we can maybe take a look together at it later.

PR #514, opened 2022-02-16
Migrate `ContributesSubcomponentHandlerGenerator` to only use `ClassReference`

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
```diff
@@ -535,15 +484,15 @@ internal class ContributesSubcomponentHandlerGenerator(
 
       if (scope != other.scope) return false
       if (parentScope != other.parentScope) return false
-      if (clazzFqName != other.clazzFqName) return false
+      if (clazz != other.clazz) return false
 
       return true
     }
 
     override fun hashCode(): Int {
```
JoelWilcox, line 492: Are we able to make the same change here as you recommended for `MergedClassKey`? Since it only has `AnnotationReference` as a field now, it feels like it would also be able to benefit from the existing equals/hashcode implementation

JoelWilcox (approved):
looks great! 🎉

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentHandlerGenerator.kt
vRallev: No, and this points out that the hash function for `AnnotationReference` is wrong. We only hash the `FqName`. That means when we have repeatable annotations with different parameters two different annotations are considered equal even though they shouldn't be. I'll address this separately.

PR #516, opened 2022-02-17
Move more utilities into a smaller scope and more specific package.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
```diff
@@ -348,3 +309,235 @@ private fun ClassReference.declaredMemberInjectParameters(
...
+  return unsubstitutedMemberScope
+    .getContributedDescriptors(DescriptorKindFilter.VARIABLES)
+    .filterIsInstance<PropertyDescriptor>()
+    // only return properties which are directly declared (including overrides)
+    .filter { it.kind == Kind.DECLARATION }
+    .filter { it.hasAnnotation(injectFqName) }
+}
+
+// TODO: use PropertyReference
+fun PropertyDescriptor.hasAnnotation(annotationFqName: FqName): Boolean {
```
JoelWilcox, line 374: Should these public extensions be `internal`?

JoelWilcox (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
vRallev: Yes!

JoelWilcox (approved):

PR #519, opened 2022-02-17
Convert `ProvidesMethodFactoryGenerator` to use `ClassReference`

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -62,67 +56,75 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
-      .classesAndInnerClasses(module)
-      .filter { it.hasAnnotation(daggerModuleFqName, module) }
+      .classAndInnerClassReferences(module)
+      .filter { it.isAnnotatedWith(daggerModuleFqName) }
       .forEach { clazz ->
-        clazz
-          .functions(includeCompanionObjects = true)
+        (clazz.companionObjects() + clazz)
+          .flatMap { it.functions }
           .asSequence()
```
vRallev, line 64: Nit: I'd even move the `asSequence()` one line up.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -62,67 +56,75 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
...
+              clazz,
+              CallableReference(function = function)
+            )
           }
 
-        clazz
-          .properties(includeCompanionObjects = true)
+        (clazz.companionObjects() + clazz)
+          .flatMap { it.properties }
           .asSequence()
```
vRallev, line 94: Same here.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -338,4 +339,36 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
       name
     }
   }
+
+  // TODO: May be worth extracting some of this to an interface to actually apply to the Reference
+  // classes if we think it would be useful in other use-cases
```
vRallev, line 344: When I looked at this I thought the same. I was thinking of introducing a `CallableReference` interface and make `FunctionReference` and `PropertyReference` implement this interface. 

On the other hand, it is only used for the `ProvidesMethodFactoryGenerator`. So maybe we could keep this private to this class for now? 

What do you think?

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
JoelWilcox: Since there aren't any other use-cases yet, I'm ok keeping it as just the private class for now. However, I think it's worth us looking at what fields/methods are common across all the different Reference types and potentially extracting a generic interface for those. E.g. it seems all References will have annotations, fqName, and visibility. I also figured if we do put any of these in interfaces, it would make sense to do it separately once you're changes are in as well to minimize the conflicts :)

PR #520, opened 2022-02-18
Convert `AssistedFactoryGenerator` to rely on `ClassReference` only. 

vRallev:
That required moving many utilities around to better places, but we were also able to delete some sketch pieces.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
@@ -9,7 +9,6 @@ import com.squareup.anvil.compiler.internal.reference.ClassReference.Psi
 import com.squareup.anvil.compiler.internal.reference.allSuperTypeClassReferences
 import com.squareup.anvil.compiler.internal.reference.asAnvilModuleDescriptor
 import com.squareup.anvil.compiler.internal.reference.canResolveFqName
-import com.squareup.anvil.compiler.internal.reference.indexOfTypeParameter
 import com.squareup.anvil.compiler.internal.reference.toClassReference
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.descriptors.ClassDescriptor
@@ -43,12 +42,8 @@ import org.jetbrains.kotlin.psi.KtValueArgument
 import org.jetbrains.kotlin.psi.KtValueArgumentName
 import org.jetbrains.kotlin.psi.psiUtil.collectDescendantsOfType
 import org.jetbrains.kotlin.psi.psiUtil.containingClass
-import org.jetbrains.kotlin.psi.psiUtil.getChildOfType
 import org.jetbrains.kotlin.psi.psiUtil.parents
 import org.jetbrains.kotlin.psi.psiUtil.parentsWithSelf
-import org.jetbrains.kotlin.types.DefinitelyNotNullType
-import org.jetbrains.kotlin.types.FlexibleType
-import org.jetbrains.kotlin.types.KotlinType
 import kotlin.reflect.KClass
 
 private val kotlinAnnotations = listOf(jvmSuppressWildcardsFqName, publishedApiFqName)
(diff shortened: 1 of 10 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -542,128 +537,6 @@ public fun KtTypeParameter.requireIdentifier(): String = nameIdentifier?.text
...
- * ```
- *
- * The KtTypeReference `T` will fail to resolve, since it isn't a type. This function will instead
- * look to the `ServiceFactory` interface, then look at the supertype declaration in order to
- * determine the type.
- *
- * @receiver The class which actually references the type. In the above example, this would be
- *   `ServiceFactory`.
- */
-@ExperimentalAnvilApi
```
vRallev, line 567: I moved these into `ClassReference.Psi`. It requires a PSI implementation, hence I used the concrete `Psi` class as new home.

vRallev: Sidenote: @RBusarow did a tremendous job with implementing this quite complicated utilities.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -61,6 +72,7 @@ public sealed class ClassReference {
   public val shortName: String get() = fqName.shortName().asString()
   public val packageFqName: FqName get() = classId.packageFqName
 
+  public abstract val constructors: List<FunctionReference>
```
vRallev, line 75: New API is needed.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -199,10 +339,15 @@ public sealed class ClassReference {
...
-    override val functions: List<FunctionReference> by lazy(NONE) {
+    override val constructors: List<FunctionReference.Descriptor> by lazy(NONE) {
+      clazz.constructors.map { it.toFunctionReference(this) }
+    }
+
+    override val functions: List<FunctionReference.Descriptor> by lazy(NONE) {
       clazz.unsubstitutedMemberScope
         .getContributedDescriptors(kindFilter = DescriptorKindFilter.FUNCTIONS)
         .filterIsInstance<FunctionDescriptor>()
+        .filterNot { it is ConstructorDescriptor }
```
vRallev, line 350: That was one mistake.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -289,13 +434,6 @@ public fun KtClassOrObject.toClassReference(module: ModuleDescriptor): Psi =
 public fun FqName.toClassReferenceOrNull(module: ModuleDescriptor): ClassReference? =
   module.asAnvilModuleDescriptor().getClassReferenceOrNull(this)
 
-@Suppress("DeprecatedCallableAddReplaceWith")
-@Deprecated("Don't rely on PSI and make the code agnostic to the underlying implementation.")
-@ExperimentalAnvilApi
-public fun FqName.toClassReferencePsiOrNull(module: ModuleDescriptor): Psi? =
```
vRallev, line 295: This guy is finally gone. Solving that took me almost 2 days.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -33,14 +33,40 @@ public sealed class FunctionReference {
...
+  public abstract val parameters: List<ParameterReference>
 
   public abstract fun isAbstract(): Boolean
   public abstract fun visibility(): Visibility
 
+  /**
+   * The return type can be null for generic types like `T`. In this case try to resolve the
+   * return type with [resolveGenericReturnOrNull].
+   */
+  public abstract fun returnTypeOrNull(): ClassReference?
```
vRallev, line 50: I don't like that this is nullable right now. Same for `ParameterReference`. I'll explore if introducing a new type `TypeReference` could make the situation easier and better.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt
```diff
@@ -50,17 +50,6 @@ public fun FqName.canResolveFqName(
   module: ModuleDescriptor
 ): Boolean = module.asAnvilModuleDescriptor().resolveClassIdOrNull(classIdBestGuess()) != null
 
-@ExperimentalAnvilApi
-@Deprecated(
-  message = "Use ClassReference instead.",
-  replaceWith = ReplaceWith("classAndInnerClassReferences(module)")
-)
-public fun Collection<KtFile>.classesAndInnerClasses(
```
vRallev, line 58: And we're done. Good job @JoelWilcox 😃

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -469,70 +335,28 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
+                is AnnotationReference.Descriptor ->
+                  annotation.annotation.argumentValue("value")
+                    ?.value
+                    ?.toString()
+              }
+            }
+            .orEmpty()
         )
       }
     }
```
vRallev, line 362: Notice how much shorter this class is now because we use `ClassReference`.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -167,30 +150,31 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
...
 
-    val typeParameters = clazz.typeVariableNames(module)
+    val typeParameters = clazz.clazz.typeVariableNames(module)
 
     fun ClassName.parameterized(): TypeName {
       return if (typeParameters.isEmpty()) this else parameterizedBy(typeParameters)
     }
 
-    val generatedFactoryTypeName = returnTypeFqName.generateClassName(suffix = "_Factory")
+    // TODO: ClassReference
```
vRallev, line 166: I'll work on this next.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
RBusarow: aww shucks

RBusarow (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -33,14 +33,40 @@ public sealed class FunctionReference {
...
+
+  public fun returnType(): ClassReference = returnTypeOrNull()
+    ?: throw AnvilCompilationExceptionFunctionReference(
+      functionReference = this,
+      message = "Unable to get the return type for function $fqName."
+    )
+
+  public abstract fun resolveGenericReturnOrNull(
+    implementingClass: ClassReference.Psi
```
JoelWilcox, line 59: I'm assuming using Psi was necessary here but wondering if there's a way to just use `ClassReference` since this is for the generic `FunctionReference`

JoelWilcox (approved):
Nice work Ralf!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
vRallev: It's required, because the `resolveTypeReference` function requires PSI.

PR #521, opened 2022-02-18
Convert `ModuleMerger` to use `ClassReference`

JoelWilcox:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -287,3 +294,23 @@ private fun excludeIndex(annotationFqName: FqName): Int {
...
+    else -> throw NotImplementedError(
+      "Couldn't find index of modules argument for $annotationFqName."
+    )
+  }
+}
+
+private fun modulesName(annotationFqName: FqName): String {
+  return when (annotationFqName) {
+    mergeComponentFqName, mergeSubcomponentFqName, contributesSubcomponentFqName -> "modules"
+    mergeModulesFqName -> "includes"
```
vRallev, line 311: Nice!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -306,6 +306,28 @@ public fun FqName.toClassReference(module: ModuleDescriptor): ClassReference {
...
+): AnnotationReference? {
+  return annotations.singleOrNull { it.fqName == annotationName }
+    ?.takeIf { scopeName == null || it.scopeOrNull()?.fqName == scopeName }
+}
+
+@ExperimentalAnvilApi
+public fun ClassReference.annotation(
+  annotationName: FqName,
+  scopeName: FqName? = null
+): AnnotationReference {
```
vRallev, line 322: I really don't like these extension. Once we have repeatable annotations they no longer work. We shouldn't move the code into the wrong direction.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -81,6 +81,7 @@ public sealed class AnnotationReference {
 
   public abstract fun replaces(parameterIndex: Int = replacesIndex(fqName)): List<ClassReference>
   public abstract fun exclude(parameterIndex: Int = excludeIndex(fqName)): List<ClassReference>
+  public abstract fun modules(parameterIndex: Int = modulesIndex(fqName)): List<ClassReference>
```
JoelWilcox, line 84: I initially added this to replace some "modules" lookups in `ModuleMerger` but ended up reverting to what `ModuleMerger` was already doing because of it needing the lower level types. However I noticed this could replace an existing extension function that was coded for just one case, so I kept this. So you won't see any usages in this PR since it replaces the existing extension usage.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
JoelWilcox: :+1: let's talk offline about what we expect things to look like once we have repeatable annotations in and see if this can be done in a more future-proof way. I just found myself writing this code manually many many times during this PR and felt like it was better to consolidate the logic somewhere, but I'm down to make changes.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,112 +84,107 @@ internal class ModuleMerger(
         // modules contain the same bindings and are contributed to the same scope. To avoid this
         // issue we filter all generated Anvil modules except for the one that was generated for
         // this specific class.
-        val fqName = it.fqNameSafe
+        val fqName = it.fqName
         !fqName.isAnvilModule() || fqName == anvilModuleName
```
vRallev, line 88: You can inline the variable now, because we cache the value:
```suggestion
        !it.fqName.isAnvilModule() || it.fqName == anvilModuleName
```

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,112 +84,107 @@ internal class ModuleMerger(
...
-          it.annotationOrNull(contributesToFqName, scope = scopeFqName)
-            ?: return@mapNotNull null
-        it to contributesAnnotation
+      .mapNotNull { contributedClass ->
+        val contributesAnnotation = contributedClass.annotationOrNull(
+          annotationName = contributesToFqName,
+          scopeName = scopeFqName
+        ) ?: return@mapNotNull null
+
+        contributedClass to contributesAnnotation
```
vRallev, line 96: Creating this pair is unnecessary now. You can just return the annotation, because it keeps a reference to its class.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,112 +84,107 @@ internal class ModuleMerger(
...
+          "${contributedClass.fqName} is contributed to the Dagger graph, but the " +
+            "module is not public. Only public modules are supported."
         }
       }
       // Convert the sequence to a list to avoid iterating it twice. We use the result twice
       // for replaced classes and the final result.
       .toList()
 
-    val excludedModules = (mergeAnnotation.getAnnotationValue("exclude") as? ArrayValue)
+    val excludedModules = (mergeAnnotation.annotation.getAnnotationValue("exclude") as? ArrayValue)
```
vRallev, line 125: `AnnotationReference` gives you the excluded values. No need to work with descriptors.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,112 +84,107 @@ internal class ModuleMerger(
...
+            classReference = excludedClass
           )
 
-        if (scopeOfExclusion.fqNameSafe != scopeFqName) {
-          throw AnvilCompilationException(
-            thisDescriptor,
-            "${thisDescriptor.fqNameSafe} with scope $scopeFqName wants to exclude " +
-              "${classDescriptorForExclusion.fqNameSafe} with scope " +
-              "${scopeOfExclusion.fqNameSafe}. The exclusion must use the same scope."
+        if (scopeOfExclusion.fqName != scopeFqName) {
```
vRallev, line 149: Prefer comparing `scopeOfExclusion` to `scope` instead of using the FqNames. We implement equals() and get more control this way.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,112 +84,107 @@ internal class ModuleMerger(
...
-      }
-      ?: emptyList()
+      } ?: emptyList()
 
     val replacedModules = modules
-      .filter { (classDescriptor, _) ->
+      .filter { (contributedClass, _) ->
         // Ignore replaced modules or bindings specified by excluded modules.
-        classDescriptor.defaultType.asmType(codegen.typeMapper) !in excludedModules
+        contributedClass.clazz.defaultType.asmType(codegen.typeMapper) !in excludedModules
```
vRallev, line 164: Is this conversion necessary? Can't `excludedModules` hold instances of `ClassReference`?

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -212,21 +199,19 @@ internal class ModuleMerger(
           annotation = annotationFqName,
           scope = scopeFqName
         )
+        .map { it.toClassReference(module) }
         .flatMap { contributedClass ->
           val annotation = contributedClass.annotation(annotationFqName)
-          if (scopeFqName == annotation.scope(module).fqNameSafe) {
-            contributedClass.toClassReference(module)
-              .annotations.single { it.fqName == annotationFqName }
+          if (scopeFqName == annotation.scope().fqName) {
```
vRallev, line 205: Same here, consider comparing ClassReferences.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -247,21 +232,21 @@ internal class ModuleMerger(
...
+      findContributedSubcomponentModules(clazz, scopeFqName, module)
         .map { codegen.typeMapper.mapType(it) }
 
     val contributedModules = modules
       .asSequence()
-      .map { codegen.typeMapper.mapType(it.first) }
+      .map { codegen.typeMapper.mapType(it.first.clazz) }
       .minus(replacedModules.toSet())
       .minus(replacedModulesByContributedBindings.toSet())
       .minus(replacedModulesByContributedMultibindings.toSet())
```
vRallev, line 252: I think I can answer my own question. Here we need this other `Type`. But I think until this point we could fully work with `ClassReference` and then below convert them to this asm type.

JoelWilcox: Hmm I had originally converted everything to fully use ClassReference and went back because it looked like the conversion couldn't be done (and I had been missing this piece until the very end since it got deleted early). I think you're right that it would work and I could just use this method for each set here at the end. I'll give that a shot.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,141 +78,127 @@ internal class ModuleMerger(
...
+    val excludedModules = mergeAnnotation.exclude()
+      .onEach { excludedClass ->
+        val contributesToAnnotation =
+          excludedClass.findAnnotations(contributesToFqName).firstOrNull()
+        val contributesBindingAnnotation =
+          excludedClass.findAnnotations(contributesBindingFqName).firstOrNull()
+        val contributesMultibindingAnnotation =
+          excludedClass.findAnnotations(contributesMultibindingFqName).firstOrNull()
+        val contributesSubcomponentAnnotation =
+          excludedClass.findAnnotations(contributesSubcomponentFqName).firstOrNull()
```
vRallev, line 130: Same here for `firstOrNull()`.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,141 +78,127 @@ internal class ModuleMerger(
...
+        .map { it.toClassReference(module) }
         .flatMap { contributedClass ->
-          val annotation = contributedClass.annotation(annotationFqName)
-          if (scopeFqName == annotation.scope(module).fqNameSafe) {
-            contributedClass.toClassReference(module)
-              .annotations.single { it.fqName == annotationFqName }
+          val annotation = contributedClass.requireAnnotations(annotationFqName).first()
+          if (mergeScope == annotation.scope()) {
+            contributedClass
+              .requireAnnotations(annotationFqName).first()
```
vRallev, line 198: `.single()`. I won't call out other instances.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -92,141 +78,127 @@ internal class ModuleMerger(
...
-              "module is not public. Only public modules are supported."
-          )
+        contributedClass.checkClassIsPublic {
+          "${contributedClass.fqName} is contributed to the Dagger graph, but the " +
+            "module is not public. Only public modules are supported."
         }
+
+        moduleAnnotation != null || mergeModulesAnnotation != null
       }
+      .cast<Sequence<AnnotationReference.Descriptor>>()
```
vRallev, line 116: This cast call is weird. You can avoid this by changing your extension functions to 
```kotlin
public fun <T : AnnotationReference> List<T>.findAnnotations(
  annotationName: FqName,
  scopeName: FqName? = null
): List<T> {
```
(I'd move it into `AnnotationReference` in this case)

vRallev: I tested it locally and it works. But there's another issue you need to fix for `ClassReference.Descriptor`:
```
    override val annotations: List<AnnotationReference.Descriptor> by lazy(NONE) {
      clazz.annotations.map { it.toAnnotationReference(this, module) }
    }
```
The current implementation returns just `AnnotationReference`.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -288,13 +258,25 @@ internal class ModuleMerger(
...
+    return cast<Sequence<ClassReference.Descriptor>>()
+      .map { codegen.typeMapper.mapType(it.clazz) }
+      .toSet()
+  }
+
+  private fun List<ClassReference>.types(codegen: ImplementationBodyCodegen): Set<Type> {
+    return cast<List<ClassReference.Descriptor>>()
+      .map { codegen.typeMapper.mapType(it.clazz) }
+      .toSet()
+  }
```
vRallev, line 271:
```suggestion
  private fun Set<ClassReference>.types(codegen: ImplementationBodyCodegen): Set<Type> {
    return cast<Set<ClassReference.Descriptor>>()
      .map { codegen.typeMapper.mapType(it.clazz) }
  }
```
I'd change it to avoid the duplication. With this change we'll also rely on the equals() implementation within ClassReference, which gives us more control. That we relied on the internal compiler implementation always made me a little uneasy.

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
JoelWilcox: 👍  updated

> With this change we'll also rely on the equals() implementation within ClassReference, which gives us more control. That we relied on the internal compiler implementation always made me a little uneasy.

Could you explain that part of the comment though? It's not clear to me why this change has an effect on equals() being used

vRallev: If you convert a list to a set, then duplicated entries in the list will removed and not added to the set. For that the runtime will call `equals()`. If you convert `List<ClassReference>` to a set, then the `equals` method for `ClassReference` will be used and not the method from the Kotlin compiler type. 

So originally it was:
```
List<ClassReference> -> map to List<Type> -> to set (uses equals from Type)
```
Now it's:
```
List<ClassReference> -> to set (uses equals from ClassReference) -> map to Set<Type> 
```

JoelWilcox: got it, thanks!

PR #522, opened 2022-02-18
Fix wrong `FqName` and resolve TODO.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/CompilerUtils.kt
```diff
@@ -231,21 +231,6 @@ public fun String.capitalize(): String = replaceFirstChar(Char::uppercaseChar)
...
-): ClassId = classIdBestGuess().generateClassName(separator = separator, suffix = suffix)
-
-@ExperimentalAnvilApi
-public fun ClassId.generateClassName(
-  separator: String = "_",
-  suffix: String = ""
-): ClassId {
-  val className = relativeClassName.asString().replace(".", separator)
-  return ClassId(packageFqName, FqName(className + suffix), false)
-}
```
vRallev, line 247: 👋🏻

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -85,7 +89,7 @@ public sealed class FunctionReference {
   public class Psi internal constructor(
     public val function: KtFunction,
     override val declaringClass: ClassReference.Psi,
-    override val fqName: FqName = function.fqName ?: declaringClass.fqName
```
vRallev, line 88: I accidentally committed that in my previous PR. This was wrong and I forgot to fix it.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryGenerator.kt
```diff
@@ -163,16 +162,15 @@ internal class AssistedFactoryGenerator : PrivateCodeGenerator() {
       return if (typeParameters.isEmpty()) this else parameterizedBy(typeParameters)
     }
 
-    // TODO: ClassReference
-    val generatedFactoryTypeName = returnType.fqName.generateClassName(suffix = "_Factory")
```
vRallev, line 167: That TODO is solved.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
```diff
@@ -91,7 +93,7 @@ internal class MembersInjectorGenerator : PrivateCodeGenerator() {
         .joinToString()
     }
 
-    val memberInjectorClass = ClassName(packageName, className)
+    val memberInjectorClass = classId.asClassName()
```
JoelWilcox, line 96: What's the difference between this and `className` (classId.relativeClassName.asString())? If this one is the fully qualified name, it might help with clarity to rename `className` more explicitly. I found the vals all blurring together while reading 😅

JoelWilcox (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MembersInjectorGenerator.kt
vRallev: A class has a package and a class name (sometimes multiple class names for inner classes. FqName can't differentiate between the package and class name, because `FqName` is basically a string, e.g. `com.squareup.Outer.Inner`. We usually consider the first element starting with an uppercase letter as the first class name, so we'd treat `com.squareup` as package and `Outer.Inner` as relative class name. That approach has problems, if you use uppercase letters in a package name and unfortunately that happens (what's wrong with those people?!).

`ClassId` (Kotlin compiler type) basically holds two FqNames: one for the package and one for the class names. `relativeClassName` would be `Outer.Inner` in this sample. I prefer `ClassId` over `FqName` any day.

`ClassName` is a type from KotlinPoet and very similar to `ClassId`. It needs to know the package and class name. Converting a `ClassId` to `ClassName` is simple, but converting a `FqName` to `ClassName` could go wrong. 

The old `generateClassName` method returned basically the relative class name only and not the package. Since I changed it to a `ClassId` we now have richer information, but therefore also must call `relativeClassName` to restore the old behavior. 

I hope that makes sense 😅

JoelWilcox: Yes that makes things much clearer, thanks for the thorough explanation!

PR #523, opened 2022-02-18
Add a testing API to remove all the boilerplate that's needed for code generators in unit tests

vRallev:
I was annoyed 🤷🏻


compiler/src/test/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtensionTest.kt
```diff
@@ -18,36 +16,19 @@ import java.io.File
...
+        ?.let {
+          //language=kotlin
+          """
+          package generated.com.squareup.test
+          
+          class Abc
+          
+          private const val abc = "${clazz.shortName}"
+        """
+        }
```
vRallev, line 31: This is so much shorter.

RBusarow (approved):
Nice!  I figured something like this would be coming.

PR #524, opened 2022-02-18
Implement several unit tests for ClassReference.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt
```diff
@@ -32,13 +32,8 @@ public interface AnvilModuleDescriptor : ModuleDescriptor {
...
    * [KtClassOrObject] if the descriptor resolution fails. This will happen if the code being
    * parsed was generated as part of the compilation round for this module.
-   *
-   * If [preferDescriptor] is `false`, then the order of resolution is reverted.
    */
-  public fun getClassReferenceOrNull(
-    fqName: FqName,
-    preferDescriptor: Boolean = true
-  ): ClassReference?
+  public fun getClassReferenceOrNull(fqName: FqName): ClassReference?
```
vRallev, line 36: A leftover that I meant to remove. That parameter was only used by a deprecated API.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+import com.squareup.anvil.compiler.internal.reference.Visibility
+import com.squareup.anvil.compiler.internal.reference.toClassReference
+import com.squareup.anvil.compiler.internal.testing.simpleCodeGenerator
+import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.OK
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtClassOrObject
+import org.jetbrains.kotlin.resolve.source.getPsi
+import org.junit.Test
+
+class ClassReferenceTest {
```
vRallev, line 15: It's a little bit funky, but works surprisingly well. I can live with that.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+      allWarningsAsErrors = false,
+      codeGenerators = listOf(
+        simpleCodeGenerator { psiRef ->
+          val descriptorRef = psiRef.toDescriptorReference()
+
+          when (psiRef.shortName) {
+            "SomeClass1" -> {
+              assertThat(psiRef.constructors).hasSize(0)
+
+              assertThat(descriptorRef.constructors).hasSize(1)
```
vRallev, line 39: Note that there are always differences between the PSI and Descriptor implementation, e.g. descriptors see the default constructor, but PSI doesn't.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+        simpleCodeGenerator { psiRef ->
+          val descriptorRef = psiRef.toDescriptorReference()
+
+          when (psiRef.shortName) {
+            "SomeClass1" -> {
+              assertThat(psiRef.functions).hasSize(0)
+              assertThat(descriptorRef.functions).hasSize(3)
+              assertThat(descriptorRef.functions.map { it.name }).containsExactly(
+                "equals", "toString", "hashCode"
+              )
```
vRallev, line 106: Similar situation for functions. There are differences.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+              ).isEqualTo(FqName("kotlin.String"))
+
+              assertThat(
+                descriptorFunction.resolveGenericReturnType(implementingClass2).fqName
+              ).isEqualTo(FqName("kotlin.String"))
+
+              // TODO: This is currently failing, but should work in theory. The problem is a mix
+              //  of ClassReferences in the super types being PSI and Descriptor implementations.
+              //  We need to implement resolving types for Descriptors as well. That it's only
+              //  supported for PSI is a smell.
```
vRallev, line 231: That's something I'd like to work on. But it was very hard to test or even to verify. That lead me down the path of finding this way to "unit test" `ClassReference`.

vRallev: Getting closer to having a solution. Having unit tests is so much easier.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+              // assertThat(
+              //   descriptorFunction.resolveGenericReturnType(implementingClass3).fqName
+              // ).isEqualTo(FqName("kotlin.String"))
+            }
+            "GenericInterface2" -> {
+              assertThat(psiRef.functions).hasSize(0)
+              assertThat(descriptorRef.functions).hasSize(4)
+
+              val descriptorFunction = descriptorRef.functions.single { it.name == "hello" }
+              assertThat(descriptorFunction.returnTypeOrNull()).isNull()
```
JoelWilcox, line 249: Is this assert to help verify that the method is using a generic type and make sure that the test source we're compiling doesn't accidentally change to return String directly?

I'm reading it as "function return types for generics are null until they're resolved"

JoelWilcox (approved):
Made a few suggestions but this looks great!

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+      }
+
+      interface GenericInterface1<T> {
+        fun hello(): T
+      }
+
+      interface GenericInterface2<S> : GenericInterface1<S>
+
+      class SomeClass2 : GenericInterface1<String> {
+        override fun hello(): String = "hello"
```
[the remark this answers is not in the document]
vRallev: Types from the Kotlin std library are actually trickier to resolve, because they require no import 😅  These tests only cover a very small percentage of features and aren't complete by any means.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
vRallev: At one point I definitely had comments. Will add them!

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -0,0 +1,291 @@
...
+                assertThat(it.fqName.asString())
+                  .isEqualTo("com.squareup.test.SomeClass3.<init>")
+              }
+
+              assertThat(descriptorRef.constructors).hasSize(2)
+              descriptorRef.constructors.forEach {
+                assertThat(it.fqName.asString())
+                  .isEqualTo("com.squareup.test.SomeClass3.<init>")
+              }
+            }
```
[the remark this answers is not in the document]
vRallev: Helper functions obfuscate what tests are doing and are often not that great. I can recommend you a good book that discusses the pros and cons in detail. I prefer not to have this function.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
vRallev: It just ensures the expectation that a generic type parameter is null as return type. The return type needs to be resolved for a specific implementation.

Those utilities are a tricky, that's why it also took me so long to refactor them and clean things up. It's confusing :-(

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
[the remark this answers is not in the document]
JoelWilcox: That's fair. I'd be interested in checking out the book.

PR #525, opened 2022-02-18
Improve resolving the return type of a function.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/RealAnvilModuleDescriptor.kt
```diff
@@ -102,7 +102,14 @@ class RealAnvilModuleDescriptor(
...
 
-    return descriptorClassReference() ?: psiClassReference()
+    // Prefer Psi to have consistent results. If the class is part of the compilation unit: it'll
+    // be a Psi implementation. If the class comes from a pre-compiled dependency, then it'll be a
+    // descriptor implementation.
+    //
+    // Otherwise, there are inconsistencies. If source code was written manually in this module,
+    // then we could use the descriptor or Psi implementation. If the source code was generated,
+    // then only Psi works.
+    return psiClassReference() ?: descriptorClassReference()
```
vRallev, line 112: That is what fixed the TODO and makes the test pass.

compiler/src/test/java/com/squareup/anvil/compiler/ClassReferenceTest.kt
```diff
@@ -250,10 +260,16 @@ class ClassReferenceTest {
...
-                descriptorFunction.resolveGenericReturnType(implementingClass).fqName
+                descriptorFunction
+                  .resolveGenericReturnType(implementingClass.toPsiReference())
+                  .fqName
+              ).isEqualTo(FqName("kotlin.String"))
+              assertThat(
+                descriptorFunction
+                  .resolveGenericReturnType(implementingClass.toDescriptorReference())
+                  .fqName
               ).isEqualTo(FqName("kotlin.String"))
```
vRallev, line 273: I'm testing every combination of descriptor and Psi now.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -59,11 +59,42 @@ public sealed class FunctionReference {
...
 
-  public abstract fun resolveGenericReturnTypeOrNull(
+  protected abstract fun resolveGenericReturnTypeOrNull(
     implementingClass: ClassReference.Psi
   ): ClassReference?
 
-  public fun resolveGenericReturnType(implementingClass: ClassReference.Psi): ClassReference =
+  public fun resolveGenericReturnTypeOrNull(
+    implementingClass: ClassReference
+  ): ClassReference? {
```
JoelWilcox, line 68: I found the relationship between this and the protected version a bit confusing; it might help to give the protected method a different name just for easier differentiation on which one is being used where, even if it's just `resolveGenericReturnTypeOrNullInternal`, `resolveGenericReturnTypeOrNullForPsi` or `_resolveGenericReturnTypeOrNull`

JoelWilcox (approved):
Nice work getting this figured out

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
vRallev: Fair point. I've changed it to `resolveGenericReturnTypeOrNullPsi`.

PR #527, opened 2022-02-19
Fix the equals() and hashCode() function for `AnnotationReference`.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -89,19 +89,21 @@ public sealed class AnnotationReference {
 
   override fun toString(): String = "@$fqName"
 
-  // TODO: equals() and hashcode() are wrong actually. To compare AnnotationReference instances we
-  //  also need to consider the annotation values and not just the FqName. Otherwise we'll have
-  //  many problems with repeatable annotations.
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
-    if (other !is ClassReference) return false
```
vRallev, line 97: This was a copy and paste error 🤦🏻

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
vRallev: I think we barely compare annotations nor put them in a set. This will happen with repeatable annotations, though. And you're right, when we look for an annotation we usually just use the fqname as identifier.

PR #528, opened 2022-02-19
Improve CI times of the Windows job.

vRallev:
Don't run the entire test suite for the Windows job, because we don't need to and it's the slowest of all jobs. The job only makes sure that we don't have any incompatibilities and therefore skipping the Dagger tests is fine.

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -31,10 +31,6 @@ ext {
   fullTestRun = (project.hasProperty('square.fullTestRun') ?
       project.getProperty('square.fullTestRun') : 'true').toBoolean()
 
-  if (ci && !fullTestRun) {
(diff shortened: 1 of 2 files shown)


dependencies.gradle
```diff
@@ -31,10 +31,6 @@ ext {
   fullTestRun = (project.hasProperty('square.fullTestRun') ?
       project.getProperty('square.fullTestRun') : 'true').toBoolean()
 
-  if (ci && !fullTestRun) {
```
JoelWilcox, line 34: Is it possible to keep this check but add an exclusion for when we're running on windows? 🤔 

The check seems like a useful safe-guard to make sure any future changes to not run all tests are intentionally committed

JoelWilcox (approved):

PR #532, opened 2022-02-22
Delete deprecated functions

vRallev:
* Add the option to get the name of an annotation argument.
* Clean up the utilities within `ClassReference`.
* Remove several unused or deprecated functions, clean up TODOs and use our reference implementations more broadly.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
@@ -5,40 +5,30 @@ package com.squareup.anvil.compiler.internal
 import com.squareup.anvil.annotations.ExperimentalAnvilApi
 import com.squareup.anvil.compiler.api.AnvilCompilationException
 import com.squareup.anvil.compiler.internal.reference.ClassReference
-import com.squareup.anvil.compiler.internal.reference.asAnvilModuleDescriptor
+import com.squareup.anvil.compiler.internal.reference.allSuperTypeClassReferences
 import com.squareup.anvil.compiler.internal.reference.canResolveFqName
 import com.squareup.anvil.compiler.internal.reference.toClassReference
+import com.squareup.anvil.compiler.internal.reference.toClassReferenceOrNull
 import org.jetbrains.kotlin.com.intellij.psi.PsiElement
-import org.jetbrains.kotlin.descriptors.ClassDescriptor
 import org.jetbrains.kotlin.descriptors.ModuleDescriptor
-import org.jetbrains.kotlin.incremental.KotlinLookupLocation
-import org.jetbrains.kotlin.name.ClassId
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.psi.KtAnnotated
 import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtCallableDeclaration
-import org.jetbrains.kotlin.psi.KtClass
-import org.jetbrains.kotlin.psi.KtClassBody
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
-import org.jetbrains.kotlin.psi.KtCollectionLiteralExpression
 import org.jetbrains.kotlin.psi.KtDotQualifiedExpression
 import org.jetbrains.kotlin.psi.KtFunction
 import org.jetbrains.kotlin.psi.KtFunctionType
 import org.jetbrains.kotlin.psi.KtNameReferenceExpression
 import org.jetbrains.kotlin.psi.KtNamedDeclaration
-import org.jetbrains.kotlin.psi.KtNamedFunction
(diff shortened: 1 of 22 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt
```diff
@@ -74,7 +79,28 @@ public sealed class AnnotationArgumentReference {
...
+            .asString()
+
+          when (parameterFqName) {
+            Boolean::class.qualifiedName -> psiElement.text.toBooleanStrictOrNull()
+            Int::class.qualifiedName -> psiElement.text.toIntOrNull()
+            Long::class.qualifiedName -> psiElement.text.toLongOrNull()
+            Double::class.qualifiedName -> psiElement.text.toDoubleOrNull()
+            Byte::class.qualifiedName -> psiElement.text.toByteOrNull()
+            Short::class.qualifiedName -> psiElement.text.toShortOrNull()
+            Float::class.qualifiedName -> psiElement.text.toFloatOrNull()
```
vRallev, line 100: I think I cover every scenario now.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -491,11 +491,9 @@ class InterfaceMergerTest(
       }
       """
     ) {
-      assertThat(exitCode).isEqualTo(INTERNAL_ERROR)
-      // Position to the class.
+      assertThat(exitCode).isError()
       assertThat(messages).contains(
-        "org.jetbrains.kotlin.util.KotlinFrontEndException: " +
-          "Exception while analyzing expression at (8,18)"
+        "Couldn't find the classId for <root>. Are we stuck in a loop while resolving super types?"
```
vRallev, line 496: Note that these tests were always failing with obscure compiler errors. Due to the refactorings we can actually catch them now and throw better error messages.

compiler/src/main/java/com/squareup/anvil/compiler/ClassScanner.kt
```diff
@@ -18,13 +19,12 @@ internal class ClassScanner {
    * Returns a sequence of contributed classes from the dependency graph. Note that the result
    * includes inner classes already.
    */
-  // TODO: Update to return [ClassReference]s
```
vRallev, line 21: @JoelWilcox I solved this too in the same PR, because I needed it.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributedBinding.kt
```diff
@@ -136,108 +120,33 @@ private fun ClassReference.checkNotGeneric(
...
-
-          val valueString = valueArgument.getArgumentExpression()?.valueString()
-
-          name + valueString
-        }
-
-      annotationFqNameString + argumentString
-    }
-
-  return boundTypeFqName.asString() + allArguments
```
vRallev, line 209: This code became so much simpler.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -275,37 +261,6 @@ internal class BindingModuleGenerator(
...
-        it.packageFqName.asString().startsWith(hintPackagePrefix)
-      }
-      .flatMap {
-        it.findChildrenByClass(KtProperty::class.java).toList()
-      }
-      .mapNotNull { ktProperty ->
-        // We can safely ignore scopes, we only care about the reference classes.
-        if (ktProperty.name?.endsWith(REFERENCE_SUFFIX) != true) return@mapNotNull null
-
-        // TODO: Figure out how we can replecate this logic for descriptors to fully convert this method to use ClassReference
```
vRallev, line 294: All this code wasn't actually needed. The new solution is so simple:
```kotlin
    contributedBindingClasses += classes.filter {
      it.isAnnotatedWith(contributesBindingFqName)
    }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -461,7 +320,7 @@ public fun PsiElement.requireFqName(
     ?.let { return it }
 
   findFqNameInSuperTypes(module, classReference)
-    ?.let { return it }
+    ?.let { return it.fqName }
```
JoelWilcox, line 323: Nit: naming feels a little off here now because it reads as `findFqName().fqName` where we're passing in something called `classReference` which seems like it would have the `FqName`. The following would feel better:

```
findClassReferenceInSuperTypes(module, className)
  ?.let { return it.fqName }
```

JoelWilcox (approved):
Looks great! :chef_kiss:

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
vRallev: Good catch. Updated

PR #534, opened 2022-02-24
Create _ReferenceIr classes and begin migrating ModulerMergerIr

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -0,0 +1,31 @@
...
+  val scope: ClassReferenceIr by lazy(NONE) {
+    scopeOrNull ?: throw AnvilCompilationException(
+      element = annotation,
+      message = "Couldn't find scope for $fqName."
+    )
+  }
+
+  val scopeOrNull: ClassReferenceIr? by lazy(NONE) {
+    context.referenceClass(annotation.scope())?.toClassReference(context)
+  }
```
vRallev, line 27: Same here for `toString()`, `hashCode()` and `equals()` (hashCode and equals might not work yet without parsing annotation arguments).

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -87,44 +90,50 @@ internal class ModuleMergerIr(
...
+        !it.fqName.isAnvilModule() || it.fqName == anvilModuleName
       }
       .mapNotNull {
         val contributesAnnotation =
-          it.owner.annotationOrNull(contributesToFqName, scope = scope)
+          it.annotations.find(annotationName = contributesToFqName, scopeName = scope)
+            .singleOrNull()
             ?: return@mapNotNull null
+        // TODO: change this to just map to annotations once [AnnotationReferenceIr] provides
+        //  its declaring class
```
vRallev, line 101: Ah, that answers my other comment :-)

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -0,0 +1,31 @@
...
+}
+
+internal fun IrConstructorCall.toAnnotationReference(context: IrPluginContext) =
```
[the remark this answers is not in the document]
JoelWilcox: Hmm I currently need context for finding the scope, using `IrPluginContext#referenceClass()`. Do you mean that the declaring class could be utilized for finding the scope?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -0,0 +1,31 @@
...
+import org.jetbrains.kotlin.ir.expressions.IrConstructorCall
+import org.jetbrains.kotlin.ir.util.parentAsClass
+import org.jetbrains.kotlin.name.FqName
+import kotlin.LazyThreadSafetyMode.NONE
+
+internal class AnnotationReferenceIr(
+  val annotation: IrConstructorCall,
+  val context: IrPluginContext
+) {
+  val fqName: FqName = annotation.symbol.owner.parentAsClass.fqName
```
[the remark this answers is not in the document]
JoelWilcox: Just to clarify, do you mean don't include any `fqName` field in this class? I saw that we still maintain the `fqName` field in `AnnotationReference` even though it's just delegating to the `ClassReference`

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/ClassReferenceIr.kt
```diff
@@ -0,0 +1,57 @@
...
+      else -> throw AnvilCompilationExceptionClassReferenceIr(
+        this,
+        "Encountered an unsupported visibility ${clazz.owner.visibility.name} for class $fqName"
+      )
+    }
+  }
+
+  override val annotations: List<AnnotationReferenceIr> by lazy(NONE) {
+    clazz.owner.annotations.map { it.toAnnotationReference(context) }
+  }
```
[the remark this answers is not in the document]
JoelWilcox: :+1: I added these in my next branch after realizing I forgot them :D, I'll add them here.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/ClassReferenceIr.kt
```diff
@@ -0,0 +1,74 @@
...
+import org.jetbrains.kotlin.ir.symbols.IrClassSymbol
+import org.jetbrains.kotlin.ir.util.isInterface
+import org.jetbrains.kotlin.name.FqName
+import kotlin.LazyThreadSafetyMode.NONE
+
+internal class ClassReferenceIr(
+  val clazz: IrClassSymbol,
+  val context: IrPluginContext
+) : AnnotatedReferenceIr {
+  val fqName: FqName by lazy { clazz.fqName }
```
vRallev, line 21: It's not necessary to make this lazy, because it's one of more commonly accessed properties. Especially not with double-checked locking.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/ClassReferenceIr.kt
```diff
@@ -0,0 +1,74 @@
...
+      )
+    }
+  }
+
+  override val annotations: List<AnnotationReferenceIr> by lazy(NONE) {
+    clazz.owner.annotations.map { it.toAnnotationReference(context, this) }
+  }
+
+  override fun toString(): String {
+    return "${this::class.qualifiedName}($fqName)"
```
vRallev, line 45: We used the qualified name for `ClassReference`, because there are sub-classes. Here it's really not needed and you can just use "ClassReferenceIr".

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -0,0 +1,80 @@
...
+  val classReference: ClassReferenceIr,
+  private val declaringClass: ClassReferenceIr?
+) {
+  val fqName: FqName
+    get() = classReference.fqName
+
+  private val context: IrPluginContext
+    get() = classReference.context
+
+  val scope: ClassReferenceIr by lazy(NONE) {
```
vRallev, line 23: Nit: this doesn't need to be lazy. A getter is enough.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -0,0 +1,80 @@
...
+    scopeOrNull ?: throw AnvilCompilationException(
+      element = annotation,
+      message = "Couldn't find scope for $fqName."
+    )
+  }
+
+  val scopeOrNull: ClassReferenceIr? by lazy(NONE) {
```
vRallev, line 30: Curious why you've chosen a property here, but not for declaringClass below. In the original implementation we use functions and hide the properties. Not a big deal, just curious :)

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotatedReferenceIr.kt
```diff
@@ -0,0 +1,21 @@
+package com.squareup.anvil.compiler.codegen.reference
+
+import org.jetbrains.kotlin.ir.util.isAnnotationWithEqualFqName
+import org.jetbrains.kotlin.name.FqName
+
+internal interface AnnotatedReferenceIr {
+  val annotations: List<AnnotationReferenceIr>
+
+  fun isAnnotatedWith(fqName: FqName): Boolean =
+    annotations.any { it.annotation.isAnnotationWithEqualFqName(fqName) }
```
vRallev, line 10:
```suggestion
    annotations.any { it.fqName == fqName }
```
That should be enough and is faster.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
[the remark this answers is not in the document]
vRallev: Just, that's what I meant. You can grab the context from there. But it doesn't work, if the class is nullable. So ignore that.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/ClassReferenceIr.kt
JoelWilcox: Switched to use `simpleName` instead so we still don't need to update it if the name ever changes

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
JoelWilcox: In retrospect, lazy also feels weird on this one compared to the nullable one

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
JoelWilcox: Good question. I usually prefer properties for things like this where there aren't any inputs and it's not something you'd expect to change over the lifetime of the class. I defaulted to functions for declaringClass because of name conflicts but I think I can actually just adjust it to not be a property constructor and make the class more consistent.

PR #536, opened 2022-02-24
Introduce `TypeReference` and provide an implementation for Psi and Descriptors

vRallev:


compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/FunctionReferenceTest.kt
```diff
@@ -0,0 +1,294 @@
...
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.compiler.compile
+import com.squareup.anvil.compiler.internal.fqName
+import com.squareup.anvil.compiler.internal.testing.simpleCodeGenerator
+import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.OK
+import org.jetbrains.kotlin.name.FqName
+import org.junit.Test
+
+class FunctionReferenceTest {
```
vRallev, line 11: I moved the existing utilities into a more appropriate class.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -48,7 +48,6 @@ import org.jetbrains.kotlin.resolve.scopes.getDescriptorsFiltered
 import org.jetbrains.kotlin.types.DefinitelyNotNullType
 import org.jetbrains.kotlin.types.FlexibleType
 import org.jetbrains.kotlin.types.KotlinType
-import org.jetbrains.kotlin.utils.addToStdlib.cast
```
vRallev, line 51: This made me a little uneasy and could break any time.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -42,33 +40,27 @@ public sealed class FunctionReference : AnnotatedReference {
...
     )
 
-  protected abstract fun resolveGenericReturnTypeOrNullPsi(
-    implementingClass: ClassReference.Psi
-  ): ClassReference?
+  public abstract fun isAbstract(): Boolean
+  public abstract fun isConstructor(): Boolean
+  public abstract fun visibility(): Visibility
 
+  // TODO: move or remove?
```
vRallev, line 56: Updated in one of the next PRs.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -150,29 +146,6 @@ public sealed class FunctionReference : AnnotatedReference {
...
-      returnTypeOrNull()?.let { return it }
-
-      val typeReference = function.typeReference
-        ?.let { typeReference ->
-          implementingClass.resolveTypeReference(typeReference)
-        }
-        ?: function.typeReference
-
-      return typeReference?.fqNameOrNull(module)?.toClassReference(module)
-    }
```
vRallev, line 175: These were moved into `TypeReference`.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -96,17 +97,17 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
     return if (index == 0) null else classes[index - 1]
   }
 
-  public abstract fun innerClasses(): List<ClassReference>
+  public open fun innerClasses(): List<ClassReference> =
```
JoelWilcox, line 100: Is this worth caching / changing to a val so we don't have to filter more than once?

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -96,17 +97,17 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
...
    * @param parameterName The name of the parameter to be found, not including any variance modifiers.
    * @return The 0-based index of a declared generic type.
    */
   protected abstract fun indexOfTypeParameter(parameterName: String): Int
 
-  public open fun companionObjects(): List<ClassReference> {
-    return innerClasses().filter { it.isCompanion() && it.enclosingClass() == this }
-  }
+  public open fun companionObjects(): List<ClassReference> =
+    innerClassesAndObjects.filter { it.isCompanion() && it.enclosingClass() == this }
```
JoelWilcox, line 110: same here

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
vRallev: The filter is so cheap, it's not really worth it ¯\\_(ツ)_/¯ At some point we should profile Anvil and investigate where most time and memory is used. I don't think it'll be the list operations.

PR #537, opened 2022-02-24
Use `TypeReference` in the `ParameterReference` implementation.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ParameterReference.kt
```diff
@@ -91,23 +79,8 @@ public sealed class ParameterReference : AnnotatedReference {
...
-    override fun resolveGenericTypeOrNull(implementingClass: ClassReference.Psi): ClassReference? {
-      typeOrNull()?.let { return it }
-
-      val typeReference = parameter.typeReference
-        ?.let { typeReference ->
-          implementingClass.resolveTypeReference(typeReference)
-        }
-        ?: parameter.typeReference
-
-      return typeReference?.fqNameOrNull(module)?.toClassReference(module)
```
vRallev, line 110: This code was pretty much duplicated between `ParameterReference` and `FunctionReference`. Now that the logic lives in `ParameterReference` it can be reused.

JoelWilcox (approved):

PR #538, opened 2022-02-24
Add an API to get the `TypeName` for a `TypeReference`.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/KotlinPoetUtils.kt
```diff
@@ -229,6 +233,10 @@ public fun KotlinType.asTypeName(): TypeName {
  *                      resolve type arguments.
  */
 @ExperimentalAnvilApi
+@Suppress("DeprecatedCallableAddReplaceWith")
+@Deprecated(
+  "Don't rely on descriptors and make the code agnostic to the underlying implementation."
+)
 public fun KotlinType.asTypeNameOrNull(
```
JoelWilcox, line 240: I noticed this one mentions descriptors but the deprecation message for `KotlinType.asTypeName()` mentions PSI. Should one be swapped to match?

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -52,12 +70,38 @@ public sealed class TypeReference {
         ?.toClassReference(module)
     }
 
+    override val typeName: TypeName? by lazy(NONE) {
+      try {
+        type.requireTypeName(module)
+      } catch (e: AnvilCompilationException) {
+        // The goal is to inline the function above and then stop throwing an exception.
```
JoelWilcox, line 77: May be worth adding a TODO here just so it's obvious this isn't final when scanning the code

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -52,12 +70,38 @@ public sealed class TypeReference {
...
+        typeReference.requireTypeName(module)
+          .let { typeName ->
+            // This is a special case when mixing parsing between descriptors and PSI.  Descriptors
+            // will always represent lambda types as kotlin.Function* types, so lambda arguments
+            // must be converted or their signatures won't match.
+            // see https://github.com/square/anvil/issues/400
+            (typeName as? LambdaTypeName)?.asFunctionType() ?: typeName
+          }
+      } catch (e: AnvilCompilationException) {
+        // The goal is to inline the function above and then stop throwing an exception.
```
JoelWilcox, line 101: Same here for adding TODO

JoelWilcox (approved):
All super minor comments

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: I removed one of these methods already in the next PR. I'll add a TODO to the remaining method in the other stacked PR.

PR #539, opened 2022-02-24
Move the utilities for resolving a generic type into the `TypeReference` class and clean them up

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -221,164 +212,7 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
...
-     * The KtTypeReference `T` will fail to resolve, since it isn't a type. This function will
-     * instead look to the `ServiceFactory` interface, then look at the supertype declaration
-     * in order to determine the type.
-     *
-     * @receiver The class which actually references the type. In the above example, this would
-     * be `ServiceFactory`.
-     * @param typeReference The generic reference to be resolved. In the above example, this
-     * is `T`.
-     */
-    public fun resolveTypeReference(typeReference: KtTypeReference): KtTypeReference? {
```
vRallev, line 300: That these functions are gone from this class makes so much more sense.

JoelWilcox: > That these functions are gone from this class makes so much more sense.

This is super nice to see.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -168,3 +231,94 @@ public fun AnvilCompilationExceptionTypReference(
...
+    val resolvedDeclaringClass = resolvedTypeReference.containingClass()
+      ?.toClassReference(implementingClass.module)
+      ?: return null
+
+    resolvedTypeReference
+      .toTypeReference(resolvedDeclaringClass)
+      .resolveTypeReference(implementingClass)
+  } else {
+    null
+  }
```
JoelWilcox, line 270: I think you can simplify this as 

```suggestion
  return resolvedTypeReference
      ?.containingClass()
      ?.toClassReference(implementingClass.module)
      ?.let { resolvedDeclaringClass ->
          resolvedTypeReference
            .toTypeReference(resolvedDeclaringClass)
            .resolveTypeReference(implementingClass)
      }
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -65,42 +82,55 @@ public sealed class TypeReference {
...
+     * interface ServiceFactory : Factory<Service>
+     * ```
+     *
+     * The KtTypeReference `T` will fail to resolve, since it isn't a type. This function will
+     * instead look to the `ServiceFactory` interface, then look at the supertype declaration
+     * in order to determine the type.
+     *
+     * @param implementingClass The class which actually references the type. In the above
+     * example, this would be `ServiceFactory`.
+     */
```
JoelWilcox, line 125: Great comment!

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: Yes!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: Thank @RBusarow

PR #542, opened 2022-02-25
Migrate `excludes` logic to use `AnnotationReferenceIr`, fix a scope IR symbol lookup issue

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/IrUtils.kt
```diff
@@ -105,9 +108,23 @@ internal fun IrConstructorCall.parentScope(): FqName {
...
+  // [FqName] being constructed handles nested classes if it's later used for finding IR symbols.
+  //
+  // e.g. we might have a signature like
+  // 'com.squareup.anvil.test/ContributedSubcomponent.ParentScope' which gets broken up as
+  // packageFqName     -> 'com.squareup.anvil.test'
+  // declarationFqName -> 'ContributedSubcomponent.ParentScope'
+  // shortName         -> 'ParentScope'
+  //
+  // Using [shortName] would produce the FqName 'com.squareup.anvil.test.ParentScope' which
+  // prevents the IR lookup utils from finding the matching symbol.
```
JoelWilcox, line 125: @vRallev do you think this comment is useful? It took a long time to track down the issue that I ran into related to this, although I'm not sure how likely it would be for someone to see this code and want to change it to use `shortName` later on 🤔

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -37,6 +40,17 @@ internal class AnnotationReferenceIr(
       message = "The declaring class was null, this means the annotation wasn't used on a class."
     )
 
+  val parentScope: ClassReferenceIr
+    get() = context.referenceClass(annotation.parentScope())?.toClassReference(context)
```
vRallev, line 44: Regarding our Slack conversation: I think that's worth putting into a lazy.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -37,6 +40,17 @@ internal class AnnotationReferenceIr(
...
+      ?: throw AnvilCompilationException(
+        element = annotation,
+        message = "Couldn't find parent scope for $fqName."
+      )
+
+  val excludedClasses: List<ClassReferenceIr> by lazy(NONE) {
+    annotation.exclude().map { it.symbol.toClassReference(context) }
```
vRallev, line 51: Do you plan to introduce `AnnotationArgumentIr`? That brought us the speed improvements for the descriptor implementation.

compiler/src/main/java/com/squareup/anvil/compiler/IrUtils.kt
vRallev: I think it's useful! The APIs for the compiler are quite consistent and `shortName` always refers to THIS class name. But a comment like this certainly doesn't hurt!

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -202,7 +214,7 @@ internal class ModuleMergerIr(
         )
         .flatMap { classSymbol ->
           val annotation = classSymbol.owner.annotation(annotationFqName)
-          if (scope == annotation.scope()) {
+          if (scope.fqName == annotation.scope()) {
```
vRallev, line 217: You should compare `ClassReferenceIr` instances and not FqName.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -297,7 +314,12 @@ internal class ModuleMergerIr(
         }
       }
 
-    declaration.annotations += annotationConstructorCall
+    declaration.addAnnotation(
+      annotationConstructorCall.toAnnotationReference(
+        pluginContext,
+        declaration
+      )
+    )
```
vRallev, line 322: Hm, interesting. Not sure if this is actually needed. Our `ClassReference` instances capture the current state of the compilation, we run our logic and then modify the existing code. After that we're done. We don't process or work with the modified code. I don't think you need to update class reference and you can just change the underlying IR structure.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
JoelWilcox: I believe so, or at least I added a TODO for it in the follow-up PR here: https://github.com/square/anvil/pull/544

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
JoelWilcox: This should be updated in #544. I just updated usages for `fqName` here to reduce the scope of this PR :).

vRallev: Sounds good 👍🏻

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
vRallev: Do you plan to address this here?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
vRallev: Do you plan to address this here?

JoelWilcox: Yes, will push changes shortly

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
JoelWilcox: Giving it a shot now and will address here if it proves unnecessary

vRallev (approved):

PR #543, opened 2022-02-25
Use a unified implementation for converting a `PropertyReference` to a `MemberInjectParameter`.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -146,6 +172,13 @@ public sealed class TypeReference {
       type.asTypeNameOrNull()?.lambdaFix()
     }
 
+    override val unwrappedFirstType: Descriptor by lazy(NONE) {
+      type.arguments
+        .first()
+        .type
+        .toTypeReference(declaringClass)
```
JoelWilcox, line 179: this is so much simpler than PSI :D

JoelWilcox (approved):

PR #544, opened 2022-02-25
Migrate the remainder of `ModuleMergerIr` to use Reference classes

JoelWilcox:
There are a couple spots with TODOs added but this completes the migration for the most part.


vRallev (approved):
This is great! All TODOs make sense to me to look into.

PR #545, opened 2022-02-25
Introduce `TypeParameterReference` and clean up the Dagger utilities. 

vRallev:


compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/ClassReferenceTest.kt
```diff
@@ -76,6 +77,85 @@ class ClassReferenceTest {
...
+      )
+      
+      class SomeClass2<V>(
+        private val v: V
+      ) where V : Appendable, V : CharSequence
+
+      class SomeClass3<T, R : Set<String>>(
+        private val t: T,
+        private val r: Lazy<R>
+      ) where T : Appendable, T : CharSequence
```
JoelWilcox, line 97: Didn't know that you could do this 👀

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/ClassReferenceTest.kt
```diff
@@ -76,6 +77,85 @@ class ClassReferenceTest {
...
+              assertThat(psiParameter.upperBounds[1].asClassReference().fqName)
+                .isEqualTo(FqName("kotlin.CharSequence"))
+
+              assertThat(descriptorParameter.upperBounds[0].asClassReference().fqName)
+                .isEqualTo(FqName("java.lang.Appendable"))
+              assertThat(descriptorParameter.upperBounds[1].asClassReference().fqName)
+                .isEqualTo(FqName("kotlin.CharSequence"))
+            }
+            "SomeClass3" -> {
+              listOf(psiRef, descriptorRef).forEach { ref ->
```
JoelWilcox, line 135: I like how you used the list to avoid duplicating the asserts between each ref here. Any reason you decided not to do the same for the other test cases?

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/ClassReferenceTest.kt
```diff
@@ -76,6 +77,85 @@ class ClassReferenceTest {
       assertThat(exitCode).isEqualTo(OK)
     }
   }
+
+  @Test fun `the type parameter list is parsed`() {
```
JoelWilcox, line 81: Great test 👍 , this helped make the usage of `upperBounds` really clear.

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/ClassReferenceTest.kt
vRallev: It's painful for PSI parsing...

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/ClassReferenceTest.kt
vRallev: Laziness 🙃  I should have done this. In general I don't like loops, because if the test fails then you don't know what the arguments are for the current iteration. But I got tired of duplicating all code, so I started using the loop.

PR #546, opened 2022-02-25
Clean up utilities

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
```diff
@@ -250,3 +251,13 @@ private fun PropertyReference.toMemberInjectParameter(
     injectedFieldSignature = fqName
   )
 }
+
+internal fun ClassName.optionallyParameterizedBy(
```
vRallev, line 255: BTW, all TODOs in this file are now resolved.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/DescriptorUtils.kt
```diff
@@ -28,7 +28,7 @@ public fun KotlinType.classDescriptorOrNull(): ClassDescriptor? {
 public fun KotlinType.classDescriptor(): ClassDescriptor {
```
vRallev, line 28: Notice the new file name. Before it was `CompilerUtils`, now it's `DescriptorUtils`.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/FqName.kt
```diff
@@ -38,3 +40,39 @@ internal val mergeModulesFqName = MergeModules::class.fqName
...
+  if (isRoot) {
+    ""
+  } else {
+    val prefix = if (dotPrefix) "." else ""
+    val suffix = if (dotSuffix) "." else ""
+    "$prefix$this$suffix"
+  }
+
+@ExperimentalAnvilApi
+public fun FqName.classIdBestGuess(): ClassId {
```
vRallev, line 68: I moved all `FqName` related utils in this file.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -101,9 +120,11 @@ public sealed class TypeReference {
 
     override val typeName: TypeName? by lazy(NONE) {
       try {
-        type.requireTypeName(module).lambdaFix()
+        type.requireTypeName().lambdaFix()
       } catch (e: AnvilCompilationException) {
-        // TODO: The goal is to inline the function above and then stop throwing an exception.
+        // Usually the method should return a nullable value, but throwing an error gives us
+        // better hints which element exactly fails to be converted to a TypeName while debugging.
+        // So it's better to keep that mechanism in place and catch the exception to move on.
```
vRallev, line 127: I removed the TODO and I hope the explanation makes sense why we keep catching the exception.

JoelWilcox: > I removed the TODO and I hope the explanation makes sense why we keep catching the exception.

Would it make sense to introduce a separate `typeNameOrNull()` method that doesn't throw the exception, similar to how we've structured a lot of other methods? Then spots like this could just call the `OrNull()` variant. 

Or if the comment is saying that even in this case we still want the the exception logged, perhaps we could still have that alternate `OrNull()` method but have it create and log the exception without throwing it to get the best of both worlds.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -158,6 +179,109 @@ public sealed class TypeReference {
...
+      }
+
+      fun KtUserType.findExtendsBound(): List<FqName> {
+        return parents.filterIsInstance<KtClassOrObject>()
+          .first()
+          .typeParameters
+          .mapNotNull { it.fqName }
+      }
+
+      fun KtTypeElement.requireTypeName(): TypeName {
```
JoelWilcox, line 205: I see that this is copy-pasta so I'm not suggesting you change anything here, but I feel like a lot of the longer methods like this are a bit difficult to follow, especially with all the different levels of nesting and could really benefit from breaking it up into more helper methods. Maybe that's something we could look at doing later.

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: I started refactoring this method and returning null instead. The problem there is that if the result is null, you don't know which element failed to parse. There are complex annotations like these `@BindingKey("abc", Unit::class, [1, 2, 3])`.

By throwing an exception, we can explicitly tag the Psi element, which was causing problems, instead of just having a null value. There are two methods:
```
  public fun asTypeNameOrNull(): TypeName? = typeName
  public fun asTypeName(): TypeName = typeName
    ?: throw AnvilCompilationExceptionTypReference(
      typeReference = this,
      message = "Unable to convert a type reference to a type reference."
    )
```
It's a little awkward right now, because we're throwing a different exception. I'll clean this up.

PR #547, opened 2022-02-25
Use our reference APIs to convert an annotation reference to an `AnnotationSpec`. 

vRallev:
Stop using descriptor / Psi specific implementations


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt
```diff
@@ -73,43 +123,117 @@ public sealed class AnnotationArgumentReference {
...
+              ?.let { return parsePrimitiveType(it) }
+
+            // If we reach here, it's a top-level constant defined in the same package but
+            // a different file. In this case, we can just do the same in our generated code
+            // because we're in the same package and can play by the same rules.
+            return psiElement.containingKtFile.packageFqName
+              .descendant(psiElement.getReferencedName())
+          }
+          is KtDotQualifiedExpression -> psiElement.requireFqName(module)
+          else -> fail()
```
vRallev, line 232: I had to heavily extend the parsing of annotation arguments. Psi is quite challenging for that. But I'm happy about the result. Descriptors are so much easier.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt
```diff
@@ -165,33 +214,6 @@ public sealed class AnnotationReference {
...
-                  simpleName = value.enumEntryName.asString()
-                )
-                addMember("${name.asString()} = %M", enumMember)
-              }
-              // String, int, long, ... other primitives.
-              else -> addMember("${name.asString()} = $value")
-            }
-          }
-        }
-        .build()
```
vRallev, line 193: Note that we have now one unified implementation. Neither of these two psi / descriptor specific implementations were complete. The good test coverage we have really stopped me from introducing regressions.

JoelWilcox (approved):

PR #548, opened 2022-02-25
Forward the better error in case a Psi element cannot be converted to a `TypeName`.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -292,10 +292,16 @@ public sealed class TypeReference {
...
 
-    override val typeName: TypeName? by lazy(NONE) {
+    override val typeNameOrNull: TypeName? by lazy(NONE) {
       type.asTypeNameOrNull()?.lambdaFix()
     }
 
+    override val typeName: TypeName
+      get() = typeNameOrNull ?: throw AnvilCompilationExceptionTypReference(
+        typeReference = this,
+        message = "Unable to convert a type reference to a type reference."
```
JoelWilcox, line 302:
```suggestion
        message = "Unable to convert a type reference to a type name"
```

Might also be worth including the `declaringClass.fqName` to help with debugging when this happens

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: Done

PR #549, opened 2022-02-26
Create `AnnotationArgumentReferenceIr` and clean up `IrUtils`

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationArgumentReferenceIr.kt
```diff
@@ -0,0 +1,58 @@
...
+import com.squareup.anvil.compiler.kclassUnwrapped
+import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
+import org.jetbrains.kotlin.ir.declarations.IrClass
+import org.jetbrains.kotlin.ir.declarations.IrValueParameter
+import org.jetbrains.kotlin.ir.expressions.IrExpression
+import org.jetbrains.kotlin.ir.expressions.IrVararg
+import kotlin.LazyThreadSafetyMode.NONE
+
+internal class AnnotationArgumentReferenceIr(
+  val argument: Pair<IrValueParameter, IrExpression>,
```
vRallev, line 12: Using a Pair feels odd here. I'd use two separate constructor arguments.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -44,35 +44,39 @@ internal class AnnotationReferenceIr(
...
   val replacedClasses: List<ClassReferenceIr> by lazy(NONE) {
-    argumentClassArray("replaces")
+    argument("replaces")?.value<List<ClassReferenceIr>>().orEmpty()
   }
 
-  // TODO: May be worth introducing AnnotationArgumentIr here
-  fun argumentClassArray(name: String): List<ClassReferenceIr> {
-    return annotation.argumentClassArray(name)
-  }
+  private fun argumentFqName(name: String): FqName {
```
vRallev, line 62: I'm a little confused why this function is needed. It should be a call similar to `excludedClasses`:
```kotlin
  val scopeOrNull: ClassReferenceIr? by lazy(NONE) {
    argument("scope")?.value<ClassReferenceIr>()
  }
```

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -291,9 +292,9 @@ internal class ModuleMergerIr(
         )
 
         fun copyArrayValue(name: String) {
-          // TODO This could also be cleaned up with an AnnotationArgumentIr type
-          declaration.clazz.owner.annotation(annotationFqName, scope.fqName)
+          declaration.annotations.find(annotationFqName, scope.fqName).single()
             .argument(name)
+            ?.argument
             ?.second
```
vRallev, line 298: Yeah, the Pair is weird here. I really suggest using two arguments in the constructor.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
JoelWilcox: Good catch, I didn't connect that this could just be replaced now instead of moved+refactored from the utils.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/reference/AnnotationReferenceIr.kt
```diff
@@ -44,35 +44,24 @@ internal class AnnotationReferenceIr(
...
-  private fun IrConstructorCall.argumentClassArray(
-    name: String
-  ): List<ClassReferenceIr> {
-    val vararg = argument(name)?.second as? IrVararg ?: return emptyList()
-
-    return vararg.elements
-      .filterIsInstance<IrExpression>()
-      .map { it.kclassUnwrapped.owner as IrClass }
-      .map { it.symbol.toClassReference(context) }
+  fun argument(name: String): AnnotationArgumentReferenceIr? {
```
vRallev, line 63: For consistency it might be good to name this method `argumentOrNull`.

vRallev (approved):
Nice clean up!

PR #551, opened 2022-03-01
Make `@ContributesTo` repeatable.

vRallev:
While you now can have multiple `@ContributesTo` annotations for Dagger modules and component interfaces, they all must have a different scope (contributing to the same scope twice is redundant).

### DIFF
--- a/compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
+++ b/compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
@@ -37,9 +37,15 @@ public class AnvilCompilation internal constructor(
     disableComponentMerging: Boolean = false,
     enableExperimentalAnvilApis: Boolean = true,
     codeGenerators: List<CodeGenerator> = emptyList(),
+    enableAnvil: Boolean = true,
   ) = apply {
     checkNotCompiled()
+    check(!anvilConfigured) { "Anvil should not be configured twice." }
+
     anvilConfigured = true
+
+    if (!enableAnvil) return@apply
+
     kotlinCompilation.apply {
       compilerPlugins = listOf(
         AnvilComponentRegistrar().also { it.addCodeGenerators(codeGenerators) }
--- a/compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
@@ -1,5 +1,6 @@
 package com.squareup.anvil.compiler
 
+import com.squareup.anvil.compiler.api.AnvilCompilationException
 import com.squareup.anvil.compiler.codegen.atLeastOneAnnotation
 import com.squareup.anvil.compiler.codegen.find
 import com.squareup.anvil.compiler.codegen.generatedAnvilSubcomponent
@@ -43,29 +44,39 @@ internal class InterfaceMerger(
       return
     }
 
-    val mergeScope = mergeAnnotation.scope()
+    val mergeScope = try {
+      mergeAnnotation.scope()
+    } catch (e: AssertionError) {
+      // In some scenarios this error is thrown. Throw a new exception with a better explanation.
+      // Caused by: java.lang.AssertionError: Recursion detected in a lazy value under
+      // LockBasedStorageManager@420989af (TopDownAnalyzer for JVM)
+      throw AnvilCompilationException(
+        classDescriptor = thisDescriptor,
+        message = "It seems like you tried to contribute an inner class to its outer class. " +
+          "This is not supported and results in a compiler error.",
+        cause = e
+      )
+    }
 
     if (!mergeAnnotatedClass.isInterface()) {
       throw AnvilCompilationExceptionClassReference(
-        mergeAnnotatedClass,
-        "Dagger components must be interfaces."
+        classReference = mergeAnnotatedClass,
+        message = "Dagger components must be interfaces."
       )
     }
 
     val classes = classScanner
       .findContributedClasses(
         module = module,
-        packageName = HINT_CONTRIBUTES_PACKAGE_PREFIX,
         annotation = contributesToFqName,
         scope = mergeScope.fqName
       )
       .filter {
         it.isInterface() && it.annotations.find(daggerModuleFqName).singleOrNull() == null
       }
-      .mapNotNull {
+      .flatMap {
         it.annotations
           .find(annotationName = contributesToFqName, scopeName = mergeScope.fqName)
-          .singleOrNull()
       }
       .onEach { contributeAnnotation ->
         val contributedClass = contributeAnnotation.declaringClass()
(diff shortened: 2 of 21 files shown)


vRallev:
Notice the diff compared to the old PR: 
image

I added the same tests as the old PR, but there are now 550 lines of code changes less in the compiler plugin. That's significant. All the refactorings made it so much easier this time.

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -41,29 +42,39 @@ internal class InterfaceMerger(
       return
     }
 
-    val mergeScope = mergeAnnotation.scope()
+    val mergeScope = try {
+      mergeAnnotation.scope()
+    } catch (e: AssertionError) {
+      // In some scenarios this error is thrown. Throw a new exception with a better explanation.
+      // Caused by: java.lang.AssertionError: Recursion detected in a lazy value under
+      // LockBasedStorageManager@420989af (TopDownAnalyzer for JVM)
```
vRallev, line 50: I'd like to point out that a unit test covers this check.

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -131,32 +129,20 @@ internal class InterfaceMerger(
...
-        val contributesBindingAnnotation = excludedClass
-          .annotations.find(contributesBindingFqName).singleOrNull()
-        val contributesMultibindingAnnotation = excludedClass
-          .annotations.find(contributesMultibindingFqName).singleOrNull()
-        val contributesSubcomponentAnnotation = excludedClass
-          .annotations.find(contributesSubcomponentFqName).singleOrNull()
+        val scopes = excludedClass.annotations.find(contributesToFqName).map { it.scope() } +
+          excludedClass.annotations.find(contributesBindingFqName).map { it.scope() } +
+          excludedClass.annotations.find(contributesMultibindingFqName).map { it.scope() } +
+          excludedClass.annotations.find(contributesSubcomponentFqName).map { it.parentScope() }
```
vRallev, line 135: I'll make this pattern a little bit nicer in a follow-up PR by introducing `findAll`. I wanted to make the change separately to keep the size of this PR a little smaller.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -84,3 +83,30 @@ internal fun FqName.isAnvilModule(): Boolean {
 internal fun <T : ClassReference> ClassId.classReferenceOrNull(
   module: ModuleDescriptor
 ): T? = asSingleFqName().toClassReferenceOrNull(module) as T?
+
+/**
+ * Returns the single element matching the given [predicate], or `null` if element was not found.
+ * Unlike [singleOrNull] this method throws an exception if more than one element is found.
+ */
+internal inline fun <T> Iterable<T>.singleOrEmpty(predicate: (T) -> Boolean): T? {
```
vRallev, line 91: In the old PR @JoelWilcox suggested renaming this method to `singleOrNull`, but `singleOrNull` already exists in the standard library. I kept `singleOrEmpty` on purpose, because the iterable object must of a single or no element (must be empty).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -154,31 +151,22 @@ internal class BindingModuleGenerator(
...
-        .filter { clazz ->
-          clazz.annotations.single { it.fqName == contributesToFqName }
-            .scope()
-            .fqName == scope
-        }
-        // Ignore replaced bindings specified by excluded modules for this scope.
-        .filter { clazz -> clazz !in excludedNames }
-        .flatMap { clazz ->
-          clazz.annotations.single { it.fqName == contributesToFqName }.replaces()
-        }
```
vRallev, line 166: This logic was duplicated above. I moved it down.

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
vRallev: #552

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
```diff
@@ -91,6 +96,17 @@ public class AnvilCompilation internal constructor(
     }
   }
 
+  /**
+   * Allows you to enable or disable Anvil in the compilation. By default Anvil is enabled.
+   */
+  public fun enableAnvil(enable: Boolean) = apply {
+    checkNotCompiled()
+    if (!enable) {
+      check(!anvilConfigured) { "You cannot disable Anvil after it has been configured." }
```
0xMatthewGroves, line 105: I can't help but feel that "configuring" and "enabling" flows could be streamlined. I'm sure they're split for good reason, but it's not clear when would Anvil be configured yet not enabled? If configuration is dependent on it being enabled, why not just enable it instead of doing a check?

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -117,58 +115,43 @@ internal class ModuleMerger(
...
-          .atLeastOneAnnotation(contributesAnnotation.fqName).single()
-          .replaces()
+        contributesAnnotation.replaces()
           .onEach { classToReplace ->
             // Verify has @Module annotation. It doesn't make sense for a Dagger module to
             // replace a non-Dagger module.
-            if (classToReplace.annotations.find(daggerModuleFqName).singleOrNull() == null &&
-              classToReplace.annotations.find(contributesBindingFqName).singleOrNull() == null &&
-              classToReplace.annotations.find(contributesMultibindingFqName).singleOrNull() == null
+            if (!classToReplace.isAnnotatedWith(daggerModuleFqName) &&
```
0xMatthewGroves, line 147: Much nicer!

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -84,3 +83,30 @@ internal fun FqName.isAnvilModule(): Boolean {
...
+        "Collection contains more than one matching element."
+      )
+      single = element
+      found = true
+    }
+  }
+  return single
+}
+
+private val truePredicate: (Any?) -> Boolean = { true }
```
0xMatthewGroves, line 106: Why not just integrate this into singleOrEmpty? I don't see it used elsewhere.

integration-tests/library/src/main/java/com/squareup/anvil/test/RepeateableContributions.kt
```diff
@@ -0,0 +1,21 @@
+package com.squareup.anvil.test
```
0xMatthewGroves, line 1: Filename has a mis-spelling. Repeatable vs Repeateable.

0xMatthewGroves (approved):
Fantastic work. Excited for this to land.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -622,4 +623,221 @@ class InterfaceMergerTest(
...
+      """
+    ) {
+      assertThat(componentInterface extends contributingInterface).isFalse()
+      assertThat(componentInterface extends secondContributingInterface).isTrue()
+      assertThat(subcomponentInterface extends contributingInterface).isTrue()
+      assertThat(subcomponentInterface extends secondContributingInterface).isFalse()
+    }
+  }
+
+  @Test fun `contributed interfaces in the old format are picked up`() {
```
JoelWilcox, line 800: Couple questions on this one:
Could you clarify what the old format is here? (is it referring to the second file in the first compilation?
Why do we disable Anvil for the first compilation?

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -622,4 +623,221 @@ class InterfaceMergerTest(
...
+        """,
+        """
+        package anvil.hint.merge.com.squareup.test
+
+        import com.squareup.test.ContributingInterface
+        import kotlin.reflect.KClass
+        
+        public val com_squareup_test_ContributingInterface_reference: KClass<ContributingInterface> = ContributingInterface::class
+        
+        // Note that the number is missing after the scope.
```
JoelWilcox, line 822: I am also unfamiliar with what this part is referencing. Would we normally expect a number after the scope here?

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -117,58 +115,43 @@ internal class ModuleMerger(
...
-          ?: contributesSubcomponentAnnotation?.parentScope()
-          ?: throw AnvilCompilationExceptionClassReference(
-            message = "Could not determine the scope of the excluded class " +
-              "${excludedClass.fqName}.",
-            classReference = excludedClass
-          )
+        val contributesToOurScope = scopes.any { it == mergeScope }
 
-        if (scopeOfExclusion != mergeScope) {
+        if (!contributesToOurScope) {
```
JoelWilcox, line 126: I love how smooth of a transition this was for the Mergers and Generators 🎉

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
JoelWilcox: > I can't help but feel that "configuring" and "enabling" flows could be streamlined. I'm sure they're split for good reason, but it's not clear when would Anvil be configured yet not enabled? If configuration is dependent on it being enabled, why not just enable it instead of doing a check?

I agree. It would be helpful to understand the use-case for disabling Anvil / when would we disable it. Maybe this info could even be included in the doc comment here.

JoelWilcox (approved):
Looks great! Also really cool to see how much smaller the PR is compared to before the refactorings.

integration-tests/library/src/main/java/com/squareup/anvil/test/RepeateableContributions.kt
JoelWilcox: I had to read this a couple times to see the typo 😁 .

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
vRallev: To avoid allocating a lambda every single time.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
vRallev: In the old format we generated two properties `reference` and `scope` for each contribution. In the new format we generate `reference` and `scope1, scope2, scope3,...`. This test makes sure that we still find a scope without an index number. 

That's why we must disable Anvil in the first compilation. If it would be enabled, then we would generate hints in the new format. That's not what we want to test. That's why I disabled Anvil and wrote the code manually that Anvil would have generated before.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
vRallev: See the comment above. I should update the Anvil internals doc.

compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
vRallev: Oh yes, that's much better now. 

There is only one use case I see mentioned below :-)

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
vRallev: @JoelWilcox I've updated the doc in the same PR. Does this make sense?

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
JoelWilcox: Ahh that makes sense, thanks for explaining 👍

PR #555, opened 2022-03-02
Skip evaluating anonymous inner classes when merging modules and interfaces

JoelWilcox:


compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -622,4 +623,26 @@ class InterfaceMergerTest(
       assertThat(componentInterface extends secondContributingInterface).isTrue()
     }
   }
+
+  @Test fun `anonymous inner classes should not be evaluated`() {
```
JoelWilcox, line 627: This case came up initially from an `InterfaceMerger`-based stacktrace, although technically this test isn't specific to `InterfaceMerger`. Open to giving it a new home if you have a better suggestion @vRallev

compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMerger.kt
```diff
@@ -30,6 +31,10 @@ internal class InterfaceMerger(
     thisDescriptor: ClassDescriptor,
     supertypes: MutableList<KotlinType>
   ) {
+    // If we're evaluating an anonymous inner class, it cannot contribute anything and will cause
```
vRallev, line 34: > it cannot contribute anything.

I think this should be "it cannot merge anything".

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
vRallev: I think that's fine.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerTest.kt
```diff
@@ -622,4 +623,26 @@ class InterfaceMergerTest(
...
+    compile(
+      """
+      package com.squareup.test
+
+      class NonContributingClass {
+        private val anonymousDoSomething = object : DoSomethingInterface {
+          override fun doSomething() { 
+            // unimportant impl
+          }
```
vRallev, line 636:
```suggestion
          override fun doSomething() = Unit
```
We use that to indicate that the body is empty on purpose.

vRallev (approved):

PR #556, opened 2022-03-02
Make `@ContributesBinding` and `@ContributesMultibinding` repeatable.

vRallev:
The size of the PR looks scary. But guess what, that's only tests. There are minimal changes in the plugin itself (just +10 lines of code in fact).

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -190,7 +190,9 @@ internal class BindingModuleGenerator(
 
         val replacedBindings = allContributedClasses
           .flatMap { classReference ->
-            classReference.annotations.single { it.fqName == annotationFqName }.replaces()
+            classReference.annotations
+              .find(annotationName = annotationFqName, scope = scope)
+              .flatMap { it.replaces() }
           }
 
         return allContributedClasses
@@ -198,17 +200,11 @@ internal class BindingModuleGenerator(
           .filterNot { it in replacedBindings }
           .filterNot { it in bindingsReplacedInDaggerModules }
           .filterNot { it in excludedNames }
-          .filter { clazz ->
+          .flatMap { clazz ->
             clazz.annotations
               .find(annotationName = annotationFqName, scope = scope)
-              .isNotEmpty()
-          }
-          .map { clazz ->
-            clazz.annotations
-              .filter { it.fqName == annotationFqName }
               .map { it.toContributedBinding(isMultibinding) }
           }
-          .flatten()
           .let { bindings ->
             if (isMultibinding) {
               bindings
--- a/compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
@@ -1,14 +1,18 @@
 package com.squareup.anvil.compiler.codegen
 
 import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.compiler.assumeIrBackend
 import com.squareup.anvil.compiler.compile
 import com.squareup.anvil.compiler.contributingInterface
 import com.squareup.anvil.compiler.hintBinding
 import com.squareup.anvil.compiler.hintBindingScope
+import com.squareup.anvil.compiler.hintBindingScopes
 import com.squareup.anvil.compiler.isError
+import com.squareup.anvil.compiler.secondContributingInterface
 import org.junit.Test
 import java.io.File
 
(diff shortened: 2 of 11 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -190,25 +190,21 @@ internal class BindingModuleGenerator(
...
+              .flatMap { it.replaces() }
           }
 
         return allContributedClasses
           .asSequence()
           .filterNot { it in replacedBindings }
           .filterNot { it in bindingsReplacedInDaggerModules }
           .filterNot { it in excludedNames }
-          .filter { clazz ->
+          .flatMap { clazz ->
```
vRallev, line 203: This and the change above is literally the only important change. That was simple 😅

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
```diff
@@ -304,4 +308,125 @@ class ContributesBindingGeneratorTest {
...
+      @com.squareup.anvil.annotations.ContributesBinding(Unit::class)
+      class ContributingInterface : ParentInterface
+      """
+    ) {
+      assertThat(contributingInterface.hintBinding?.java).isEqualTo(contributingInterface)
+      assertThat(contributingInterface.hintBindingScopes).containsExactly(Any::class, Unit::class)
+    }
+  }
+
+  @Test fun `multiple contribution can have different bound types`() {
```
JoelWilcox, line 384:
```suggestion
  @Test fun `multiple contributions can have different bound types`() {
```

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
```diff
@@ -304,4 +308,125 @@ class ContributesBindingGeneratorTest {
...
+      interface ParentInterface1
+      interface ParentInterface2
+
+      @ContributesBinding(Any::class, boundType = ParentInterface1::class)
+      @ContributesBinding(Unit::class, boundType = ParentInterface2::class)
+      class ContributingInterface : ParentInterface1, ParentInterface2
+      """
+    ) {
+      assertThat(contributingInterface.hintBinding?.java).isEqualTo(contributingInterface)
+      assertThat(contributingInterface.hintBindingScopes).containsExactly(Any::class, Unit::class)
```
JoelWilcox, line 402: I feel like we need an additional (or different) assert here to verify the different boundTypes were maintained. It would be possible for this test to continue passing even if the compile source was changed to not specify any bound types.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGeneratorTest.kt
```diff
@@ -397,4 +401,128 @@ class ContributesMultibindingGeneratorTest {
...
+        .containsExactly(Any::class, Unit::class)
+        .inOrder()
+
+      assertThat(secondContributingInterface.hintMultibindingScopes)
+        .containsExactly(Any::class, Unit::class)
+        .inOrder()
+    }
+  }
+
+  @Test fun `there are multiple hints for contributed bindings with fully qualified names`() {
```
JoelWilcox, line 457:
```suggestion
  @Test fun `there are multiple hints for contributed multibindings with fully qualified names`() {
```

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGeneratorTest.kt
```diff
@@ -397,4 +401,128 @@ class ContributesMultibindingGeneratorTest {
...
+      interface ParentInterface2
+
+      @ContributesMultibinding(Any::class, boundType = ParentInterface1::class)
+      @ContributesMultibinding(Unit::class, boundType = ParentInterface2::class)
+      class ContributingInterface : ParentInterface1, ParentInterface2
+      """
+    ) {
+      assertThat(contributingInterface.hintMultibinding?.java).isEqualTo(contributingInterface)
+      assertThat(contributingInterface.hintMultibindingScopes)
+        .containsExactly(Any::class, Unit::class)
```
JoelWilcox, line 498: Same here (re: needing another assert)

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
vRallev: Hm, this test only generates the hint and not the actual binding method. I agree that it's confusing and moved the test into the `BindingModuleGeneratorTest` file.

PR #557, opened 2022-03-03
Fix detection of generic types for automatically adding `@JvmSuppressWildcards`

JoelWilcox:
Co-authored by @vRallev

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
@@ -95,7 +95,7 @@ public sealed class TypeReference {
       ?: asTypeNameOrNull()
   }
 
-  public fun isGenericType(): Boolean = asClassReferenceOrNull() == null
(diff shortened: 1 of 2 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -146,6 +146,12 @@ public sealed class TypeReference {
       return resolveTypeReference(implementingClass)
     }
 
+    override fun isGenericType(): Boolean {
```
vRallev, line 149: So the issue was the wrong implementation for `isGenericType`?

vRallev (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
JoelWilcox: Yes, this was returning false for the register use-case, as well as more basic types like Set<String>. The new PSI impl is basically copy-pasta from what we did before the ClassReference refactorings.

PR #558, opened 2022-03-03
Improve the detection of generic types.

vRallev:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -95,7 +95,7 @@ public sealed class TypeReference {
       ?: asTypeNameOrNull()
   }
 
-  public abstract fun isGenericType(): Boolean
+  public fun isGenericType(): Boolean = asClassReferenceOrNull()?.isGenericClass() ?: true
```
JoelWilcox, line 98: 😮  Forgot that was something available on the class. Nice.

JoelWilcox (approved):

Issue #422, opened 2021-11-10
Remove support for the old compiler backend


vRallev:
Kotlin 1.6 will deprecate the old compiler backend. The IR compiler backend for JVM/Android projects became the default with Kotlin 1.5. At this point there should be no reason why someone on those platforms still uses the old backend.

Anvil still supports both. But supporting the old backend will become a blocker for #236. Repeated annotations with a runtime / compile time retention are only implemented in the IR backend and not in the old backend. Anvil will remove support for the old backend to support repeated annotations.

My plan is to bump Kotlin 1.6 for Anvil and release a final version supporting both backends. This will unblock anyone still using the old backend and Kotlin 1.6. There is a binary incompatible change within the Kotlin compiler that makes Anvil versions compiled with 1.5 incompatible with 1.6. This last release built with 1.6 for both backends is therefore necessary.

After that support for the old backend will be removed. Newer Anvil versions will only support the IR backend.

vRallev:
I just learnt that the stub generating task is always using the old backend:
![image](https://user-images.githubusercontent.com/1558045/141246338-c4f6c16e-7363-402a-a947-714552b02ce0.png)

We probably won't be able to remove any code. But this doesn't seem to be a big problem. We only need to make sure that Anvil itself is compiled with the IR backend. Consumers can still use both.

For #236 this means if you use the old backend, then you can only apply the repeatable annotations once. The Kotlin compiler will fail otherwise:
>  Repeatable annotations with non-SOURCE retention are only supported starting from Kotlin 1.6

The old backend assumes language level 1.5. This restriction doesn't seem to exist for the stub generating task. 

If you use the IR backend, then you can use the repeatable annotations as you'd expect.

vRallev:
Update: there will be [a solution](https://youtrack.jetbrains.com/issue/KT-49682) in Kotlin 1.7.

SteinerOk:
@vRallev I apologize for the off-topic question, but I think it's better than creating a separate topic because of a trifle.
Anvil is compatible with Kotlin 1.6.10, can I update my project to this version?

vRallev:
@SteinerOk Please see the release notes: https://github.com/square/anvil/releases Use version `2.3.10-1-6-0`, it's also compatible with Kotlin 1.6.10.

vRallev:
[...]

PR #559, opened 2022-03-03
Make all merge annotations repeatable.

vRallev:


compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -321,50 +327,17 @@ internal class ModuleMerger(
   }
 }
 
-@Suppress("DataClassPrivateConstructor")
-private data class AnnotationHolder private constructor(
```
vRallev, line 325: These wrapper classes weren't that useful anymore and it was easier to work with `AnnotationReference` directly instead.

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMerger.kt
```diff
@@ -48,27 +45,29 @@ internal class ModuleMerger(
...
     val anvilModuleName = createAnvilModuleName(clazz)
 
-    val modules = classScanner
-      .findContributedClasses(
-        module = module,
-        annotation = contributesToFqName,
-        scope = mergeScope.fqName
-      )
+    val contributesAnnotations = scopes
+      .flatMap { scope ->
```
vRallev, line 63: This was the general approach. Instead of querying a single scope, I had to flatmap the collections for multiple scopes.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/MergeAnnotationsCheckGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+          .findAll(
+            mergeComponentFqName,
+            mergeSubcomponentFqName,
+            mergeModulesFqName,
+            mergeInterfacesFqName
+          )
+          .ifEmpty { return@forEach }
+
+        annotations.checkSingleAnnotation()
+        annotations.checkNoDuplicateScope(contributeAnnotation = false)
```
vRallev, line 39: I had to duplicate these checks between the `InterfacerMerger`, `ModuleMerger` and `ModuleMergerIr`. That was annoying. Instead, it was a lot easier to move these checks into the compiler frontend with its own code generator.

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerRepeatableTest.kt
```diff
@@ -0,0 +1,347 @@
...
+        if (isFullTestRun()) {
+          add(MergeSubcomponent::class)
+          add(MergeModules::class)
+        }
+      }
+    }
+  }
+
+  @Test fun `duplicate scopes are an error`() {
+    assumeIrBackend()
```
vRallev, line 39: It's quite unfortunate that we can't test the old backend. The old backend is still used for the stub generating task. I added some integration tests for that.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/MergeAnnotationsCheckGenerator.kt
```diff
@@ -0,0 +1,76 @@
...
+
+        // Note that we only allow a single type of `@Merge*` annotation through the check above.
+        // The same class can't merge a component and subcomponent at the same time. Therefore,
+        // all annotations must have the same FqName and we can use the first annotation to check
+        // for the Dagger annotation.
+        annotations
+          .firstOrNull { it.fqName != mergeInterfacesFqName }
```
JoelWilcox, line 46: Why do we check specifically for MergeInterfaces her? (compared to the other merge annotations that this generator applies to)

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerRepeatableTest.kt
```diff
@@ -0,0 +1,189 @@
...
+    ) {
+      assertThat(exitCode).isError()
+      assertThat(messages).contains(
+        "com.squareup.test.ComponentInterface merges multiple times to the same scope: [Any]. " +
+          "Merging multiple times to the same scope is forbidden and all scopes must be distinct."
+      )
+    }
+  }
+
+  @Test fun `different kind of merge annotations are forbidden`() {
```
JoelWilcox, line 58: Given that the class/suite is parameterized, will this test still run for each annotation even though we're not using the parameter/`annotationClass` field in this test?

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerRepeatableTest.kt
JoelWilcox: > It's quite unfortunate that we can't test the old backend. The old backend is still used for the stub generating task. I added some integration tests for that.

Could you explain this more? Would we not be using the old backend if we didn't have `assumeIrBackend()` at the start of this test?

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerRepeatableTest.kt
vRallev: Without this and using the old backend you always get this error:
```
Repeatable annotations with non-SOURCE retention are only supported starting from Kotlin 1.6
```
The old backend doesn't support language features from Koltin 1.6 and newer. But for some reason it still somehow works for the stub generating task and this check seems to be disabled. Jetbrains is working to remove the old backend for the stub generating task. See https://github.com/square/anvil/issues/422#issuecomment-989458608

compiler/src/main/java/com/squareup/anvil/compiler/codegen/MergeAnnotationsCheckGenerator.kt
vRallev: Because this check below makes sure that you don't use `@MergeComponent` and `@dagger.Component` at the same time. Anvil will generate the Dagger component. There's no Dagger annotation for `@MergeInterfaces`, so this check doesn't make sense.

compiler/src/test/java/com/squareup/anvil/compiler/InterfaceMergerRepeatableTest.kt
vRallev: Yes, it will. That's why the test uses `assumeTrue(annotationClass == MergeComponent::class)` to make it run only once. For all other annotations the test will be skipped.

PR #561, opened 2022-03-03
Ignore classes without a classId when merging interfaces and modules

JoelWilcox:
Co-authored by @vRallev


compiler/src/test/java/com/squareup/anvil/compiler/ModuleMergerTest.kt
```diff
@@ -1128,4 +1130,50 @@ class ModuleMergerTest(
...
+      package com.squareup.test
+      
+      import com.squareup.anvil.annotations.ContributesTo
+      import dagger.Module
+      import dagger.Provides
+      $import
+
+      class Utils {
+        companion object {
+            val EXTRA_UTILS: ExtraUtils
```
vRallev, line 1148: Super nit: the indent is slightly off by one space

vRallev (approved):

PR #577, opened 2022-03-24
Don't add the "Get" prefix to generated Factory classes, if the provider method name starts with "is"

### LINKED ISSUE #574: "is..."-named Module properties break build
A module property with a name starting with "is" that acts as a provider breaks Dagger processing when processing the module with Anvil 2.4.0-M2. This does not happen when using straight Dagger for annotation processing. I suppose there is some problem with different beans-like name processing being applied by the processors. At a minimum, this is super-confusing; I was chasing this problem for hours before thinking of the property name and finding the actual repro path.

For example:

    @Module
    @ContributesTo(MyComponentScope::class)
    class MyModule {
        @get:Provides
        val isValidCache: BooleanArray = booleanArrayOf(false)
    }

The element needs to be injected somewhere to trigger the error:

    > Task :app:compileDebugJavaWithJavac FAILED
    MyComponent.java:86: error: cannot find symbol
    import com.example.MyModule_IsValidCacheFactory;
    ^
    symbol:   class MyModule_IsValidCacheFactory
    location: package com.example
    app/build/generated/source/kapt/debug/com/example/DaggerMyComponent.java:800: error: cannot find symbol
    return (T) new SampleViewModel(MyModule_IsValidCacheFactory.isValidCache(myComponentImpl.myModule));

If I use a different property name like 'validCache', Dagger prefixes 'get' when generating the component code: 'getValidCache()' instead of 'isValidCache()'

vRallev:
Fixes #574

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -116,6 +116,8 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
     val isObject = isCompanionObject || clazz.isObject()
 
     val isProperty = declaration.isProperty
+    val declarationName = declaration.fqName.shortName().asString()
+    val useGetPrefix = isProperty && !declarationName.startsWith("is")
 
     val isMangled = !isProperty &&
       declaration.visibility == INTERNAL &&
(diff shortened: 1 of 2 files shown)


RBusarow (approved):
I think there are still edge cases for `@get:JvmName(...)`, `@JvmField`, and `@JvmStatic` if it's an object?

vRallev:
> I think there are still edge cases for `@get:JvmName(...)`, `@JvmField`, and `@JvmStatic` if it's an object?

Possible, but that's a story for another day 🙃

PR #580, opened 2022-03-28
fix provides method names with an 'is-' prefix which is part of a longer word

RBusarow:
The kotlin/java interop only omits "get-" if the `is` prefix is the _word_ "is", like `isAndroid` or `is_android`.

There's no special case for `isandroid`, `issues`, etc.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -24,6 +24,7 @@ import com.squareup.anvil.compiler.internal.reference.classAndInnerClassReferenc
 import com.squareup.anvil.compiler.internal.reference.generateClassName
 import com.squareup.anvil.compiler.internal.safePackageString
 import com.squareup.anvil.compiler.internal.withJvmSuppressWildcardsIfNeeded
+import com.squareup.anvil.compiler.isWordPrefixRegex
 import com.squareup.anvil.compiler.publishedApiFqName
 import com.squareup.kotlinpoet.ClassName
 import com.squareup.kotlinpoet.FileSpec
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -117,7 +117,9 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
 
     val isProperty = declaration.isProperty
     val declarationName = declaration.fqName.shortName().asString()
-    val useGetPrefix = isProperty && !declarationName.startsWith("is")
+    // omit the `get-` prefix for property names starting with the *word* `is`, like `isProperty`,
+    // but not for names which just start with those letters, like `issues`.
+    val useGetPrefix = isProperty && !"^is([^a-z_].*)".toRegex().matches(declarationName)
```
vRallev, line 122: Should the regex be a constant for perf reasons?

vRallev (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -3357,6 +3357,44 @@ public final class DaggerModule1_ProvideFunctionFactory implements Factory<Set<S
...
+      package com.squareup.test
+      
+      import dagger.Module
+      import dagger.Provides
+      import javax.inject.Singleton
+      
+      @Module
+      class DaggerModule1 {
+        @get:Provides
+        val issues: List<String> = listOf("a")
```
JoelWilcox, line 3373: Could you also add a test for the _ case? Like `is_valid_cache`. It looks like we don't currently have a test for that case in the preexisting tests

JoelWilcox (approved):
Good catch!

vRallev:
Merging on your behalf to cut a new release.

PR #590, opened 2022-04-04
Resolve generic types for both Psi and Descriptors

JoelWilcox:


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -421,37 +420,54 @@ public fun AnvilCompilationExceptionTypReference(
...
-
-  return resolvedTypeReference
-    ?.containingClass()
-    ?.toClassReference(implementingClass.module)
-    ?.let { resolvedDeclaringClass ->
-      resolvedTypeReference
-        .toTypeReference(resolvedDeclaringClass)
-        .resolveTypeReference(implementingClass)
-    }
+  if (implementingClass is ClassReference.Psi) {
```
vRallev, line 434: Please use a `when` here.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -421,37 +420,54 @@ public fun AnvilCompilationExceptionTypReference(
...
+    return resolvedTypeReference
+      ?.containingClass()
+      ?.toClassReference(implementingClass.module)
+      ?.let { resolvedDeclaringClass ->
+        resolvedTypeReference
+          .toTypeReference(resolvedDeclaringClass)
+          .resolveTypeReference(implementingClass)
+      }
+  } else {
+    implementingClass as ClassReference.Descriptor
```
vRallev, line 456: The `when` will also avoid this weird statement.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -76,14 +77,12 @@ public sealed class TypeReference {
...
-  // implementation, e.g. FunctionReference resolves the return type or ParameterReference.
-  internal abstract fun resolveGenericTypeOrNull(
-    implementingClass: ClassReference.Psi
+  public abstract fun resolveGenericTypeOrNull(
+    implementingClass: ClassReference
   ): TypeReference?
 
-  internal fun resolveGenericTypeNameOrNull(
-    implementingClass: ClassReference.Psi
+  public fun resolveGenericTypeNameOrNull(
```
vRallev, line 84: I think it would make sense to add tests for that in `TypeReferenceTest`

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -56,35 +56,9 @@ public sealed class FunctionReference : AnnotatedReference {
...
-          .singleOrNull { function ->
-            function.overriddenFunctions.any { it.fqNameSafe == fqName }
-          }
-
-        implementingFunction?.returnTypeOrNull()?.asClassReference()
-      }
-    }
+    return returnType
+      ?.resolveGenericTypeOrNull(implementingClass)
+      ?.asClassReferenceOrNull()
```
vRallev, line 61: 💥

vRallev (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
JoelWilcox: Good call, I was thinking this should be covered implicitly but it's probably worth having explicit coverage for this.

JoelWilcox: Tests are pushed now, let me know if it still looks good to you.

vRallev: Tests look great :shipit:

PR #591, opened 2022-04-04
Fix dagger factory member injection not handling generics.

### LINKED ISSUE #486: [dagger-factory] Member injection for super types fails when passing a generic type
Hey, I was trying out the member injection for supertype added in the latest release but I found that it fails when we are passing a generic type to the base class. Following is the sample to reproduce the issue -

```
package com.sample.anvil

import javax.inject.Inject

abstract class Base<S> {
    @Inject
    lateinit var strings: List<S>
}
```

```
package com.sample.anvil

import javax.inject.Inject

class InjectClass : Base<String>() {

    @Inject
    lateinit var numbers: List<Int>
}
```

The MemberInjector generated for InjectClass throws compilation error with the message  `Unresolved reference: S`. I also noticed the MemberInjector generated for Base class also has the same error, so this might not be related to any recent changes but I am not sure. Thanks in advance

JoelWilcox:
Fixes #486

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
@@ -128,23 +128,24 @@ public sealed class TypeReference {
     override val unwrappedTypes: List<Psi> by lazy(NONE) {
       type.typeElement!!.children
         .filterIsInstance<KtTypeArgumentList>()
-        .single()
-        .children
-        .filterIsInstance<KtTypeProjection>()
-        .map { typeProjection ->
+        .singleOrNull()
+        ?.children
+        ?.filterIsInstance<KtTypeProjection>()
+        ?.map { typeProjection ->
           typeProjection.children
             .filterIsInstance<KtTypeReference>()
             .single()
             .toTypeReference(declaringClass)
-        }
+        } ?: emptyList()
     }
 
+    @Suppress("RedundantNullableReturnType")
     override fun resolveGenericTypeOrNull(implementingClass: ClassReference): TypeReference? {
       return resolveTypeReference(implementingClass)
     }
 
     /**
-     * Safely resolves a PSI [KtTypeReference], when that type reference may be a generic
+     * Safely resolves a type reference, when that type reference may be a generic
      * expressed by a type variable name. This is done by inspecting the class hierarchy to
      * find where the generic type is declared, then resolving *that* reference.
      *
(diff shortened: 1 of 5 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -158,21 +158,32 @@ public sealed class TypeReference {
...
-    ): TypeReference? {
+    ): TypeReference {
       // If the element isn't a type variable name like `T`, it can be resolved through imports.
       type.typeElement?.fqNameOrNull(module)
         ?.let { return this }
 
-      return resolveGenericTypeReference(implementingClass, declaringClass, type.text)
+      // When we fail to resolve the generic type here, we're potentially at the end of a chain of
+      // generics, so we return the current type instead of null.
+      return resolveGenericTypeReference(implementingClass, declaringClass, type.text) ?: this
```
JoelWilcox, line 186: With this change, a number of `orNull` style methods in `TypeReference` can no longer return null. I wanted to see how you feel about this change before cleaning those up, Ralf. As far as I can tell, we don't currently rely on being able to get null returned when the generic can't be resolved. The cases that were using the `orNull` calls appear to then try and get the class reference which will end up null anyway.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
```diff
@@ -248,10 +270,27 @@ private fun PropertyReference.toMemberInjectParameter(
...
   )
 }
 
+internal fun TypeReference.isGenericExcludingTypeAliases(): Boolean {
+  // A TypeReference for 'typealias StringList = List<String> would still show up as generic but
+  // would have no unwrapped types available (String).
+  return isGenericType() && unwrappedTypes.isNotEmpty()
+}
+
+internal fun ClassName.optionallyParameterizedByNames(
```
vRallev, line 284: internal or private?

vRallev (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
JoelWilcox: Actually looks like a numer of the `orNull`s are still valid, so I only udpated the Psi one where it's always non-null now.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -128,23 +128,23 @@ public sealed class TypeReference {
...
           typeProjection.children
             .filterIsInstance<KtTypeReference>()
             .single()
             .toTypeReference(declaringClass)
-        }
+        } ?: emptyList()
     }
 
-    override fun resolveGenericTypeOrNull(implementingClass: ClassReference): TypeReference? {
+    override fun resolveGenericTypeOrNull(implementingClass: ClassReference): TypeReference {
```
vRallev, line 142: The new return type with this method name is kinda weird. I think I'd keep a nullable return type and just suppress the warning in the IDE.

vRallev (approved):

PR #610, opened 2022-06-09
create `PropertyReference.Psi` from primary constructor properties

### LINKED ISSUE #609: primary constructor properties are not included in `ClassReference.Psi.properties`
Given something like:

```kotlin
class Subject(
  val name: String
)
```

A `ClassReference.Descriptor` instance will include `name` as a property, but the `ClassReference.Psi` version will not.  `Descriptor`'s version is correct.

RBusarow:
fixes #609

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
@@ -1,0 +1,619 @@
+package com.squareup.anvil.compiler.internal.reference
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.compiler.compile
+import com.squareup.anvil.compiler.internal.testing.simpleCodeGenerator
+import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.OK
+import org.junit.Test
+
+class PropertyReferenceTest {
+
+  @Test fun `primary constructor val properties are properties`() {
+    compile(
+      """
+      package com.squareup.test
+
+      class Subject(
+        val name: String,
+        val age: Int
+      )
+      """,
+      allWarningsAsErrors = false,
+      codeGenerators = listOf(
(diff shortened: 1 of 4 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
```diff
@@ -0,0 +1,619 @@
...
+      )
+    ) {
+      assertThat(exitCode).isEqualTo(OK)
+    }
+  }
+
+  fun ClassReference.propertyTypeNames() = properties
+    .map { "${it.name}: ${it.type().asTypeName()}" }
+
+  fun List<PropertyReference>.named(name: String) = single { it.name == name }
```
vRallev, line 618: These functions should be private.

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
```diff
@@ -0,0 +1,619 @@
...
+              .inOrder()
+          }
+
+          null
+        }
+      )
+    ) {
+      assertThat(exitCode).isEqualTo(OK)
+    }
+  }
```
vRallev, line 613: Thanks for all these tests!

vRallev (approved):

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
RBusarow: fixed

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/PropertyReferenceTest.kt
RBusarow: Well, the tooling makes it so easy!

PR #623, opened 2022-07-25
Support contributing types with `Any` as `boundType`.

### LINKED ISSUE #619: boundType does not appear to support Any
This piece of code:

```
@ContributesMultibinding(ActivityInjection::class, boundType = Any::class)
class TestViewModel @Inject constructor() : SimpleStackViewModel() {
...
}
```

Will throw an error:

```TestViewModel contributes a binding for kotlin.Any, but doesn't extend this type.```

Obviously, this class DOES extend Any.

This worked in Anvil 2.3.4 but it's now broken in 2.4.1.

Setting bound type to the `SimpleStackViewModel` works, so it appears just using `Any` as a boundType is an issue.

### PR DESCRIPTION
**Pull request by vRallev**
Fixes #619

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
@@ -1,5 +1,6 @@
 package com.squareup.anvil.compiler.codegen
 
+import com.squareup.anvil.compiler.anyFqName
 import com.squareup.anvil.compiler.assistedInjectFqName
 import com.squareup.anvil.compiler.contributesMultibindingFqName
 import com.squareup.anvil.compiler.injectFqName
@@ -80,6 +81,10 @@ internal fun ClassReference.checkClassExtendsBoundType(
       message = "Couldn't find the bound type.",
       classReference = this
     )
+
+  // The boundType is declared explicitly in the annotation. Since all classes extend Any, we can
+  // stop here.
+  if (boundType.fqName == anyFqName) return
 
   if (allSuperTypeClassReferences().none { it.fqName == boundType.fqName }) {
     throw AnvilCompilationExceptionClassReference(
--- a/compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
@@ -309,6 +309,22 @@ class ContributesBindingGeneratorTest {
     }
   }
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

File: compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt [hunk 1]
```diff
@@ -81,6 +82,10 @@ internal fun ClassReference.checkClassExtendsBoundType(
       classReference = this
     )
 
+  // The boundType is declared explicitly in the annotation. Since all classes extend Any, we can
+  // stop here.
+  if (boundType.fqName == anyFqName) return
```
JoelWilcox, line 87: I'm wondering if it would make sense for us to instead modify `allSuperTypeClassReferences()` to include `Any` instead of special casing it here? Efficiency wise this makes sense, but I wonder if we'll potentially run into similar issues later where we're looking for Any and it's not found

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
```diff
@@ -309,6 +309,22 @@ class ContributesBindingGeneratorTest {
...
+      """
+      package com.squareup.test
+
+      import com.squareup.anvil.annotations.ContributesBinding
+
+      @ContributesBinding(Int::class, boundType = Any::class)
+      interface ContributingInterface
+      """
+    ) {
+      assertThat(contributingInterface.hintBinding?.java).isEqualTo(contributingInterface)
```
JoelWilcox, line 323: Is this assert the one that's checking for `Any`? This one isn't clear to me. E.g. why are we checking `hintBinding?.java` instead of just `hintBinding`, and why are we checking if it's equal to `contributingInterface` instead of `Any::class`

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ClassReferenceExtensions.kt
vRallev: I don't think so. That might break some existing code, so this is the safer bet. This method also mimics existing APIs from the Kotlin compiler and they exclude `Any` as well.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesBindingGeneratorTest.kt
vRallev: A hint is always the type that is contributed and the scope. Based on that information Anvil can fetch the annotation and read all values. `hintBinding` returns a `KClass` and `contributingInterface` is a `Class`, so one or the other has to be converted.

JoelWilcox (approved):

PR #625, opened 2022-07-26
Unwrap types from type aliases for `TypeReference`.

### LINKED ISSUE #611: Regression: generic typealias expands only one type argument
Anvil 2.4.1
Kotlin 1.7.0

Given:
    
    class RealType<T, U>
    
    typealias AliasType<X> = RealType<String, X>
    
    class User {
        @Inject
        lateinit var value: AliasType<Int>
    }

Anvil turns `AliasType<Int>` into `RealType<Int>` in the generated code, which causes a build error because `RealType` takes two type arguments, not one.

This is a regression from Anvil 2.4.0 and Kotlin 1.6.21 where this issue did not occur.

vRallev:
Fixes #611

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
@@ -128,7 +128,7 @@ public sealed class TypeReference {
       }
 
     override val unwrappedTypes: List<TypeReference> by lazy(NONE) {
-      type.typeElement!!.children
+      val unwrappedTypes = type.typeElement!!.children
         .filterIsInstance<KtTypeArgumentList>()
         .singleOrNull()
         ?.children
@@ -145,6 +145,8 @@ public sealed class TypeReference {
               .toTypeReference(declaringClass)
           }
         } ?: emptyList()
+
+      findUnwrappedTypesWithTypeAlias(unwrappedTypes)
     }
 
     @Suppress("RedundantNullableReturnType")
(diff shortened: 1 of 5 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -306,6 +308,42 @@ public sealed class TypeReference {
...
+      unwrappedTypes: List<TypeReference>
+    ): List<TypeReference> {
+      // For Psi we must inspect the type and check if further arguments have been specified in
+      // a type alias, e.g.
+      //
+      //   typealias MyType<X> = Pair<X, String>
+      //
+      //   val myType: MyType<Int>
+      //
+      // Without inspecting the type, Psi would only tell us that there's on argument `Int`. Type
```
JoelWilcox, line 322:
```suggestion
      // Without inspecting the type, Psi would only tell us that there's one argument `Int`. Type
```

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -306,6 +308,42 @@ public sealed class TypeReference {
...
+      // aliases are pretty much inlined, so we need to treat them specially. The idea is to check
+      // if the type is a type alias and if the type arguments differ. If no, then nothing special
+      // needs to happen. If yes, then we need to merge both type argument lists.
+
+      val children = type.typeElement?.children ?: return unwrappedTypes
+      if (children.size != 2 || children[1] !is KtTypeArgumentList) return unwrappedTypes
```
JoelWilcox, line 328: Are we guaranteed this is always exactly 2 for the case we care about? Or should this be `< 2`?

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TypeReferenceTest.kt
```diff
@@ -214,4 +214,52 @@ class TypeReferenceTest {
       assertThat(exitCode).isEqualTo(OK)
     }
   }
+
+  @Test fun `unwrapped types contain types from type aliases`() {
```
JoelWilcox, line 218: Great test 👨‍🍳

JoelWilcox (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
vRallev: There's always two for our case.

PR #626, opened 2022-07-26
Shared output directory

### LINKED ISSUE #602: Shared output directory for `DisableIncrementalCompilationTask`
image

vRallev:
Don't share the output directory for the `DisableIncrementalCompilationTask`, if there are multiple Kotlin compilation tasks for the same module.

Fixes #602

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/DisableIncrementalCompilationTask.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/DisableIncrementalCompilationTask.kt
@@ -23,7 +23,7 @@ internal abstract class DisableIncrementalCompilationTask : DefaultTask() {
 
   @Suppress("unused")
   @get:OutputFile @get:Optional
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/java/com/squareup/anvil/plugin/DisableIncrementalCompilationTask.kt
```diff
@@ -23,7 +23,7 @@ internal abstract class DisableIncrementalCompilationTask : DefaultTask() {
```
JoelWilcox, line 23: Is this Gradle Task still needed? Looks like https://youtrack.jetbrains.com/issue/KT-38570 is fixed now :)

JoelWilcox (approved):

gradle-plugin/src/main/java/com/squareup/anvil/plugin/DisableIncrementalCompilationTask.kt
vRallev: I need to find time to test it. I filed #627.

PR #629, opened 2022-07-28
Upgrade Kotlin to 1.7.10. 

vRallev:
Anvil will not be compatible with Kotlin 1.6 anymore.


dependencies.gradle
```diff
@@ -9,13 +9,12 @@ ext {
...
   kotlinVersion = project.hasProperty('square.kotlinVersion') ?
-      project.getProperty('square.kotlinVersion') : '1.6.10'
+      project.getProperty('square.kotlinVersion') : '1.7.10'
 
   kotlinUseIR = (project.hasProperty('square.useIR') ?
       project.getProperty('square.useIR') : 'true').toBoolean()
 
-  // TODO: remove this property with Kotlin 1.7.0.
-  kotlinPreview = !kotlinVersion.startsWith('1.6')
+  kotlinPreview = !kotlinVersion.startsWith('1.7.2')
```
RBusarow, line 17: Is this inverted?

vRallev: Yes, nice catch!

RBusarow (approved):

settings.gradle
```diff
@@ -6,14 +6,7 @@ include ':integration-tests:code-generator'
...
 include ':integration-tests:library'
-
-// TODO: There seems to be a bug with Kotlin 1.7 and MPP projects. Disable the Android
-//  module for now until the issue is resolved
-def kotlinVersion = hasProperty('square.kotlinVersion') ? getProperty('square.kotlinVersion').toString() : '1.6.10'
-if (VersionNumber.parse(kotlinVersion).baseVersion < VersionNumber.parse('1.7.0')) {
-  include ':integration-tests:mpp:android-module'
-}
-
+include ':integration-tests:mpp:android-module'
```
JoelWilcox, line 9: If I'm reading the previous condition correctly, weren't we only including this module for version < 1.7 before? (which would make me expect that we remove it altogether now that we're dropping 1.6)

JoelWilcox (approved):

settings.gradle
vRallev: We were only including it for 1.6, because there was an issue with 1.7. I resolved that issue by moving a source file and now I can include the module for 1.7 as well.

PR #630, opened 2022-07-28
Remove the implementation for the old compiler backend.

### LINKED ISSUE #422: Remove support for the old compiler backend

vRallev:
The stub generating Gradle task uses the IR backend by default starting with Kotlin 1.8.20. We no longer need to support the old backend therefore.

Fixes #422


vRallev:
7 tests are failing that use the Kotlin compile testing library and rely on Anvil to run during the stub generation. This issue doesn't seem to happen normally and only when using Kotlin compile testing.

Anvil uses an `IrGenerationExtension` to add the Dagger `@Component` annotation whenever `@MergeComponent` is used. This has to happen in the stub generating task, because Dagger must see the `@Component` annotation. In normal Kotlin compilations I can see the `@Component` annotation being added to the stubs, but not with Kotlin compile testing. With Kotlin compile testing I see how `AnvilComponentRegistrar` is called and the IR extension is registered for the stub task, but it’s not called in the stub generating task for some reason.

Because the `@Component` annotation isn't added Dagger doesn't run and the tests fail. These 7 tests are the only tests exercising this codepath and relying on that behavior.

vRallev:
After further testing I noticed that `kapt.use.jvm.ir` makes no difference and the tests fail because the implementation for the old backend `ModuleMerger` has been removed. With the Kotlin compile testing library we always exercise old backend codepath no matter what.

vRallev:
The new compile testing library fixes that.

JoelWilcox (approved):
🙌

vRallev:
This PR evolved over 9 months and finally can be merged 🙌

PR #632, opened 2022-08-03
Fix annotation arguments using string templates not being parsed correctly

### LINKED ISSUE #631: AnnotationArgumentReference.Psi does not parse string templates correctly
The string template parsing [Here](https://github.com/square/anvil/blob/855e55baeb8afdd81b22277f8477e86490ec5e2d/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt#L267) does not currently grab the entire string. When given a blob like

```kotlin
const val SOME_STR: String = "World!"

@AnnotatedWith(value = "Hello, $SOME_STR")
object SomeObj {}
```

`annotationArgRef.value<Any>()` returns the string `"Hello, "`. Seeing as the compiler enforces that all template entries are constantly resolvable the entire value should be resolvable, which would bring this in line with the other available expression types.

JoelWilcox:
Fixes #631

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt
@@ -14,6 +14,7 @@ import org.jetbrains.kotlin.com.intellij.psi.PsiElement
 import org.jetbrains.kotlin.lexer.KtTokens
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
+import org.jetbrains.kotlin.psi.KtBlockStringTemplateEntry
 import org.jetbrains.kotlin.psi.KtClass
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtCollectionLiteralExpression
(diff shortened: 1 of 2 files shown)


RBusarow (approved):
LGTM!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationArgumentReference.kt
```diff
@@ -263,11 +265,19 @@ public sealed class AnnotationArgumentReference {
...
-              ?.text
-              ?: fail()
+          is KtStringTemplateExpression -> {
+            val parts = psiElement.getChildrenOfType<KtStringTemplateEntry>()
+
+            parts.map { templateEntry ->
+              if (templateEntry is KtSimpleNameStringTemplateEntry ||
+                templateEntry is KtBlockStringTemplateEntry
+              ) {
+                parsePsiElement(templateEntry.expression as PsiElement)
```
0xMatthewGroves, line 275: It seems like the bug is that we're calling getChildOfType instead of getChildrenOfType which is why the rest of the string doesn't come along. Why are you parsing both of the elements instead of just doing a joinToString(", ") on both parts?

JoelWilcox: In this case there were two separate issues. You're correct about the first one which was that we were only grabbing text for the first part of the template regardless of template length. The second issue was that we were just treating the template as plain text, which results in constants staying as their code text instead of their value (e.g. returning `${CONSTANT}` instead of resolving it).

vRallev:
Beautiful PR. Thanks!

PR #634, opened 2022-08-17
Align Anvil with Dagger and don't support member injection for private fields

### LINKED ISSUE #341: MembersInjectorGenerator will generate nullable type args for nullable injections
Perhaps this is correct behavior, but to my knowledge Dagger does not allow nulls to be passed through directly.

```kotlin
class Example @Inject constructor() {
  @Inject
  var injected: String? = null
}
```

Will result in member injection refs like this

```kotlin
public class Example_Factory(
  private val param0: Provider<String?>
) : Factory<Example> {
  public override fun get(): Example {
    val instance = newInstance()
    Example_MembersInjector.injectInjected(instance, param0.get())
    return instance
  }
}
```

vRallev:
Fixes #341

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
@@ -5,6 +5,7 @@ import com.squareup.anvil.compiler.daggerDoubleCheckFqNameString
 import com.squareup.anvil.compiler.daggerLazyFqName
 import com.squareup.anvil.compiler.injectFqName
 import com.squareup.anvil.compiler.internal.capitalize
+import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionPropertyReference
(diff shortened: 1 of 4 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/DaggerGenerationUtils.kt
```diff
@@ -197,6 +199,16 @@ private fun PropertyReference.toMemberInjectParameter(
   uniqueName: String,
   implementingClass: ClassReference
 ): MemberInjectParameter {
+  if (!isLateinit() && !isAnnotatedWith(jvmFieldFqName)) {
+    // Technically this works with Anvil and we could remove this check. But we prefer consistency
+    // with Dagger.
+    throw AnvilCompilationExceptionPropertyReference(
+      propertyReference = this,
+      message = "Dagger does not support injection into private fields. Either use a " +
+        "'lateinit var' or '@JvmField'."
```
vRallev, line 208: Dagger only prints the first part of the message. But I thought it would be helpful to make this clearer.

RBusarow (approved):

PR #654, opened 2022-11-18
Validate @Binds methods when generating dagger factories

### LINKED ISSUE #649: @Binds are not validated with `generateDaggerFactories`
With pure dagger this will lead to a compilation error:

```kotlin
@Module
abstract class OTestModule {

    @Binds
    abstract fun testBinding(impl: Context): StringBuilder
}
```

But when `generateDaggerFactories` is enabled, compilation is successful and we get an error during runtime. 

It would be great to get a compilation failure also with `generateDaggerFactories` enabled.

JoelWilcox:
Fixes #649

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
@@ -1,0 +1,180 @@
+package com.squareup.anvil.compiler.dagger
+
+import com.google.common.truth.Truth.assertThat
+import com.squareup.anvil.compiler.USE_IR
+import com.squareup.anvil.compiler.WARNINGS_AS_ERRORS
+import com.squareup.anvil.compiler.internal.testing.compileAnvil
+import com.squareup.anvil.compiler.isError
+import com.squareup.anvil.compiler.isFullTestRun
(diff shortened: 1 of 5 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -0,0 +1,165 @@
...
+      interface Bar
+
+      @Module
+      abstract class BarModule {
+        @Binds
+        abstract fun bindsBar(impl: Foo): Bar
+      }
+      """
+    ) {
+      assertThat(exitCode).isEqualTo(COMPILATION_ERROR)
```
vRallev, line 49: There's a shortcut: `assertThat(exitCode).isError()`. In all these instances you should also verify the error message to ensure that the correct error is printed. This is quite important.

vRallev (changes requested):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
JoelWilcox: Good catch, updated now including error messages updated to exactly match Dagger's error messages.

vRallev (approved):

PR #655, opened 2022-11-19
Fix inconsistency between Dagger and Anvil for generated factory names…

### LINKED ISSUE #653: Dagger factory generator breaks in modules containing `-` in name
### Current behavior

Here `app-display` is the consuming app module and `display-legacy` is the producing library module. The generated code expects class `FakeRotationModule_Companion_FakeRotation$display_legacy_debugFactory`.
```
> Task :app-display:compileDebugJavaWithJavac FAILED
C:\gradle\build\...\app-display\generated\source\kapt\debug\...DaggerHomeComponent.java:6: error: cannot find symbol
import ....FakeRotationModule_Companion_FakeRotation$display_legacy_debugFactory;
                                                         ^
  symbol:   class FakeRotationModule_Companion_FakeRotation$display_legacy_debugFactory
  location: package ...
```
The factory exists on disk with a dash; Anvil generates class `FakeRotationModule_Companion_FakeRotation$display-legacy_debugFactory`.

image

### What's likely happening

`-` is a valid part of a Kotlin identifier so the source file is generated with a dash. Dagger would generate an underscore instead. This mismatch causes compile-time errors when attempting to generate a component from Anvil-generated `@Module`s in Gradle modules whose names contain `-`.

### Expected behavior

Anvil should use Java validation rules for class names when generating Dagger factories.

### Workarounds

* Use only valid Java identifiers as module names
* Use kapt and actual Dagger compiler

JoelWilcox:
… involving a dash-separated module name

Fixes #653

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
@@ -333,7 +333,8 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
   }
 
   private fun ModuleDescriptor.mangledNameSuffix(): String {
-    val name = name.asString()
+    // We replace - with _ to maintain interoperability with Dagger's expected generated identifiers
+    val name = name.asString().replace('-', '_')
     return if (name.startsWith('<') && name.endsWith('>')) {
       name.substring(1, name.length - 1)
     } else {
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
```diff
@@ -333,7 +333,8 @@ internal class ProvidesMethodFactoryGenerator : PrivateCodeGenerator() {
   }
 
   private fun ModuleDescriptor.mangledNameSuffix(): String {
-    val name = name.asString()
+    // We replace - with _ to maintain interoperability with Dagger's expected generated identifiers
+    val name = name.asString().replace('-', '_')
```
vRallev, line 337: I'm curious, without this fix would the test below fail in CI (Windows in particular)?

vRallev (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/ProvidesMethodFactoryGenerator.kt
JoelWilcox: Good question! It fails without the fix on my machine, but I just opened https://github.com/square/anvil/pull/656 to verify we see a failure in CI as well.

PR #660, opened 2022-12-09
Test Kotlin 1.8.0-RC in CI

JoelWilcox:


.github/workflows/ci.yml
```diff
@@ -19,7 +19,7 @@ jobs :
       # solution.
       fail-fast : false
       matrix :
-        kotlin-version : [ 1.7.20 ]
+        kotlin-version : [ 1.8.0-RC ]
```
vRallev, line 22:
```suggestion
        kotlin-version : [ 1.7.20, 1.8.0-RC ]
```
And everywhere else.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
```diff
@@ -429,7 +429,7 @@ class BindingModuleGeneratorTest(
     ) {
       assertThat(exitCode).isError()
 
-      assertThat(messages).contains("Source0.kt: (6, 11)")
+      assertThat(messages).contains("Source0.kt:6:11")
```
vRallev, line 432: Tedious change :/

dependencies.gradle
```diff
@@ -9,7 +9,7 @@ ext {
       project.getProperty('square.agpVersion') : '4.2.2'
 
   kotlinVersion = project.hasProperty('square.kotlinVersion') ?
-      project.getProperty('square.kotlinVersion') : '1.7.20'
+      project.getProperty('square.kotlinVersion') : '1.8.0-RC'
```
vRallev, line 12: Always a stable release should be the default.

dependencies.gradle
```diff
@@ -78,7 +78,7 @@ ext {
 
       kotlin: [
           annotation_processing_embeddable: "org.jetbrains.kotlin:kotlin-annotation-processing-embeddable:$kotlinVersion",
-          compile_testing: "com.github.tschuchortdev:kotlin-compile-testing:1.4.9",
+          compile_testing: "dev.zacsweers.kctfork:core:0.1.0-1.8.0-Beta01",
```
vRallev, line 81: That needs to go into the changelog.

vRallev (changes requested):

vRallev:
```
com.squareup.anvil.compiler.InterfaceMergerTest > anonymous inner classes should not be evaluated[class com.squareup.anvil.annotations.MergeComponent] STANDARD_OUT
    e: -Xuse-old-backend is no longer supported. Please migrate to the new JVM IR backend
```
Unsure what to do here. Because this is not true for the stub generating task: https://youtrack.jetbrains.com/issue/KT-53135

I think the path forward, if there is no other solution, is to keep supporting Kotlin 1.7.20 in parallel (until Kotlin 1.8 stable is out anyways) and disable these tests with Kotlin 1.8.

.github/workflows/ci.yml
JoelWilcox: For sure! Wasn't intending these as the final changes, just wanted to shake out any 1.8 issues before going through the more final changes of supporting both at once (due to the new @ExperimentalCompilerAPI changes)

dependencies.gradle
JoelWilcox: Same as other comment, this is temporary and will be changed back in the final changes

JoelWilcox:
> I think the path forward, if there is no other solution, is to keep supporting Kotlin 1.7.20 in parallel (until Kotlin 1.8 stable is out anyways) and disable these tests with Kotlin 1.8.

I agree, likely we'll disable the old-backend shard for 1.8+

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleGeneratorTest.kt
JoelWilcox: Aye, thanks @ZacSweers for handling the initial batch :). Needed to conditionally add back the old check now but at least using intellij find+replace with regex makes it pretty fast.

.github/workflows/ci.yml
```diff
@@ -19,7 +19,8 @@ jobs :
       # solution.
       fail-fast : false
       matrix :
-        kotlin-version : [ 1.7.20 ]
+        # We do not run this job with Kotlin 1.8 because the old backend is removed in 1.8
+        kotlin-version : [ 1.7.20]
```
vRallev, line 23: 👍🏻

vRallev (approved):
Do you also plan to automatically publish versions with Kotlin 1.8 in the `release.yaml`?

JoelWilcox:
> Do you also plan to automatically publish versions with Kotlin 1.8 in the release.yaml?

Good question -- yes, although I was planning to do a follow-up commit/PR based on it looking like you kept them separate previously (in the commits you shared). Was there any functional concern behind that? If not I could also update the release + snapshot workflows in this PR (still as their own commit).

vRallev:
No reason. Either a separate commit or PR works.

.github/workflows/release.yaml
```diff
@@ -20,8 +20,17 @@ jobs :
...
+      - name : Publish Release 1.7
+        run : ./gradlew clean publish --no-build-cache --no-daemon --stacktrace --no-parallel -Psquare.kotlinVersion=1.7.20 && cd gradle-plugin && ./gradlew clean publish --no-build-cache --no-daemon --stacktrace -Psquare.kotlinVersion=1.7.20 && cd ..
+        env :
+          ORG_GRADLE_PROJECT_mavenCentralUsername : ${{ secrets.SONATYPE_NEXUS_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword : ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+          ORG_GRADLE_PROJECT_signingInMemoryKey : ${{ secrets.ARTIFACT_SIGNING_PRIVATE_KEY }}
+          ORG_GRADLE_PROJECT_signingInMemoryKeyPassword : ''
+
+      - name : Publish Release 1.8
+        run : ./gradlew clean publish --no-build-cache --no-daemon --stacktrace --no-parallel -Psquare.kotlinVersion=1.8.0-RC -PVERSION_NAME=2.4.3-1-8-SNAPSHOT && cd gradle-plugin && ./gradlew clean publish --no-build-cache --no-daemon --stacktrace -Psquare.kotlinVersion=1.8.0-RC -PVERSION_NAME=2.4.3-1-8-SNAPSHOT && cd ..
```
vRallev, line 33: `-PVERSION_NAME=2.4.3-1-8-SNAPSHOT` looks wrong here. This is not a snapshot.

JoelWilcox: 🤦

PR #662, opened 2022-12-10
Handle extension functions when validating @Binds methods with generateDaggerFactories

### LINKED ISSUE #661: BindsMethodValidator does not account for extension functions
When using `generateDaggerFactories`, this results in anvil failing to compile for a @Binds extension function with no parameters (when Dagger passes), and to pass for a @Binds extension function with a parameter (when Dagger fails).

Thanks @efemoney for [raising the extension use-case](https://github.com/square/anvil/pull/654#issuecomment-1324879737).

JoelWilcox:
Fixes #661

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
@@ -6,6 +6,7 @@ import com.squareup.anvil.compiler.WARNINGS_AS_ERRORS
 import com.squareup.anvil.compiler.internal.testing.compileAnvil
 import com.squareup.anvil.compiler.isError
 import com.squareup.anvil.compiler.isFullTestRun
+import com.tschuchort.compiletesting.KotlinCompilation.ExitCode.OK
 import com.tschuchort.compiletesting.KotlinCompilation.Result
 import org.intellij.lang.annotations.Language
 import org.junit.Test
@@ -138,6 +139,35 @@ class BindsMethodValidatorTest(
   }
 
   @Test
+  fun `an extension function binding with a parameter fails to compile`() {
+    compile(
+      """
+      package com.squareup.test
+ 
+      import dagger.Binds
+      import dagger.Module
+      import javax.inject.Inject
+
+      class Foo @Inject constructor() : Bar
+      class Hammer @Inject constructor() : Bar
+      interface Bar
+
+      @Module
+      abstract class BarModule {
+        @Binds
+        abstract fun Foo.bindsBar(impl2: Hammer): Bar
+      }
+      """
+    ) {
+      assertThat(exitCode).isError()
+      assertThat(messages).contains(
+        "@Binds methods must have exactly one parameter, " +
+          "whose type is assignable to the return type"
+      )
+    }
+  }
+
+  @Test
   fun `a binding with no return type fails to compile`() {
     compile(
       """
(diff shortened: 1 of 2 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -164,6 +194,30 @@ class BindsMethodValidatorTest(
...
+      interface Bar
+
+      @Module
+      abstract class BarModule {
+        @Binds
+        abstract fun Foo.bindsBar(): Bar
+      }
+      """
+    ) {
+      assertThat(exitCode).isEqualTo(OK)
```
vRallev, line 217: I'm not sure what you're testing here. Remove Anvil and the Dagger annotation processor from the build and the test would be green no matter what. Do you want to add a Dagger component to verify the binding method works?

Also, Anvil isn't doing anything with binding methods. There's also no Anvil annotation present in this entire test. It's a no-op at the moment.

JoelWilcox: This is testing that our @Binds validation logic doesn't fail for a valid use-case; without the changes to `BindsMethodValidator` in this PR, this test fails because we would see that there was no parameter and throw an exception, while Dagger would compile successfully. Does that make sense, or do you think this is not worth including? For the purposes of this test it doesn't seem like including a Dagger component is necessary ; or at least, it's not clear why that would be needed here but not the other test cases. Open to being convinced otherwise :)

vRallev: I guess it doesn't hurt. I suggested to add a component and make sure that the binding works, to give this test more meaning in the sense: hey, here's a binding method and it works at runtime with Anvil or Dagger. I've done this at a few other places.

vRallev (approved):

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
JoelWilcox: Gotcha, that makes sense; I went ahead and added it to help with that clarity.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -164,14 +194,56 @@ class BindsMethodValidatorTest(
...
+    compile(
+      """
+      package com.squareup.test
+
+      import dagger.Component
+      import javax.inject.Singleton
+
+      @Component(modules = [BarModule::class])
+      @Singleton
+      interface ComponentInterface
```
vRallev, line 229:
```suggestion
      interface ComponentInterface {
        fun bar(): Bar
      }
```
Otherwise Dagger won't care about the binding (or in different words, you could remove the Dagger module from the component and it still would compile).

JoelWilcox: Good catch 👍

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -164,14 +194,58 @@ class BindsMethodValidatorTest(
...
+
+    compile(
+      """
+      package com.squareup.test
+
+      import dagger.Component
+      import javax.inject.Singleton
+
+      @Component(modules = [BarModule::class])
+      @Singleton
```
vRallev, line 228: Super nit, but the `@Singleton` isn't needed.

vRallev (approved):
👍🏻 LGTM

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
JoelWilcox: Will make a mental note to remove separately :D, want to get this in before merging the release updates

PR #664, opened 2022-12-15
Update FqName detection to fix validation of extension function bindings with qualifiers

JoelWilcox:
Currently when something like @receiver:Qualifier is used on an extension function, the FqName utils don't have any cases to handle it and try to resolve the combined string of `@receiver:Qualifier ReceivingType` which results in us not finding the FqName for `ReceivingType`

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
@@ -15,6 +15,7 @@ import org.jetbrains.kotlin.psi.KtAnnotationEntry
 import org.jetbrains.kotlin.psi.KtClassLiteralExpression
 import org.jetbrains.kotlin.psi.KtClassOrObject
 import org.jetbrains.kotlin.psi.KtConstructorCalleeExpression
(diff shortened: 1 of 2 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -225,7 +225,6 @@ class BindsMethodValidatorTest(
       import javax.inject.Singleton
 
       @Component(modules = [BarModule::class])
-      @Singleton
```
JoelWilcox, line 228: This wasn't actually important for this test

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/PsiUtils.kt
```diff
@@ -125,6 +126,14 @@ public fun PsiElement.requireFqName(
           // Fallback to the text representation.
           text
         }
+      } else if (children.size == 2 && children[0] is KtDeclarationModifierList) {
```
vRallev, line 129: That's now a `when` statement.

PR #666, opened 2022-12-15
Improve the error message for failed binding validations when the parameter and return type do not match

JoelWilcox:
This should make it more obvious why a given binding failed without needing to check the generated code.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
@@ -7,6 +7,7 @@ import com.squareup.anvil.compiler.codegen.PrivateCodeGenerator
 import com.squareup.anvil.compiler.daggerBindsFqName
 import com.squareup.anvil.compiler.daggerModuleFqName
 import com.squareup.anvil.compiler.internal.reference.AnvilCompilationExceptionFunctionReference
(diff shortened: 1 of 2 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -75,28 +76,43 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
     )
 
     if (!function.parameterMatchesReturnType() && !function.receiverMatchesReturnType()) {
+      val returnType = function.returnType().asClassReference().shortName
+      val paramSuperTypes = (function.parameterSuperTypes() ?: function.receiverSuperTypes())!!
```
ZacSweers, line 80: May be worth tweaking the message if there's no supertypes or it's only Any, but up to you.

ZacSweers (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
JoelWilcox: Good point, updated to adjust the message when there are no supertypes, and to not list the parameter type itself in the supertype list. Will leave out special casing for Any for now as I believe we actually already filter it out for getting the supertypes.

PR #675, opened 2023-01-11
Allow querying top-level functions and properties

### LINKED ISSUE #644: No way to create a `FunctionReference` to a top level function
We are looking at code gen use case such as below

```kt
@ContributesJsonAdapter
internal fun Moshi.Builder.myAdapter(
  one: One,
  @Named("foo") another: Another,
) {
  add(MyAdapter(one, other))
}
```

that will generate:

```kt
@ContributesMultibinding(AppScope)
class MyAdapterFn_Registrar(
  one: One,
  @Named("foo") another: Another,
): JsonAdapterRegistrar {

  override fun register(moshi: Moshi.Builder) {
    moshi.myAdapter(one, another)
  }
}
```

There are no out-of-the-box APIs to get all top level functions (at least not as cleanly as with `ClassRefeference`s). Now even when we get the top-level `KtNamedFunction`s from the file, we cannot convert them to `FunctionReference`s because that API needs a `declaringClass` `ClassReference` which (at least to my knowlege) we cannot get from `KtNamedFunction` or `KtFile`

Also tried `KtFile.facadeClass().[...]` and that was a dead end too. 

How can we represent a top level function as a `FunctionReference`?

vRallev:
Fixes #644

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
@@ -97,7 +97,7 @@ public sealed class FunctionReference : AnnotatedReference {
     }
 
     override val returnType: TypeReference.Psi? by lazy(NONE) {
-      function.typeReference?.toTypeReference(declaringClass)
+      function.typeReference?.toTypeReference(declaringClass, module)
     }
 
     override val parameters: List<ParameterReference.Psi> by lazy(NONE) {
@@ -141,7 +141,7 @@ public sealed class FunctionReference : AnnotatedReference {
     }
 
     override val returnType: TypeReference.Descriptor? by lazy(NONE) {
-      function.returnType?.toTypeReference(declaringClass)
+      function.returnType?.toTypeReference(declaringClass, module)
     }
 
     internal val overriddenFunctions by lazy(NONE) {
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionalReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionalReference.kt
@@ -1,0 +1,55 @@
+package com.squareup.anvil.compiler.internal.reference
+
+import com.squareup.anvil.annotations.ExperimentalAnvilApi
+import com.squareup.anvil.compiler.api.AnvilCompilationException
+import com.squareup.anvil.compiler.internal.reference.FunctionalReference.Descriptor
+import com.squareup.anvil.compiler.internal.reference.FunctionalReference.Psi
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtFunction
+
+@ExperimentalAnvilApi
+public sealed interface FunctionalReference {
+  public val fqName: FqName
+  public val name: String get() = fqName.shortName().asString()
+
+  public val module: AnvilModuleDescriptor
+
+  public val parameters: List<ParameterReference>
+
+  public fun returnTypeOrNull(): TypeReference?
+  public fun returnType(): TypeReference = returnTypeOrNull()
+    ?: throw AnvilCompilationExceptionFunctionalReference(
+      functionalReference = this,
+      message = "Unable to get the return type for function $fqName."
+    )
+
+  public fun visibility(): Visibility
+
+  public sealed interface Psi : FunctionalReference {
+    public val function: KtFunction
+  }
+
+  public sealed interface Descriptor : FunctionalReference {
+    public val function: FunctionDescriptor
+  }
+}
+
+@ExperimentalAnvilApi
+@Suppress("FunctionName")
+public fun AnvilCompilationExceptionFunctionalReference(
+  functionalReference: FunctionalReference,
+  message: String,
+  cause: Throwable? = null
+): AnvilCompilationException = when (functionalReference) {
+  is Psi -> AnvilCompilationException(
+    element = functionalReference.function,
+    message = message,
+    cause = cause
+  )
+  is Descriptor -> AnvilCompilationException(
+    functionDescriptor = functionalReference.function,
+    message = message,
+    cause = cause
+  )
+}
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
@@ -50,7 +50,7 @@ import kotlin.LazyThreadSafetyMode.NONE
 @ExperimentalAnvilApi
 public sealed class TypeReference {
 
-  public abstract val declaringClass: ClassReference
+  public abstract val declaringClass: ClassReference?
 
   /**
    * If the type refers to class and is not a generic type, then the returned reference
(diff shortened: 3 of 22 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/TypeReference.kt
```diff
@@ -50,7 +50,7 @@ import kotlin.LazyThreadSafetyMode.NONE
 @ExperimentalAnvilApi
 public sealed class TypeReference {
 
-  public abstract val declaringClass: ClassReference
+  public abstract val declaringClass: ClassReference?
```
vRallev, line 53: I had to change this for `TypeReference` to share the class between member references and top-level references.

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelFunctionReferenceTest.kt
```diff
@@ -0,0 +1,161 @@
...
+      codeGenerators = listOf(
+        object : CodeGenerator {
+          override fun isApplicable(context: AnvilContext): Boolean = true
+
+          override fun generateCode(
+            codeGenDir: File,
+            module: ModuleDescriptor,
+            projectFiles: Collection<KtFile>
+          ): Collection<GeneratedFile> {
+            val functions = projectFiles.topLevelFunctionReferences(module).toList()
```
vRallev, line 50: That's the entry point similar to `classesAndInnerClasses(module)`.

integration-tests/code-generator/src/main/java/com/squareup/anvil/test/TestCodeGenerator.kt
```diff
@@ -173,6 +175,56 @@ class TestCodeGenerator : CodeGenerator {
           ),
         )
       }
+      .plus(
```
vRallev, line 178: An integration test to make sure it works in practice.

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionalReference.kt
```diff
@@ -0,0 +1,55 @@
...
+import com.squareup.anvil.annotations.ExperimentalAnvilApi
+import com.squareup.anvil.compiler.api.AnvilCompilationException
+import com.squareup.anvil.compiler.internal.reference.FunctionalReference.Descriptor
+import com.squareup.anvil.compiler.internal.reference.FunctionalReference.Psi
+import org.jetbrains.kotlin.descriptors.FunctionDescriptor
+import org.jetbrains.kotlin.name.FqName
+import org.jetbrains.kotlin.psi.KtFunction
+
+@ExperimentalAnvilApi
+public sealed interface FunctionalReference {
```
ZacSweers, line 12: `FunctionalReference` seems a bit of an odd name, vs maybe doing `FunctionReference` with subtypes of `TopLevelFunctionReference` and `RegularFunctionReference`? I admit at a glance I'm not sure I followed why they need to be separate implementations

ZacSweers (approved):

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionalReference.kt
vRallev: We'd need to rename the current class, which would also be a significant breaking change for consumers and custom code generators. We don't give any API stability guarantees, but nonetheless I wanted to avoid this. Initially I considered:
```
FunctionReference
-> MemberFunctionReference
-> TopLevelFunctionReference
```

> I admit at a glance I'm not sure I followed why they need to be separate implementations

Answer is here: https://github.com/square/anvil/pull/675#discussion_r1067292750

ZacSweers: Gotcha. I admit the name doesn't feel great but I don't have a better alternative to suggest!

compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/FunctionReference.kt
```diff
@@ -73,7 +61,7 @@ public sealed class FunctionReference : AnnotatedReference {
 
   override fun equals(other: Any?): Boolean {
     if (this === other) return true
-    if (other !is ClassReference) return false
+    if (other !is FunctionReference) return false
```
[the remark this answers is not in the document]
JoelWilcox: I'm a bit surprised this didn't result in any test failures previously 🙈

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelPropertyReferenceTest.kt
```diff
@@ -0,0 +1,189 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtProperty
+import org.jetbrains.kotlin.resolve.scopes.DescriptorKindFilter
+import org.jetbrains.kotlin.resolve.source.getPsi
+import org.junit.Test
+import java.io.File
+
+class TopLevelPropertyReferenceTest {
+
+  @Test fun `top level properties are parsed for PSI and Descriptor APIs correctly`() {
```
JoelWilcox, line 24: same here for having tests with more explicit expectations / use-cases

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelPropertyReferenceTest.kt
```diff
@@ -0,0 +1,189 @@
...
+                      .isEqualTo("javax.inject.Inject")
+                    assertThat(ref.setterAnnotations).isEmpty()
+                    assertThat(ref.isLateinit()).isFalse()
+                    assertThat(ref.visibility()).isEqualTo(PUBLIC)
+                    assertThat(ref.type().asClassReference().fqName.asString())
+                      .isEqualTo("kotlin.String")
+                  }
+
+                  "prop6" -> {
+                    // lateinit var prop6: String
```
JoelWilcox, line 129: Any reason to leave this comment here compared to the other test cases?

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelPropertyReferenceTest.kt
vRallev: Forgot to remove, only added it because my screen was too small.

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelPropertyReferenceTest.kt
vRallev: Done

compiler/src/test/java/com/squareup/anvil/compiler/internal/reference/TopLevelFunctionReferenceTest.kt
```diff
@@ -0,0 +1,161 @@
...
+import org.jetbrains.kotlin.psi.KtFile
+import org.jetbrains.kotlin.psi.KtFunction
+import org.jetbrains.kotlin.resolve.scopes.DescriptorKindFilter
+import org.jetbrains.kotlin.resolve.source.getPsi
+import org.junit.Test
+import java.io.File
+
+class TopLevelFunctionReferenceTest {
+
+  @Test fun `top level functions are parsed for PSI and Descriptor APIs correctly`() {
```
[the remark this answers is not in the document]
vRallev: Done

JoelWilcox (approved):
Thanks for making the test changes! Much easier to grok now

PR #678, opened 2023-01-13
Modernize build

ZacSweers:
**Highlights**
- Gradle 7.6
- Switch to using version catalogs
- Modernize gradle build bits using things like `pluginManagement`, `dependencyResolutionManagement`, etc.
- Use Java 11 for integration and samples
  - Necessary because of Dagger using the Java 11+ Generated annotation even though the release target is 8. Figure this is more or less ok but lmk if you want to do something different.
- Compile/target SDK 33
- Use plugins {} syntax in gradle
- Update Actions CI for new override pattern. Overrides are set from env vars in `settings.gradle` files.
- Update maven publishing plugin
- Testing previews requires the `anvil.allowSnapshots` gradle property.


.github/workflows/ci.yml
```diff
@@ -22,18 +25,20 @@ jobs :
...
+      - uses : actions/setup-java@v3
         with :
           distribution : 'adopt'
           java-version : '11'
           check-latest : true
 
       - name : Test on Ubuntu
-        run : ./gradlew assemble test -Psquare.useIR=false --no-build-cache --no-daemon --stacktrace -Psquare.kotlinVersion=${{ matrix.kotlin-version }}
+        env:
+          DEP_OVERRIDE_config-useIR: "false"
```
JoelWilcox, line 37: How does this work now when running from commandline locally? Same with e.g. wanting to run a command with a different version of Kotlin

gradle-plugin/settings.gradle
```diff
@@ -1 +1,46 @@
...
+    google()
+    gradlePluginPortal()
+    mavenCentral()
+  }
+}
+
+dependencyResolutionManagement {
+  versionCatalogs {
+    libs {
+      // TODO remove this once the dirs aren't symlinked anymore
```
JoelWilcox, line 12: Does this mean remove the commented out `from(files` line, the `if (System.getenv("DEP_OVERRIDES") == "true")` block, or something else? Could use a little more explicitness on the comment. 

Also, are we not able to remove the symlinks now?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,64 @@
...
+agp = "7.1.0"
+autoService = "1.0.1"
+autoValue = "1.10"
+dagger = "2.39.1"
+espresso = "3.5.1"
+kotlin = "1.8.0"
+ktlint = "0.41.0"
+
+# Configs that we can override in CI. Not exactly "versions" but useful to repurpose it here.
+config-fullTestRun = "true"
```
JoelWilcox, line 11: I feel like this chunk of the previous comment from dependencies.gradle is worth preserving:
```
We don't need to run all tests for sanity checks. If the flag is
false, then we run each test with one annotation instead of all options. We also skip tests
that run the Dagger annotation processor (KAPT is slow).
```

build.gradle
```diff
@@ -68,42 +69,44 @@ subprojects {
...
+    compilerOptions {
+      if (isIntegrationTestOrSample) {
+        jvmTarget.set(org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_11)
+      } else {
+        jvmTarget.set(org.jetbrains.kotlin.gradle.dsl.JvmTarget.JVM_1_8)
       }
 
-      allWarningsAsErrors = rootProject.ext.makeWarningsErrors
-      freeCompilerArgs += ["-opt-in=kotlin.RequiresOptIn"]
+      allWarningsAsErrors.set(isCi && (useKotlinIr) && false)
```
JoelWilcox, line 88: This comment/TODO seems like it should be preserved as well
```
// TODO: Remove when we remove the old backend.
  // TODO Repeatable: After cleaning up all deprecated methods remove the `&& false`.
```

I'm not sure if the first comment was just referring to `useKotlinIr` or not. @vRallev do you remember?

settings.gradle
```diff
@@ -1,3 +1,46 @@
...
+  repositories {
+    google()
+    gradlePluginPortal()
+    mavenCentral()
+  }
+}
+
+dependencyResolutionManagement {
+  versionCatalogs {
+    if (System.getenv("DEP_OVERRIDES") == "true") {
```
JoelWilcox, line 11: Can you add a helper method for handling the overrides stuff so that we don't need to duplicate it? I think that would be possible if we defined it in buildSrc? 🤔

integration-tests/library/build.gradle
```diff
@@ -18,5 +20,5 @@ kotlin {
 }
 
 dependencies {
-  api deps.dagger2.dagger
+  api libs.dagger2
```
JoelWilcox, line 23: 😂 I feel like we could also remove the 2 (not actually asking you to do this)

gradle-plugin/build.gradle
```diff
@@ -62,12 +62,11 @@ ktlint {
...
-  implementation libs.kotlin.gradlePlugin
-  implementation libs.kotlin.gradlePluginApi
-
   // Necessary to bump a transitive dependency.
   compileOnly libs.kotlin.reflect
 
-  // Compile only so that Java / Kotlin modules don't download it unnecessarily.
+  // Compile only so we don't pre-empty what the consuming project actually uses
+  compileOnly libs.kotlin.gradlePlugin
+  compileOnly libs.kotlin.gradlePluginApi
```
JoelWilcox, line 70: Would this be a visible change to consumers / worth calling out in the changelog?

On that note, can you update the changelog for the overall PR? Going back through all the commits later would be a bit of a hassle unless you want to rebase and squash some of them 😁

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -136,7 +136,6 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
     if (variant.variantFilter.syncGeneratedSources) {
       val isIdeSyncProvider = project.providers
         .systemProperty("idea.sync.active")
-        .forUseAtConfigurationTime()
```
JoelWilcox, line 139: Does this not need to be replaced by a new api call?

build.gradle
```diff
@@ -94,7 +94,7 @@ subprojects {
     }
   }
 
-  configurations.all {
+  configurations.configureEach {
```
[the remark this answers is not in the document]
ZacSweers: Just something I noticed, `all` eagerly configures all of these whereas `configureEach` does so lazily and includes all future added configurations

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.5.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+networkTimeout=10000
```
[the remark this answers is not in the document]
ZacSweers: This is generated by gradle so I was opting not to change it. Can remove if you want, but you'd have to manually remove it every time the wrapper's updated

.github/workflows/ci.yml
ZacSweers: you can either just manually change it in `libs.versions.toml` or do `DEP_OVERRIDES=true DEP_OVERRIDE_kotlin=1.8.10-beta ./gradlew <task>`

gradle/libs.versions.toml
```diff
@@ -0,0 +1,64 @@
...
+agp-library = { id = "com.android.library", version.ref = "agp" }
+anvil = { id = "com.squareup.anvil", version = "2.4.4" }
+gradlePublish = { id = "com.gradle.plugin-publish", version = "0.15.0" }
+kotlin-android = { id = "org.jetbrains.kotlin.android", version.ref = "kotlin" }
+kotlin-dokka = { id = "org.jetbrains.dokka", version = "1.7.20" }
+kotlin-jvm = { id = "org.jetbrains.kotlin.jvm", version.ref = "kotlin" }
+kotlin-kapt = { id = "org.jetbrains.kotlin.kapt", version.ref = "kotlin" }
+kotlin-multiplatform = { id = "org.jetbrains.kotlin.multiplatform", version.ref = "kotlin" }
+ktlint = { id = "org.jlleitschuh.gradle.ktlint", version = "10.2.0" }
+mavenPublish = { id = "com.vanniktech.maven.publish", version = "0.22.0" }
```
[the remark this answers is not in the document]
ZacSweers: the semantics are different here - `maven-publish` would require you to declare like `libs.plugins.maven.publish` despite that there are no other nested artifacts. The camel case here is the convention for single artifacts to disambiguate them from nested ones

settings.gradle
ZacSweers: you can't share logic in pluginManagement or dependencyResolutionManagement blocks unfortunately :/

gradle-plugin/build.gradle
ZacSweers: Definitely good to mention in the changelog. Will add it 👍

gradle/libs.versions.toml
```diff
@@ -0,0 +1,64 @@
+[versions]
+agp = "7.1.0"
```
vRallev, line 2: Worth adding an entry to the CHANGELOG.

.github/workflows/ci.yml
vRallev: What exactly is the command? I frequently disable running all tests either through my `gradle.properties` in my home dir `square.fullTestRun=false` or as `-Psquare.fullTestRun=false`. I can't follow what I do need to do in both cases now.

vRallev (commented):
This PR is much bigger than I anticipated. I'm really not a fan of mixing the migration to a `.toml` file with upgrading AGP and Gradle. This should have been a separate PR for each. Now, if all these commits are squashed it's not possible to revert one or the other anymore. Would it be possible to separate these things? I'm almost afraid to say that almost every bullet point from the PR message should have been a separate PR. At the moment I can't tell if something will break, there are just too many changes at once.

Other than it's unclear to me how the overrides now work. I asked this with a comment inline as well.

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
vRallev: Maybe we should hold off with the Gradle bump and make this separate: https://github.com/gradle/gradle/issues/23215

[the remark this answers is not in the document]
ZacSweers: We haven't noticed it in our project tbh

.github/workflows/ci.yml
ZacSweers: instead of `-Psquare.fullTestRun=false` you would prefix your command with this

```
DEP_OVERRIDES=true DEP_OVERRIDE_config-fullTestRun=true ./gradlew <your check>
```

`DEP_OVERRIDES` is a flag to inform dependency management that we have overrides in env vars it should look for. `DEP_OVERRIDE_<toml version key>` is the syntax for replacing a toml version key with something else.

gradle-plugin/settings.gradle
ZacSweers: With version catalogs you could actually start using Renovate and let it handle updates like that for you. Gradle in general really doesn't like symlinks. I can leave it in place and remove the comment if you want though

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
ZacSweers: It was deprecated in Gradle 7.4 though, which is quite old now. Does that seem ok?

build.gradle
```diff
@@ -68,42 +69,44 @@ subprojects {
...
 
-  //noinspection UnnecessaryQualifiedReference
-  tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompilationTask).configureEach {
-    kotlinOptions {
-      if (parent?.path != ':integration-tests:mpp') {
-        // The configuration for KMP projects is different. We should fix that eventually.
-        //
-        // Caused by: groovy.lang.MissingPropertyException: Could not set unknown property 'jvmTarget' for object of type org.jetbrains.kotlin.gradle.tasks.internal.KotlinMultiplatformCommonOptionsCompat.
-        jvmTarget = JavaVersion.VERSION_1_8
+  tasks.withType(KotlinCompile).configureEach {
```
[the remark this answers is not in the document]
ZacSweers: wait does spotless not clear unused imports in your build files?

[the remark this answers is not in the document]
ZacSweers: I guess this project doesn't have spotless enabled for groovy files

build.gradle
```diff
@@ -68,42 +69,44 @@ subprojects {
...
-      freeCompilerArgs += ["-opt-in=kotlin.RequiresOptIn"]
+      allWarningsAsErrors.set(isCi && (useKotlinIr) && false)
+      freeCompilerArgs.add("-opt-in=kotlin.RequiresOptIn")
 
       boolean optInExperimental = project.name != "annotations"
           && project.name != "scopes"
           && project.name != "dagger-factories-only"
       if (optInExperimental) {
-        freeCompilerArgs += "-opt-in=com.squareup.anvil.annotations.ExperimentalAnvilApi"
+        freeCompilerArgs.add("-opt-in=com.squareup.anvil.annotations.ExperimentalAnvilApi")
```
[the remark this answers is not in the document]
ZacSweers: Kotlin 1.8 introduced lazy property APIs

publishing.gradle
```diff
@@ -1,8 +0,0 @@
-apply plugin: 'com.vanniktech.maven.publish'
-apply plugin: 'org.jetbrains.dokka'
-
-plugins.withId("com.vanniktech.maven.publish") {
-  mavenPublish {
-    sonatypeHost = "S01"
-  }
-}
```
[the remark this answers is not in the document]
ZacSweers: It didn't need to stay separated since this setup is trivial at this point

gradle-plugin/build.gradle
```diff
@@ -74,16 +58,15 @@ kotlin {
...
-
   // Necessary to bump a transitive dependency.
-  compileOnly deps.kotlin.reflect
+  compileOnly libs.kotlin.reflect
 
-  // Compile only so that Java / Kotlin modules don't download it unnecessarily.
-  compileOnly deps.android_gradle_plugin
+  // Compile only so we don't pre-empty what the consuming project actually uses
+  compileOnly libs.kotlin.gradlePlugin
+  compileOnly libs.kotlin.gradlePluginApi
```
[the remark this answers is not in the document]
ZacSweers: the API plugin in particular or making these compileOnly in general?

gradle/libs.versions.toml
```diff
@@ -0,0 +1,64 @@
...
+
+# Configs that we can override in CI. Not exactly "versions" but useful to repurpose it here.
+config-fullTestRun = "true"
+config-generateDaggerFactoriesWithAnvil = "true"
+config-useIr = "true"
+
+[plugins]
+agp-application = { id = "com.android.application", version.ref = "agp" }
+agp-library = { id = "com.android.library", version.ref = "agp" }
+anvil = { id = "com.squareup.anvil", version = "2.4.4" }
```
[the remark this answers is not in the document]
ZacSweers: Unfortunately yes, gradle doesn't let you have a version-less plugin declaration even if it's unused

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
JoelWilcox: If it's auto-generated then that's fine :+1:

I would agree on waiting to bump until that issue is resolved, we're seeing similar issues with a draft upgrade internally

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
JoelWilcox: If it's deprecated and there's not equivalent replacement or workaround from Gradle then removing it seems reasonable. But we should verify the new behavior and add a note in the changelog if that new restriction exists now.

gradle/libs.versions.toml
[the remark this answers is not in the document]
JoelWilcox: I take it that it's not possible to set it dynamically with a function like we were previously, correct? Perhaps it would make sense to move the definition of `VERSION_NAME` to this file to keep that value being defined on only one place even if it doesn't technically make sense as part of `libs`. Just trying to think of what other options we have here.

gradle-plugin/build.gradle
ZacSweers: Done

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
ZacSweers: This isn't propagated to downstream users though and isn't affected in this project itself. Are you seeing issues on this project?

gradle/libs.versions.toml
ZacSweers: 310f051

gradle/libs.versions.toml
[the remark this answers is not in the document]
ZacSweers: I'm not sure I understand, the version here does not matter as we replace it with dependency substitution in the build itself. The toml format just needs to have one set for its own sanity

buildSrc/settings.gradle
```diff
@@ -0,0 +1,43 @@
...
+      from(files("../gradle/libs.versions.toml"))
+      if (System.getenv("DEP_OVERRIDES") == "true") {
+        Map<String, String> overrides = new LinkedHashMap()
+        def mergedProperties = System.getenv().plus(System.getProperties())
+        mergedProperties.forEach { key, value ->
+          String keyString = key.toString()
+          if (keyString.startsWith("DEP_OVERRIDE_")) {
+            overrides.put(keyString, value.toString())
+          }
+        }
```
efemoney, line 21: Since 7.5 there is `providers.xxxPrefixedBy(...)` for `systemProperties` & `environmentVariables` (and since 8.0 there is the equivalent for `gradleProperties`), might make this entire block and the subsequent code slightly more readable (plus compatible with config cache)

ZacSweers: Those aren't available in this context

ZacSweers:
Updated per our chat in kotlin-lang, now it uses system props and you can override by using the `-Doverride_{toml version}=<version>` syntax

```bash
./gradlew tasks -Doverride_kotlin=1.7.22
```

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
JoelWilcox: Hmm without doing some testing I think it would be hard to say that we're not affected by it (unless you mean you compared the memory usage already), even if it's not causing any failures yet. E.g. memory usage could be inflated for us now but not quite hitting the point of getting OOMs yet. 

My thought process is: is there any particular improvement in 7.6 we need right now compared to waiting for 7.6.1 when this issue will be resolved? Open to being convinced to leave it but if we don't have a need for it right now, waiting seems preferable.

gradle/libs.versions.toml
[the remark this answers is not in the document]
ZacSweers: [d4a3ea3](https://github.com/square/anvil/pull/678/commits/d4a3ea335c6067b0bf7cc5eb04382fe71820b43f)

gradle/wrapper/gradle-wrapper.properties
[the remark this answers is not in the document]
ZacSweers: [5c2aff3](https://github.com/square/anvil/pull/678/commits/5c2aff3a9b75894fe8065b162756de2b50a7ad18)

gradle/libs.versions.toml
```diff
@@ -0,0 +1,64 @@
+[versions]
+agp = "7.1.0"
+autoService = "1.0.1"
+autoValue = "1.10"
+dagger = "2.39.1"
+espresso = "3.5.1"
+kotlin = "1.8.0"
+ktlint = "0.41.0"
+
+# Configs that we can override in CI. Not exactly "versions" but useful to repurpose it here.
```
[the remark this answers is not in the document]
ZacSweers: [e3091bc](https://github.com/square/anvil/pull/678/commits/e3091bcd7cf321d5749cc6a66f123ae2d24c4468)

gradle/libs.versions.toml
```diff
@@ -0,0 +1,71 @@
...
+#
+# Example: ./gradlew tasks -Doverride_kotlin=1.7.22
+#
+
+# We don't need to run all tests for sanity checks. If the flag is
+# false, then we run each test with one annotation instead of all options. We also skip tests
+# that run the Dagger annotation processor (KAPT is slow).
+config-fullTestRun = "true"
+config-generateDaggerFactoriesWithAnvil = "true"
```
JoelWilcox, line 20: It looks like there are two more references to the old property for this that need to be updated, in `compare_size.sh`

gradle-plugin/build.gradle
```diff
@@ -1,35 +1,20 @@
...
-apply plugin: 'java-gradle-plugin'
-apply plugin: 'org.jetbrains.kotlin.jvm'
-apply plugin: 'com.gradle.plugin-publish'
-apply plugin: 'org.jlleitschuh.gradle.ktlint'
-
 apply from: rootProject.file('generate_build_properties.gradle')
-apply from: rootProject.file('../publishing.gradle')
+
+// Pull from the GROUP property and assign to the project
+group = GROUP
```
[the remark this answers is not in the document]
ZacSweers: This was needed before too, I'm actually not sure why it worked before. The gradle plugin publish plugin needs to be set

JoelWilcox (approved):
:bless:

compiler-utils/build.gradle
```diff
@@ -1,6 +1,10 @@
-apply plugin: 'org.jetbrains.kotlin.jvm'
-apply plugin: 'java-test-fixtures'
-apply from: rootProject.file('publishing.gradle')
+import org.jetbrains.kotlin.gradle.tasks.KotlinCompile
+
+plugins {
+  alias(libs.plugins.kotlin.jvm)
+  alias(libs.plugins.mavenPublish)
+  id 'java-test-fixtures'
```
vRallev, line 6: In general with Gradle, this is painful to read 😅

compiler/generate_build_properties.gradle
```diff
@@ -34,3 +34,8 @@ def generateBuildProperties = project.tasks.register('generateBuildProperties')
 tasks.named("compileKotlin").configure {
   it.dependsOn generateBuildProperties
 }
+
+// So Gradle stops complaining about the missing dependency
+tasks.named("runKtlintCheckOverTestSourceSet").configure {
+  it.dependsOn generateBuildProperties
```
vRallev, line 40: I believe a `mustRunAfter` is good enough here. The benefit is that it doesn't make the other task run unnecessarily, but still teaches Gradle about the ordering.

vRallev (approved):

PR #679, opened 2023-01-19
Fix duplicate generated binding method names.

vRallev:
If a class contributes multiple bindings and the bound types have the same short name, then Anvil would generate methods with duplicate names that clash in the end.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -210,15 +210,21 @@ internal class BindingModuleGenerator(
           .toList()
       }
 
-      val generatedMethods = getContributedBindingClasses(
-        collectedClasses = contributedBindingClasses,
-        annotationFqName = contributesBindingFqName,
-        isMultibinding = false
-      ) + getContributedBindingClasses(
-        collectedClasses = contributedMultibindingClasses,
-        annotationFqName = contributesMultibindingFqName,
-        isMultibinding = true
-      )
+      val generatedMethods =
+        getContributedBindingClasses(
+          collectedClasses = contributedBindingClasses,
+          annotationFqName = contributesBindingFqName,
+          isMultibinding = false
+        )
+          .plus(
+            getContributedBindingClasses(
+              collectedClasses = contributedMultibindingClasses,
+              annotationFqName = contributesMultibindingFqName,
+              isMultibinding = true
+            )
+          )
+          .renameDuplicateFunctions()
+          .sorted()
 
       val content = daggerModuleContent(
         scopes = scopes,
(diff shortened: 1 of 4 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
```diff
@@ -210,15 +210,21 @@ internal class BindingModuleGenerator(
...
+        )
+          .plus(
+            getContributedBindingClasses(
+              collectedClasses = contributedMultibindingClasses,
+              annotationFqName = contributesMultibindingFqName,
+              isMultibinding = true
+            )
+          )
+          .renameDuplicateFunctions()
+          .sorted()
```
vRallev, line 227: I think it was a problem that they never were sorted.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
```diff
@@ -2420,7 +2420,7 @@ public final class DaggerComponentInterface implements ComponentInterface {
       """
     ) {
       val factoryClass = componentInterface.anvilModule
-        .moduleFactoryClass("provideComSquareupTestContributingObjectParentInterface")
+        .moduleFactoryClass("provideParentInterface")
```
vRallev, line 2423: The new logic produces much easier to read method names.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
JoelWilcox: > I think it was a problem that they never were sorted.

Can you clarify why that was a problem? Vs just renaming the duplicate functions

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
JoelWilcox: Hmm I was going to leave a comment about avoiding the duplicate name indices by basically using the full package as part of the name and now see that is what we were already doing. Could you clarify the original use-case where you found the issue? Was it two classes/interfaces that also shared the same package?

compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
vRallev: For stable caching it's important to produce the same output the for the same input. I'm honestly not sure right now if it has an impact in practice, but it's always better to have a stable sort.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/ProvidesMethodFactoryGeneratorTest.kt
vRallev: The old method name was `bind${fqName}${boundType.simpleName}`. Now if the bound type uses the same simple name, but different package, then the methods would clash and result in errors. I considered using the FqName for the bound type and it would work, but it looks ugly and is very hard to read. This is way is much better in my opinion.

JoelWilcox (approved):

PR #683, opened 2023-01-21
Add option to change the JVM target when testing code generators with the custom `AnvilCompilation` class

### LINKED ISSUE #682: Allow configuration of KotlinCompiler used in AnvilCompilation to set jvm target
Hi there,
I just upgraded a project to target java 11 and I have a unit test for a custom annotation which uses the internal `compileAnvil` method.

https://github.com/square/anvil/blob/4b925f7b50e330acafa396e239a03bcf7858671a/compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt#L193
The JVM target is hard-coded to target 1.8 and when I run the test it fails with the following error:

`Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Please specify proper '-jvm-target' option`

Would it be possible to allow the user to configure the JVM target for the `compileAnvil` method?

vRallev:
Fixes #682

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@
 
 ### Custom Code Generator
 
+- Add option to change the JVM target when testing code generators with the custom `AnvilCompilation` class, see #682.
+
 # [2.4.4] - 2023-01-12
 
 ### Added
--- a/compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
+++ b/compiler-utils/src/testFixtures/java/com/squareup/anvil/compiler/internal/testing/AnvilCompilation.kt
@@ -137,6 +137,11 @@ public class AnvilCompilation internal constructor(
     kotlinCompilation.addPreviousResultToClasspath(result)
   }
 
+  public fun jvmTarget(jvmTarget: JvmTarget) {
+    checkNotCompiled()
+    kotlinCompilation.jvmTarget = jvmTarget.description
+  }
+
   /**
    * Returns an Anvil-generated file with the given [packageName] and [fileName] from its expected
    * path.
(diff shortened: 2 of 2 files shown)


RBusarow:
The system property `java.vm.specification.version` holds the version strings like "1.8", "11", etc.

So, it's possible to set the target's default based upon the runtime, instead of the hard-coded 1.8.

```kotlin
companion object {
  public operator fun invoke(): AnvilCompilation {
    return AnvilCompilation(
      KotlinCompilation().apply {
        // ...
        jvmTarget = JvmTarget.fromString(System.getProperty("java.vm.specification.version"))
        // ...
      }
    )
  }
}
```

vRallev:
> The system property `java.vm.specification.version` holds the version strings like "1.8", "11", etc.
> 
> So, it's possible to set the target's default based upon the runtime, instead of the hard-coded 1.8.
> 
> ```kotlin
> companion object {
>   public operator fun invoke(): AnvilCompilation {
>     return AnvilCompilation(
>       KotlinCompilation().apply {
>         // ...
>         jvmTarget = JvmTarget.fromString(System.getProperty("java.vm.specification.version"))
>         // ...
>       }
>     )
>   }
> }
> ```

I implemented this locally, and then discarded all changes. There's also `Runtime.version().feature()` now. 

The reason why I discarded it is that many projects use newer JDK versions at build time and target older versions at runtime. Most Android projects do that. Anvil itself is doing that. It also opens up the door for different behavior locally and during CI. I changed my mind and believe being explicit is the way to go.

RBusarow:
> It also opens up the door for different behavior locally and during CI.

I was kind of on the fence, but I hadn't considered this one.  I'm convinced. :)

RBusarow (approved):

vRallev:
We think alike, I had the same idea 😅

PR #690, opened 2023-02-14
Fix version overrides.

vRallev:
The correct key wasn't used to override the version when the name uses camel case.

### DIFF
--- a/gradle-plugin/settings.gradle
+++ b/gradle-plugin/settings.gradle
@@ -19,7 +19,7 @@ dependencyResolutionManagement {
       for (Map.Entry<String, String> entry : overrides.entrySet()) {
         String key = entry.getKey()
         String value = entry.getValue()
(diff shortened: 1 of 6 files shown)


gradle-plugin/settings.gradle
```diff
@@ -19,7 +19,7 @@ dependencyResolutionManagement {
       for (Map.Entry<String, String> entry : overrides.entrySet()) {
         String key = entry.getKey()
         String value = entry.getValue()
-        String catalogKey = key.substring("override_".length()).toLowerCase(Locale.US)
+        String catalogKey = key.substring("override_".length())
```
vRallev, line 22: @ZacSweers Do you remember why you did this? This was causing issues.

ZacSweers (approved):

gradle-plugin/settings.gradle
ZacSweers: I don't remember

JoelWilcox (approved):

PR #720, opened 2023-06-20
Only clear the output dir the first round

### LINKED ISSUE #693: Compiler doesn't pick incremental changes to multibindings in some circumstances
When:

* We have classes with `@ContributesMultibinding` annotation in a library module
* Said library module includes Jetpack Compose compiler
* [New incremental compilation](https://blog.jetbrains.com/kotlin/2022/07/a-new-approach-to-incremental-compilation-in-kotlin/) is enabled (`kotlin.incremental.useClasspathSnapshot=true` in `gradle.properties`)

Changes on those annotations may not get picked up when making incremental builds.

Steps to reproduce:
1. Download and open [Project.zip](https://github.com/square/anvil/files/10837039/Project.zip)
2. Run project
3. Check logcat
4. Note that logs will print two entries (as it should)
5. Open LibraryMultibinds.kt
6. Comment out `@ContributesMultibinding` annotation
7. Run again and check logcat
8. Logs print two entries again, even though we have commented one

ZacSweers:
Hopeefully resolves #693

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
@@ -27,6 +27,7 @@ internal class CodeGenerationExtension(
 ) : AnalysisHandlerExtension {
 
   private var didRecompile = false
+  private var compilationRound = 0
 
   private val codeGenerators = codeGenerators
     .onEach {
@@ -59,6 +60,7 @@ internal class CodeGenerationExtension(
     bindingTrace: BindingTrace,
     files: Collection<KtFile>
   ): AnalysisResult? {
+    compilationRound++
     if (didRecompile) return null
     didRecompile = true
 
@@ -70,12 +72,16 @@ internal class CodeGenerationExtension(
 
     val anvilContext = commandLineOptions.toAnvilContext(anvilModule)
 
-    codeGenDir.listFiles()
-      ?.forEach {
-        check(it.deleteRecursively()) {
-          "Could not clean file: $it"
+    if (compilationRound == 1) {
+      // Only clear the output directory the first round
+      // https://github.com/square/anvil/issues/693#issuecomment-1585545532
+      codeGenDir.listFiles()
+        ?.forEach {
+          check(it.deleteRecursively()) {
+            "Could not clean file: $it"
+          }
         }
-      }
+    }
 
     val generatedFiles = mutableMapOf<String, GeneratedFile>()
 


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -70,12 +72,16 @@ internal class CodeGenerationExtension(
 
     val anvilContext = commandLineOptions.toAnvilContext(anvilModule)
 
-    codeGenDir.listFiles()
-      ?.forEach {
-        check(it.deleteRecursively()) {
-          "Could not clean file: $it"
+    if (compilationRound == 1) {
```
JoelWilcox, line 75: If I'm reading correctly, wouldn't this still behave the same as the existing behavior? Because we already have the `didRecompile` flag that should prevent us getting to this point after the first invocation of `analysisCompleted`

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -59,6 +60,7 @@ internal class CodeGenerationExtension(
     bindingTrace: BindingTrace,
     files: Collection<KtFile>
   ): AnalysisResult? {
+    compilationRound++
     if (didRecompile) return null
     didRecompile = true
```
vRallev, line 65: I'm now confused. This boolean should make sure that we never delete files in the 2nd round.

ZacSweers: I think this was doing the inverse - it would always wipe the 2nd+ round when this seeks to only wipe the first round it processes. Should we remove the didRecompile? I wasn't quite sure how it worked

vRallev: Let me clarify the use case. We need to wipe the directory to avoid any stale files. The compiler APIs require a directory as additional root, it would be easier if we could just return a set of files. 

What I've seen in the past is that `analysisCompleted` gets called multiple time with the same inputs. That's why I've added the `didCompile` boolean flag. It's `false` by default, which will make all the code run during the first call, but not any further calls. Therefore we only wipe the directory per compilation only once and then not again. This counter isn't necessary.

What I understood from the ticket is that the inputs `analysisCompleted` are changing now and we therefore should run our code, but not wipe the directory again. Is this correct?

ZacSweers: I'm not sure now. @hungvietnguyen could you offer insight?

ZacSweers: Hmm, what if it worked like this:
1. Anvil always clears the output dir first before the KotlinCompile task runs (via doFirst)
2. Anvil creates the output dir the first time it runs, never clears it during the KotlinCompile task. If the output dir exists, assume that anvil has already run a previous round.

That sort of solves the "has a previous round run" problem by never trying to intligently determine rounds

hungvietnguyen: > 1. Anvil always clears the output dir first before the KotlinCompile task runs (via doFirst)

It's okay to clean up the output in a non-incremental run, but in an incremental run, the output cannot be cleaned up. The key idea of incrementality is that we partially re-generate *only some* of the output. If we delete all of the output, then that means we'll need to re-generate all of them, which means we don't have incrementality.

ZacSweers: Hmm. Does anvil need to clear outputs at all? If kotlincompile Will do that for us? Otherwise I kind of would expect anything generating code during AnalydisHandlerExtension would be incompatible with IC

hungvietnguyen: It depends:
  - In a non-incremental run, Gradle / KGP should clean the outputs, it's optional for Anvil to do it.
  - In an incremental run, no one should clean the outputs.

> Otherwise I kind of would expect anything generating code during AnalydisHandlerExtension would be incompatible with IC

If Anvil output files have 1-1 correspondence with Kotlin source file, then I think it can be made compatible (similar to [isolating annotation processors](https://docs.gradle.org/current/userguide/java_plugin.html#isolating_annotation_processors)). Otherwise, it needs more investigation.

PR #721, opened 2023-07-06
Upgrade to Kotlin 1.9.0

JoelWilcox:
~Debating whether to set-up for a dual release to still support 1.8.x for a bit, but getting this up to verify there are no new issues since the RC release (#719).~

This PR:
- Ugprades to Kotlin 1.9.0
- Updates our release process to publish parallel releases for Kotlin 1.9.x and 1.8.x
- Upgrades from Kotlin 1.8.21 -> 1.8.22 for the 1.8.x releases
- Removes a `kotlinPreview` flag that we have not needed or used in a while
- Upgrades a couple other misc dependencies


JoelWilcox:
> I'd kinda vote a dual release since we're weeks aways from a stable studio release that supports kotlin 1.9

👍 Yea I agree, and there are no source/API changes necessary so supporting both doesn't add any extra complexity right now.

ZacSweers (approved):

msfjarvis:
Both [Renovate](https://github.com/msfjarvis/whetstone/commit/7a95154bf8899a68d7f48b8acccc9388b200c73b) and AGP try to upgrade folks to the 1.8 build so you might get some reports about it 🙃 

```
/home/runner/work/compose-lobsters/compose-lobsters/gradle/libs.versions.toml:100: Error: A newer version of com.squareup.anvil than 2.4.7 is available: 2.4.7-1-8 [NewerVersionAvailable]
anvil = "com.squareup.anvil:2.4.7"
        ~~~~~~~~~~~~~~~~~~~~~~~~~~

   Explanation for issues of type "NewerVersionAvailable":
   This detector checks with a central repository to see if there are newer
   versions available for the dependencies used by this project. This is
   similar to the GradleDependency check, which checks for newer versions
   available in the Android SDK tools and libraries, but this works with any
   MavenCentral dependency, and connects to the library every time, which
   makes it more flexible but also much slower.

```

PR #725, opened 2023-07-16
add optional annotations artifact with SingleIn and ForScope

### LINKED ISSUE #692: Provide SingleIn and AppScope as part of Anvil
It might not fit 100% but I think it would help with standardization in the community instead of each project defining their slightly different own variant of this. Especially `SingleIn` fits really well with Anvil's other annotations from a code style perspective.

gabrielittner:
Closes #692

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -106,6 +106,7 @@ subprojects {
   configurations.configureEach {
     resolutionStrategy.dependencySubstitution {
       substitute module("com.squareup.anvil:annotations") using project(':annotations')
+      substitute module("com.squareup.anvil:annotations-optional") using project(':annotations-optional')
       substitute module("com.squareup.anvil:compiler") using project(':compiler')
       substitute module("com.squareup.anvil:compiler-api") using project(':compiler-api')
       substitute module("com.squareup.anvil:compiler-utils") using project(':compiler-utils')
--- a/build.gradle
+++ b/build.gradle
@@ -94,6 +94,7 @@ subprojects {
       freeCompilerArgs.add("-opt-in=kotlin.RequiresOptIn")
 
       boolean optInExperimental = project.name != "annotations"
+          && project.name != "annotations-optional"
           && project.name != "scopes"
           && project.name != "dagger-factories-only"
       if (optInExperimental) {
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
@@ -51,7 +51,7 @@ public abstract class AnvilExtension @Inject constructor(objects: ObjectFactory)
     .convention(false)
 
   /**
-   * Add the `annotations-optional` artifact as a dependency to make annotations like `@SingleIn` and `@ForScope` 
+   * Add the `annotations-optional` artifact as a dependency to make annotations like `@SingleIn` and `@ForScope`
    * available to use. These are annotations that are not strictly required but which we've found to be helpful with managing larger dependency graphs.
    */
   public val addOptionalAnnotations: Property<Boolean> = objects.property(Boolean::class.java)
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
@@ -50,6 +50,13 @@ public abstract class AnvilExtension @Inject constructor(objects: ObjectFactory)
   public val syncGeneratedSources: Property<Boolean> = objects.property(Boolean::class.java)
     .convention(false)
 
+  /**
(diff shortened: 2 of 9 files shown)


ZacSweers (approved):

kingsleyadio:
(opinion): I think optional already has a meaning in Dagger world, which might make this naming a little confusing.
I propose changing for example, to something like `annotations-extended`. Other than that, love the reasoning behind including these directly in Anvil

gabrielittner:
I mostly went with the suggestion from the issue but don't really have a strong opinion on the artifact name. Happy to update the pr with something else

vRallev:
It would be great to update the commit message with a better description. I'm okay with `optional` as name, but what I'd like to see is a DSL option that adds this dependency for you, similar to how the Gradle plugin adds the annotations artifact by default. This can be a separate PR, but this one shouldn't resolve #692.

ZacSweers:
I think optional is a bad name because dagger has specialized support for an "optional" semantic via `@BindsOptionalOf`. I think scope-annotations or something would be better.

Also - does this need to be configurable via the DSL since it technically doesn't really have anything to do with anvil? 🤔

vRallev:
If it doesn't have anything to do with Anvil, then it shouldn't be part of the repository altogether. By allowing it to be configured through the DSL we don't have to deal with Maven coordinates and it's more discoverable IMO.

JoelWilcox:
I agree that having DSL support for configuring/adding this dependency would be nice. @gabrielittner up to you if you want to take a stab at that, if not I'm happy to add it separately.

JoelWilcox:
> I think optional is a bad name because dagger has specialized support for an "optional" semantic via @BindsOptionalOf. I think scope-annotations or something would be better.

Hmm this is a fair point, although I feel there aren't a lot of other name options that still accurately capture it as being a set of optional annotations while still being simple. I think something like `scope-annotations` could be confusing in other ways since the main annotations inherently deal with scopes too, and it could be limiting in the future if there are other optional opinionated annotations we want to add (which maybe don't directly deal with scopes).

At the moment I'd lean more towards some potential initial confusion with Dagger's concept of `Optional` since it should quickly become obvious it's unrelated when looking at the documentation for the artifact or directly at the included annotations. Only other name that comes to mind so far as a middle-ground would be something like `annotations-extra`.

gabrielittner:
@vRallev I've rebased and updated the commit message.

The DSL now has `addOptionalAnnotations` to add the dependency. It's also on `VariantFilter` to match the other options but if you think that's too much we can remove it from there and just directly use the main property from the extension.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -50,6 +50,13 @@ public abstract class AnvilExtension @Inject constructor(objects: ObjectFactory)
   public val syncGeneratedSources: Property<Boolean> = objects.property(Boolean::class.java)
     .convention(false)
 
+  /**
+   * Add the `annotations-optional` artifact as a dependency to make `SingleIn` and `ForScope` 
+   * available to use.
```
JoelWilcox, line 55: Minor re-wording suggestion:
```suggestion
   * Add the `annotations-optional` artifact as a dependency to make annotations like `@SingleIn` and `@ForScope` 
   * available to use. These are annotations that are not strictly required but which we've found to be helpful with managing larger dependency graphs.
```

gradle-plugin/src/main/java/com/squareup/anvil/plugin/VariantFilter.kt
```diff
@@ -36,6 +36,12 @@ public interface VariantFilter : Named {
    * IDE. This can be useful in debugging and is disabled by default.
    */
   public var syncGeneratedSources: Boolean
+
+  /**
+   * Add the `annotations-optional` artifact as a dependency to make `SingleIn` and `ForScope` 
+   * available to use.
```
JoelWilcox, line 42:
```suggestion
   * Add the `annotations-optional` artifact as a dependency. The default
   * value comes from the [AnvilExtension]. See [AnvilExtension.addOptionalAnnotations] for more
   * details.
```

build.gradle
```diff
@@ -94,6 +94,7 @@ subprojects {
       freeCompilerArgs.add("-opt-in=kotlin.RequiresOptIn")
 
       boolean optInExperimental = project.name != "annotations"
+          && project.name != "annotations-optional"
           && project.name != "scopes"
           && project.name != "dagger-factories-only"
       if (optInExperimental) {
```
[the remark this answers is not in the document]
gabrielittner: Added it [6f175ec](https://github.com/square/anvil/pull/725/commits/6f175ec5a64e55dbceab6c8f8345f7202b35288d)

JoelWilcox (approved):
Lgtm! Thanks for adding this

vRallev (approved):

matejdro:
Could a mention of this be added to the `README`? Right now this is not very discoverable.

PR #726, opened 2023-07-21
Allow contributing bindings with generic supertypes

### LINKED ISSUE #694: Allow type parameters binding, with star type added
When attempting to multibind classes with type parameters:

```
@ContributesMultibinding(SomeScope::class)
class MyClass: SomeInterface<String>
```

Anvil will complain that `Type parameters in bindings are not supported. This binding needs to be contributed in a Dagger module manually.`

I understand that this is a problem, because Anvil can't know what type parameter should be used on bound type (there is no way to specify that in annotation). But, maybe we could just support the most basic version?

We could just add star to all type bindings, so above example would get bound to the `SomeInterface<*>` type. It does not cover many use cases, but I think it's still better than not supporting this at all. Any other non-star cases can still use module like now.

I'm willing to submit PR for this, if desired.

matejdro:
Instead of giving up, whenever user attempts to `@Contributes(Multi)Binding` on a generic parent interface, it will just bind to a variant of the parent interface with generic parameters filled as stars. 

It does not cover all use cases, but it covers more use cases than the old solution. And user can still use regular Module to achieve the rest of the use cases, so this seems like a win win in my book.

This fixes #694.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/BindingModuleGenerator.kt
@@ -28,6 +28,8 @@ import com.squareup.kotlinpoet.AnnotationSpec
 import com.squareup.kotlinpoet.FileSpec
 import com.squareup.kotlinpoet.FunSpec
 import com.squareup.kotlinpoet.KModifier.ABSTRACT
+import com.squareup.kotlinpoet.ParameterizedTypeName.Companion.parameterizedBy
+import com.squareup.kotlinpoet.STAR
 import com.squareup.kotlinpoet.TypeSpec
 import dagger.Binds
 import dagger.Module
@@ -354,6 +356,15 @@ private fun ContributedBinding.toGeneratedMethod(
     }
   }
 
+  val boundTypeWithStarGenerics = boundType.asClassName()
+    .let {
+      if (boundType.typeParameters.isEmpty()) {
+        it
+      } else {
+        it.parameterizedBy(boundType.typeParameters.map { STAR })
+      }
+    }
+
   return if (contributedClass.isObject()) {
     ProviderMethod(
       spec = FunSpec.builder(name = "provide$methodNameSuffix")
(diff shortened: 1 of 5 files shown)


gabrielittner:
For multibinding using `*` definitely makes sense.

However I'm wondering whether for regular bindings using the actual type parameter would make more sense. For example if I bind [SQLDelight's ColumnAdpater](https://cashapp.github.io/sqldelight/2.0.0-rc01/2.x/runtime/app.cash.sqldelight/-column-adapter/index.html), I'd probably would want the bound type for `class InstantColumnAdapter : ColumnAdapter<String, Instant>` to be `ColumnAdapter<String, Instant>`.

vRallev:
This was debated over and over again and there are other tickets discussing this. We didn't want to support this, because it adds inconsistency with `boundType`. Imagine this example:
```
@ContributesBinding(AppScope::class, boundType = CharSequence::class)
@ContributesMultbinding(AppScope::class)
object MyClass : List<String>, CharSequence { .. }
```
Which super type should be used is unclear, you'd need to use `boundType` similar to `ContributesBinding` in the sample. But `boundType` doesn't allow define the generic type. 

You can always support your use case with a custom code generator.

matejdro:
For the `@ContributesBinding` issue, what if we just take the actual parent type instead of `*`? So in your example `@ContributesBinding` would provide `List<String>` instead of `ColumnAdapter<String, Instant>`. However, `@ContributesMultibinding` would still bind star bindings, so it would bind `ColumnAdapter<*, *>`.

As for the`boundType` issue, what if you just need to specify the type and type parameter are inferred from actual paremters? So, in your case, you would enter `@ContributesMultbinding(AppScope::class, boundType = List::class)` and Anvil would know that you meant `List<String>`, since `MyClass` extends it.

blakelee:
This has come up a few times on my team and we'd really like to have this functionality. Even if it doesn't fix every issue, it would be nice to at least do something, even if it's only generic supertypes for now.

ramzyabushaaban:
Any solution for this use case?

matejdro:
I can look into solving conflicts and getting this into mergeable state if I get some feedback from Anvil team that this is actually something they want to get merged. They have been pretty silent so far.

creativedrewy:
Allow me to +1 getting this merged! Would be really useful in my current project.

PR #727, opened 2023-07-21
Dagger factories source sets

matejdro:
Fixes #178

### DIFF
--- a/README.md
+++ b/README.md
@@ -237,6 +237,17 @@ sources either.
 When you enable this feature, don't forget to remove the Dagger annotation processor. You should
 keep all other dependencies.
 
+If you want to only enable Dagger factory generation on some source sets 
+(such as only in main source set or only in tests), you can add those source sets
+to the whitelist:
+
+```groovy
+anvil {
+  generateDaggerFactoriesSourceSetWhitelist.add("main")
+}
+```
+
+
 ## Extending Anvil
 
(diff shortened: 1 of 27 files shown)


ZacSweers (commented):
Three high level comments
- Absolute paths are likely to cause gradle caching issues as they cannot be cached across machines or even different repo installations
- With KSP, this doesn't become possible I think?
- Instead of whitelist/blacklist, prefer allowlist/blocklist as both more descriptive names and more inclusive naming.

matejdro:
Huh you are right, about build cache I totally forgot about that. I just noticed that there is a FilesSubpluginOption (which is already used for build dir argument, that I totally missed). I think switching to that should resolve this issue?

To be honest, I have no idea how KSP works, so I'm not sure if it would work or not. Does KSP not give us a list of files to work with?

I can change the PR to allowlist/blocklist, if desired.

matejdro:
Doh, I have managed to find much simpler solution that does not even touch the compiler and thus sidestepping the absolute path issue entirely (possibly also sidesteps KSP problem?).

I have also renamed all whitelist references to allowlist.

JoelWilcox:
> With KSP, this doesn't become possible I think?

@ZacSweers Are you able to provide more details on this? Is it related to us no longer manually managing the output dir/location?

PR #730, opened 2023-07-27
Log an error when K2 is enabled until we support it

JoelWilcox:
Compiling a project with Anvil and K2 enabled would fail anyway, this just makes it more obvious that it's a known incompatibility for now.

### DIFF
--- a/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
+++ b/gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
@@ -20,6 +20,8 @@ import org.gradle.api.artifacts.Configuration
 import org.gradle.api.provider.Provider
 import org.gradle.api.tasks.TaskContainer
 import org.gradle.api.tasks.TaskProvider
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_1_9
+import org.jetbrains.kotlin.gradle.dsl.KotlinVersion.KOTLIN_2_0
 import org.jetbrains.kotlin.gradle.internal.KaptGenerateStubsTask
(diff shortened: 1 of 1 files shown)


compiler/build.gradle
```diff
@@ -37,6 +38,10 @@ tasks.withType(KotlinCompile).configureEach {
     freeCompilerArgs.add("-Xjvm-default=all")
     freeCompilerArgs.add("-opt-in=org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi")
   }
+  if (compilerOptions.useK2.get() || compilerOptions.languageVersion.getOrElse(
+      KotlinVersion.KOTLIN_1_9) >= KotlinVersion.KOTLIN_2_0) {
+    throw new IllegalStateException("Anvil is currently incompatible with the K2 compiler.")
```
vRallev, line 43: I think it's worth adding a link to a ticket here for people to follow.

vRallev (approved):

ZacSweers (commented):
Can there be an opt-out for this? For better or worse, it does actually help with K2 prep to have anvil be inert while testing other things

ZacSweers (commented):
Also, doesn't this only affect the anvil repo? Or was that the goal? Maybe I misunderstood but not sure why the anvil repo needs this check

JoelWilcox:
> Can there be an opt-out for this? For better or worse, it does actually help with K2 prep to have anvil be inert while testing other things

Discussed offline -- this was a misunderstanding on my part around where failures were already occurring vs where Anvil was effectively just not running. Updated to only log an error message

ZacSweers (approved):

JoelWilcox:
> Also, doesn't this only affect the anvil repo? Or was that the goal? Maybe I misunderstood but not sure why the anvil repo needs this check

No you're right, thanks for catching that. Updated to apply at the plugin level

PR #732, opened 2023-07-28
Update kotlin compile testing to 0.3.1

ZacSweers:
https://github.com/ZacSweers/kotlin-compile-testing/releases/tag/0.3.1

https://github.com/ZacSweers/kotlin-compile-testing/releases/tag/0.3.0


ZacSweers:
I don't see how this is gonna work in a dual release setup 🤔

JoelWilcox:
> I don't see how this is gonna work in a dual release setup 🤔

What do you mean by this? Are you expecting these changes to have issues on 1.8.22? (So far looks like all the tests passed at least)

ZacSweers:
The KCT dependency update requires kotlin 1.9.0 and has some breaking changes, so I'm wondering if the 1.8.22 tests are silently getting updated to kotlin 1.9.0 due to the transitive library bump

JoelWilcox:
Hmm interesting, thinking about what we could do to verify that's not happening / ensure something like that never happens even for other changes 🤔. Possibly adding gradle constraints?

JoelWilcox:
Tested in https://github.com/square/anvil/pull/737 and verified failures, so we should be good for this

JoelWilcox (approved):

PR #734, opened 2023-07-30
Update some impl IR APIs to more modern alternatives

ZacSweers:
Just a little cleanup to use more stable IR factory functions rather than direct impl types

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
@@ -15,21 +15,21 @@ import com.squareup.anvil.compiler.internal.safePackageString
 import dagger.Module
 import org.jetbrains.kotlin.backend.common.extensions.IrGenerationExtension
 import org.jetbrains.kotlin.backend.common.extensions.IrPluginContext
+import org.jetbrains.kotlin.backend.common.lower.createIrBuilder
+import org.jetbrains.kotlin.backend.jvm.ir.kClassReference
 import org.jetbrains.kotlin.ir.IrStatement
-import org.jetbrains.kotlin.ir.UNDEFINED_OFFSET
+import org.jetbrains.kotlin.ir.builders.IrBuilderWithScope
+import org.jetbrains.kotlin.ir.builders.irCallConstructor
+import org.jetbrains.kotlin.ir.builders.irVararg
 import org.jetbrains.kotlin.ir.declarations.IrClass
 import org.jetbrains.kotlin.ir.declarations.IrModuleFragment
+import org.jetbrains.kotlin.ir.expressions.IrExpression
 import org.jetbrains.kotlin.ir.expressions.IrVararg
-import org.jetbrains.kotlin.ir.expressions.impl.IrClassReferenceImpl
-import org.jetbrains.kotlin.ir.expressions.impl.IrConstructorCallImpl
-import org.jetbrains.kotlin.ir.expressions.impl.IrVarargImpl
-import org.jetbrains.kotlin.ir.types.defaultType
-import org.jetbrains.kotlin.ir.types.impl.IrDynamicTypeImpl
+import org.jetbrains.kotlin.ir.types.starProjectedType
 import org.jetbrains.kotlin.ir.util.defaultType
 import org.jetbrains.kotlin.ir.visitors.IrElementTransformerVoid
 import org.jetbrains.kotlin.name.FqName
 import org.jetbrains.kotlin.name.Name
-import org.jetbrains.kotlin.types.Variance.INVARIANT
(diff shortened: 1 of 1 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -253,30 +259,20 @@ internal class ModuleMergerIr(
...
         putValueArgument(
           index = 0,
-          valueArgument = IrVarargImpl(
-            startOffset = UNDEFINED_OFFSET,
-            endOffset = UNDEFINED_OFFSET,
-            type = pluginContext.symbols.array.defaultType,
-            varargElementType = IrDynamicTypeImpl(null, emptyList(), INVARIANT),
-            elements = contributedModules
+          valueArgument = irVararg(
+            elementType = pluginContext.irBuiltIns.kClassClass.starProjectedType,
```
ZacSweers, line 272: This is a change from before, but I think this is more correct (`Array<KClass<*>>`) as `dynamic` is only really intended for JS targets iirc

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -53,15 +53,21 @@ internal class ModuleMergerIr(
             .findAll(mergeComponentFqName, mergeSubcomponentFqName, mergeModulesFqName)
             .ifEmpty { return super.visitClass(declaration) }
 
-          generateDaggerAnnotation(annotations, moduleFragment, pluginContext, declarationReference)
+          pluginContext.irBuiltIns.createIrBuilder(declaration.symbol)
+            .generateDaggerAnnotation(
```
ZacSweers, line 57: Creating an IR builder exposes more factory extension functions, including some of the ones used below

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
```diff
@@ -291,12 +287,11 @@ internal class ModuleMergerIr(
...
-              startOffset = 0,
-              endOffset = 0,
-              type = varargArguments[0].type,
-              varargElementType = varargArguments[0].varargElementType,
-              elements = varargArguments.flatMap { it.elements }
+            valueArgument = irVararg(
+              elementType = varargArguments[0].varargElementType,
+              // These are always IrExpression instances too
+              values = varargArguments.flatMap { it.elements }
+                .filterIsInstance<IrExpression>()
```
JoelWilcox, line 294: Are we sure this is guaranteed? Wondering why the extension function enforces `IrExpressions` when the impl class doesn't

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
JoelWilcox: What do you mean by this being a change from before?

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
JoelWilcox: Do we need to add any followup `build()` equivalent for the IrBuilder? Seems like the answer is no but just wanted to verify

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
ZacSweers: Yeah if you look at the impl that's all it allows. My guess is it's a bit of a marker interface tomfoolery

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
ZacSweers: Before it was surprisingly the kotlin dynamic type, which I suspect was for lack of a better star alternative: https://kotlinlang.org/docs/dynamic-type.html

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
ZacSweers: Not as far as I know, it's more of a context thing than a true builder

compiler/src/main/java/com/squareup/anvil/compiler/ModuleMergerIr.kt
JoelWilcox: `The dynamic type is not supported in code targeting the JVM.` - mildly concerning :D, apparently it works though 😆. Good catch.

JoelWilcox (approved):

PR #736, opened 2023-07-30
Support explicit API mode for complex map keys

### LINKED ISSUE #735: Annotation with `@MapKey(unwrapValue = false)` ("complex map key") doesn't support explicit API mode
The source that `MapKeyCreatorGenerator` generates is missing an explicit return type, giving this error message:
```
Return type must be specified in explicit API mode
```

I'm interested in contributing a fix (along with additional tests) and will be raising a PR in a few minutes. 😃

bddckr:
Resolves #735

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MapKeyCreatorGenerator.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MapKeyCreatorGenerator.kt
@@ -178,6 +178,7 @@ internal class MapKeyCreatorGenerator : PrivateCodeGenerator() {
         className,
         properties.entries.map { it.value.callExpression }.joinToCode()
       )
+      .returns(className)
       .build()
   }
 }
--- a/integration-tests/library/build.gradle
+++ b/integration-tests/library/build.gradle
@@ -12,6 +12,10 @@ if (libs.versions.config.generateDaggerFactoriesWithAnvil.get().toBoolean()) {
 
   dependencies {
     kapt libs.dagger2.compiler
+
+    // Necessary because this is what dagger uses when it runs to support instantiating annotations at runtime
+    implementation libs.auto.value.annotations
+    kapt libs.auto.value.processor
   }
 }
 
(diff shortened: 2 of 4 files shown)


integration-tests/library/build.gradle
```diff
@@ -12,6 +12,10 @@ if (libs.versions.config.generateDaggerFactoriesWithAnvil.get().toBoolean()) {
 
   dependencies {
     kapt libs.dagger2.compiler
+
+    // Necessary because this is what dagger uses when it runs to support instantiating annotations at runtime
+    implementation libs.auto.value.annotations
```
ZacSweers, line 17: I'm curious why this is needed now and not before?

bddckr: tl;dr: [Dagger only creates the annotation creator class for the map key when `@MapKey(unwrapValue = false)` is used.](https://github.com/google/dagger/blob/2067f51a476b1f60c3aa94812af0f8f0db661521/java/dagger/internal/codegen/processingstep/MapKeyProcessingStep.java#L68-L69)

As seen in that link, when `unwrapValue = true`, it uses `UnwrappedMapKeyGenerator` instead, which doesn't use `@AutoAnnotation`.

---

Just to clarify:

Without
```kotlin
implementation libs.auto.value.annotations
kapt libs.auto.value.processor
```

we get
```
> Task :integration-tests:library:kaptKotlin
C:\Projects\anvil\integration-tests\library\build\tmp\kapt3\stubs\main\com\squareup\anvil\test\WrappedBindingKey.java:6: error: @AutoAnnotation is a necessary dependency if @MapKey(unwrapValue = false). Add a dependency for the annotation, "com.google.auto.value:auto-value-annotations:<current version>", and the annotation processor, "com.google.auto.value:auto-value:<current version>"
public abstract @interface WrappedBindingKey {
                 ^
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/MapKeyCreatorGenerator.kt
```diff
@@ -178,6 +178,7 @@ internal class MapKeyCreatorGenerator : PrivateCodeGenerator() {
         className,
         properties.entries.map { it.value.callExpression }.joinToCode()
       )
+      .returns(className)
```
JoelWilcox, line 181: Nice, coincidentally this is also needed for upgrading to newer KotlinPoet versions

integration-tests/library/build.gradle
JoelWilcox: Thanks for sharing the extra context :+1:

JoelWilcox (approved):

PR #740, opened 2023-08-02
Implement KSP support for ContributesMultibinding

ZacSweers:
Fairly straightforward, but added support for a `SimpleSymbolProcessor` for use in tests.


compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGeneratorTest.kt
```diff
@@ -325,32 +353,47 @@ class ContributesMultibindingGeneratorTest {
...
+        val codeGenerator = simpleCodeGenerator { clazz ->
+          clazz
+            .takeIf { it.isAnnotatedWith(mergeComponentFqName) }
+            ?.let { stubContentToGenerate }
+        }
+        AnvilCompilationMode.Embedded(listOf(codeGenerator))
+      }
+      is AnvilCompilationMode.Ksp -> {
+        val processor = simpleSymbolProcessor { resolver ->
+          resolver.getSymbolsWithAnnotation(MergeComponent::class.qualifiedName!!)
```
JoelWilcox, line 391: Is there any reason to use `qualifiedName` over `canonicalName` or vice versa? Just asking since I noticed you used `getSymbolsWithAnnotation(ContributesMultibinding::class.java.canonicalName)` in the code gen

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ksp/SimpleSymbolProcessor.kt
```diff
@@ -0,0 +1,40 @@
...
+  private val mapper: AnvilSymbolProcessor.(resolver: Resolver) -> List<String>
+) : AnvilSymbolProcessor() {
+  override fun processChecked(resolver: Resolver): List<KSAnnotated> {
+    this.mapper(resolver)
+      .map { content ->
+        val (packageName, fileName) = parseSimpleFileContents(content)
+
+        val dependencies = Dependencies(aggregating = false, sources = emptyArray())
+        val file = env.codeGenerator.createNewFile(dependencies, packageName, "$fileName.kt")
+        // Don't use writeTo(file) because that tries to handle directories under the hood
```
JoelWilcox, line 30: What ends up happening if `writeTo` was used here? Would it result in an obvious failure or something more cryptic/delayed? Thinking about if this is a footgun that could bite us elsewhere in the future

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGeneratorTest.kt
ZacSweers: No functional difference, one is just java reflection and the other is kotlin reflection. Thinking we could maybe standardize toward the latter to make it more KMP-friendly in the future?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ksp/SimpleSymbolProcessor.kt
ZacSweers: Under the hood Kotlinpoet tries to make directories for package segments, but KSP does that for us already so we don't want to double-nest. It wouldn't obviously fail, it would just create a duplicate bested package structure. I don't think it's something we need to really guard against as this is the only place we do this custom.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/ContributesMultibindingGeneratorTest.kt
JoelWilcox: 👍 Sounds good to me

ZacSweers:
idk what to do about that windows job

PR #746, opened 2023-09-06
Fix suspend lambdas not being supported

### LINKED ISSUE #745: Suspend lambda parameters are not supported when generating code for AssistedInject/AssistedFactory
With a source file like:

```
data class AssistedService @AssistedInject constructor(
        @Assisted val action: suspend () -> String?
      )
```

We get an error like:

```
    e: Class 'AssistedServiceFactory_Impl' is not abstract and does not implement abstract member public abstract fun create(stringFactory: suspend (Int) -> String): AssistedService defined in com.squareup.test.AssistedServiceFactory
    e: AssistedServiceFactory_Impl.kt:25:3 'create' overrides nothing
```

JoelWilcox:
Fixes #745.

This use-case primarily popped up with assisted parameter and factory generation. We would get an error like: e: Class 'AssistedServiceFactory_Impl' is not abstract and does not implement abstract member public abstract fun create(stringFactory: suspend (Int) -> String): AssistedService defined in com.squareup.test.AssistedServiceFactory e: AssistedServiceFactory_Impl.kt:25:3 'create' overrides nothing

This is because we manually override the generated code for lambdas to address differences between Psi and Descriptors. There is a possibility that we could continue overriding the default generated code for suspend lambdas with modifications to what we currently do, but I found doing so to introduce problematic casting requirements that would spread beyond this isolated area of handling the lambdas themselves. Instead, we can first change to generating the default expected suspend lambda signature to fix this primary use-case, and wait to see if any similar Psi vs Descriptor use-cases pop up later. Thus far I haven't been able to find any reproable test cases where this solution doesn't work.

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedInjectGeneratorTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedInjectGeneratorTest.kt
@@ -96,6 +96,38 @@ public final class AssistedService_Factory {
     }
   }
 
+  @Test fun `a factory class is generated with a suspend lambda assisted parameter`() {
+    compile(
+      """
+      package com.squareup.test
+      
+      import dagger.assisted.Assisted
+      import dagger.assisted.AssistedInject
+      
+      data class AssistedService @AssistedInject constructor(
+        @Assisted val action: suspend () -> String?
+      )
+      """,
+    ) {
+      val factoryClass = assistedService.factoryClass()
+
+      val constructor = factoryClass.declaredConstructors.single()
+      assertThat(constructor.parameterTypes.toList()).isEmpty()
+
+      val staticMethods = factoryClass.declaredMethods.filter { it.isStatic }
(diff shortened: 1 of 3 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/dagger/AssistedInjectGeneratorTest.kt
```diff
@@ -642,12 +674,14 @@ public final class AssistedService_Factory {
 
   private fun compile(
     @Language("kotlin") vararg sources: String,
-    block: JvmCompilationResult.() -> Unit = { }
+    previousCompilationResult: JvmCompilationResult? = null,
+    block: JvmCompilationResult.() -> Unit = { },
```
JoelWilcox, line 678: - [ ] Undo - previously made this while trying out other test cases but this proved to not be needed.

RBusarow (approved):

vRallev (approved):

PR #752, opened 2023-09-28
Gradle `8.3` and AGP `7.3.1`

RBusarow:
The AGP update becomes necessary in order to resolve:

```
Could not determine the dependencies of task ':sample:app:testDebugUnitTest'.
> Could not create task ':sample:app:processDebugResources'.
   > Cannot use @TaskAction annotation on method IncrementalTask.taskAction$gradle_core() because interface org.gradle.api.tasks.incremental.IncrementalTaskInputs is not a valid parameter to an action method.
```

### DIFF
--- a/gradle/libs.versions.toml
+++ b/gradle/libs.versions.toml
@@ -8,7 +8,7 @@
 #
 
 [versions]
-agp = "7.1.0"
+agp = "7.3.1"
 autoService = "1.1.1"
 autoValue = "1.10.1"
 dagger = "2.46.1"
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-all.zip
 networkTimeout=10000
 zipStoreBase=GRADLE_USER_HOME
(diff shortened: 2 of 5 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.6-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-8.3-all.zip
```
vRallev, line 3: `bin` is good enough

vRallev (approved):

gradle/wrapper/gradle-wrapper.properties
RBusarow: Technically, but it's a pain in the butt when you're doing Gradle things and want to click into the source or even just read the javadoc.

Since Gradle's not being bundled into the release, it doesn't do any harm to just improve the devx a little bit.

vRallev (approved):

gradle/libs.versions.toml
```diff
@@ -8,7 +8,7 @@
 #
 
 [versions]
-agp = "7.1.0"
+agp = "7.3.1"
```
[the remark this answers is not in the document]
RBusarow: The lint issue was easy enough to fix, since it's just a sample app.

RBusarow:
I would like to replace the references to the deprecated `com.android.build.gradle.api.BaseVariant` and its subtypes in `AnvilPlugin`, but that stuff should really be backed by integration tests.  So, coming soon™.

PR #753, opened 2023-10-03
Update ComponentDetectorCheck for ksp

jhperezd:
This PR adds KSP support for ComponentDetectorCheck
- The changes to `ComponentDetectorCheck` follows the same pattern already used in `AnvilAnnotationDetectorCheck`. No additional logic/further changes have been introduced here.
- The corresponding test class was updated to handle for KSP and the current system by making the expected errors a bit more generic. Specifically, the error message returned during the test `a Dagger component causes an error` includes "Source0.kt:6" for KSP whereas it is "Source0.kt:6:11" for the current implementation. Similarly, the test `a Dagger component causes an error inner class` returns ":7" instead of ":7:13".


ZacSweers (approved):

JoelWilcox (approved):
Lgtm, thanks for the contribution!

JoelWilcox (approved):

Issue #754, opened 2023-10-06
Incremental compilation breaks when modifying constructor properties


Laimiux:
## What
Incremental builds consistently fail to compile after adding or removing constructor parameters to a class marked with `@ContributesBinding`. I was able to reproduce it with some slight modifications to the sample within the Anvil repo
- Checkout the branch or copy code over [link](https://github.com/square/anvil/compare/main...Laimiux:anvil:laimonas/incremental-build-failure?expand=1).
- First run `./gradlew :sample:assembleDebug`
- Add or remove a parameter within `RealParents` class
- Run `./gradlew :sample:assembleDebug` again
- You should see an error

```
/Users/laimonasturauskas/android/anvil/sample/app/build/tmp/kapt3/stubs/debug/com/squareup/anvil/sample/AppComponent.java:7: error: [Dagger/MissingBinding] com.squareup.anvil.sample.Parents cannot be provided without an @Provides-annotated method.
public abstract interface AppComponent extends com.squareup.anvil.sample.RandomComponent, com.squareup.anvil.sample.DescriptionComponent {
                ^
  
  Missing binding usage:
      com.squareup.anvil.sample.Parents is requested at
          com.squareup.anvil.sample.DescriptionComponent.parents()

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':sample:app:kaptDebugKotlin'.
> A failure occurred while executing org.jetbrains.kotlin.gradle.internal.KaptWithoutKotlincTask$KaptExecutionWorkAction
   > java.lang.reflect.InvocationTargetException (no error message)
```

The issue goes away when running `./gradlew :sample:asDebug --rerun-tasks`. One potential hint is that the `library/build/anvil/src-gen-main/` folder to be empty after incremental build compilation.

JoelWilcox:
Closing this as a duplicate since it looks like the root issue is the same as #693. Feel free to reopen/add more details if you think this is caused by a separate issue.

PR #755, opened 2023-10-06
[Bugfix] Incremental compilation via constructor parameter change.

Laimiux:
Attempting to fix https://github.com/square/anvil/issues/754, it would be good for someone with a better understanding of compiler plugins to take a look at this. I was able to test various scenarios successfully, but they might not be exhaustive.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
@@ -70,12 +70,17 @@ internal class CodeGenerationExtension(
 
     val anvilContext = commandLineOptions.toAnvilContext(anvilModule)
 
-    codeGenDir.listFiles()
-      ?.forEach {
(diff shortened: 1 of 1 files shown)


JoelWilcox:
> @Laimiux Thanks for taking care of the fix! I also ran it in the sample project and on my own project with different scenarios, and it indeed fixes the problematic behaviour.

➕ Thanks for taking a stab at this @Laimiux!

> @JoelWilcox could you please share the vision of the Anvil team on this fix? I see that it can solve the problem, but as @Laimiux said, it may not be applicable for all the use-cases in complex projects.

Due to the nature of the incremental compilation issues and how easy it is for regressions to happen in new areas even if we fix some cases, we're making sure we get appropriate integration test coverage in place before moving forward with related fixes. @RBusarow is currently working on that and applying the fix described in https://github.com/square/anvil/issues/693#issuecomment-1744013947.

JoelWilcox:
Closing now in favor of https://github.com/square/anvil/pull/836

PR #756, opened 2023-10-10
Update AnvilMergeAnnotationDetector for ksp

jhperezd:
This PR adds KSP support for `AnvilMergeAnnotationDetector`

- The changes to `AnvilMergeAnnotationDetector` follows the same pattern already used in AnvilAnnotationDetectorCheck. 
- Added `kspArgs["disable-component-merging"]` to `AnvilCompilation.kt` when mode=KSP, this is to resolve an issue where the annotations `@merge<component/subcompnent/module/Interface>` where being ignored.
- The corresponding test class was updated to handle KSP and the current system by making the expected errors a bit more generic. Specifically, the error message includes "Source0.kt:6" for KSP whereas it is "Source0.kt:6:7" for the current implementation.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilMergeAnnotationDetectorCheck.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilMergeAnnotationDetectorCheck.kt
@@ -70,7 +70,8 @@ internal object AnvilMergeAnnotationDetectorCheck : AnvilApplicabilityChecker {
       val clazz = projectFiles
         .classAndInnerClassReferences(module)
         .firstOrNull {
-          clazz -> ANNOTATIONS_TO_CHECK.any { clazz.isAnnotatedWith(it) }
+          clazz ->
+          ANNOTATIONS_TO_CHECK.any { clazz.isAnnotatedWith(it) }
         }
 
       if (clazz != null) {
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AnvilMergeAnnotationDetectorCheck.kt
```diff
@@ -14,32 +22,62 @@ import org.jetbrains.kotlin.descriptors.ModuleDescriptor
...
+  override fun isApplicable(context: AnvilContext) = context.disableComponentMerging
+
+  internal class KspGenerator(
+    override val env: SymbolProcessorEnvironment,
+  ) : AnvilSymbolProcessor() {
+    @AutoService(SymbolProcessorProvider::class)
+    class Provider : AnvilSymbolProcessorProvider(AnvilMergeAnnotationDetectorCheck, ::KspGenerator)
+
+    override fun processChecked(resolver: Resolver): List<KSAnnotated> {
+      val clazz = ANNOTATIONS_TO_CHECK
```
ZacSweers, line 44: Let's make this a sequence so it lazily resolves the types below

```suggestion
      val clazz = ANNOTATIONS_TO_CHECK
        .asSequence()
```

jhperezd: I've tested the change locally and pushed the change as suggested. 🙇

ZacSweers (approved):

JoelWilcox (approved):

JoelWilcox (approved):

PR #760, opened 2023-10-26
build logic + better publishing

RBusarow:
This introduces a convention plugin to handle publishing for the main build and the `gradle-plugin` build.  It mostly just uses the newer [base](https://vanniktech.github.io/gradle-maven-publish-plugin/base/) plugin, and most of the configuration still just comes from `gradle.properties`.  The important difference is that each module now sets its artifact ID, pom name, and pom description from inside the DSL, such as:

```groovy
publish {
  configurePom(
      artifactId: 'annotations-optional',
      pomName: 'Anvil Optional Annotations',
      pomDescription: 'Optional annotations that we\'ve found to be helpful with managing larger dependency graphs'
  )
}
```

This allows us to just symlink the `gradle-plugin/gradle.properties` without worrying about breaking publishing.

I tried to avoid yak-shaving.  There's lots more than can be moved to conventions, but this is enough to unblock local publishing for integration tests.


build-logic/build.gradle
```diff
@@ -0,0 +1,42 @@
+plugins {
+  alias(libs.plugins.kotlin.jvm)
+  alias(libs.plugins.ktlint)
+  id 'java-gradle-plugin'
+}
+
+gradlePlugin {
+  plugins {
+    module {
+      id = "library"
```
vRallev, line 10: "library" is so generic, why not `com.squareup.anvil.library`?

gradle/libs.versions.toml
```diff
@@ -87,6 +90,8 @@ kotlin-test = { module = "org.jetbrains.kotlin:kotlin-test-junit", version.ref =
 kotlinpoet = { module = "com.squareup:kotlinpoet", version.ref = "kotlinpoet" }
 kotlinpoet-ksp = { module = "com.squareup:kotlinpoet-ksp", version.ref = "kotlinpoet" }
 
+kgx = { module = "com.rickbusarow.kgx:kotlin-gradle-extensions", version.ref = "kgx" }
```
vRallev, line 93: Sneaky

vRallev (approved):

build-logic/build.gradle
RBusarow: In my head, I consider local convention plugins to be closely related to precompiled script plugins or built-in plugins.  Since the precompiled plugins often have short names and the built-in ones all have short names, I'm mostly convinced that this type of plugin should also have a short name.

It's not something I feel strongly about, though.  If you're unconvinced, I can change it.

vRallev: I don't feel strongly either. But my first impression was "what's that and where is it coming from". Searching only for "library" also takes longer.

gradle/libs.versions.toml
RBusarow: Hiding in plain sight!

Actually, I forgot about this or I would have called it out.  It has lots of excellent build-logic yak-shaving applications.

build-logic/build.gradle
RBusarow: I will continue to slightly lean in the other direction, but the IDs have now been lengthened.

PR #786, opened 2023-11-17
replace the Dokka Gradle plugin with Dokkatoo

RBusarow:
https://github.com/adamko-dev/dokkatoo

It's slowly being upstreamed into Dokka.  It still uses Dokka to generate the actual docs, but does so using much better Gradle logic.


RBusarow:
We're going to hold off a bit to see what Dokka 1.9.20 looks like.

PR #811, opened 2023-12-08
Convert all Gradle script files to the Kotlin DSL

RBusarow:


integration-tests/code-generator/build.gradle.kts
```diff
@@ -0,0 +1,25 @@
...
+
+  compileOnly(libs.auto.service.annotations)
+  kapt(libs.auto.service.processor)
+
+  testImplementation(testFixtures("com.squareup.anvil:compiler-utils"))
+  testImplementation(libs.junit)
+  testImplementation(libs.truth)
+}
+
+tasks.withType(KotlinCompile::class).configureEach {
```
JoelWilcox, line 21: Looks like you could use the new conventions extension here

integration-tests/mpp/android-module/build.gradle.kts
```diff
@@ -0,0 +1,88 @@
...
+  // but the new `androidTarget()` alternative doesn't exist in KGP 1.8.x.
+  @Suppress("DEPRECATION")
+  android()
+
+  sourceSets {
+    val androidMain by getting {
+      dependencies {
+
+        val kapt by configurations.getting
+        kapt.dependencies.addLater(libs.dagger2.compiler)
```
JoelWilcox, line 52: Why do we apply this here compared to before when it was in the `dependencies` block at the end of this file? And why `addLater` instead of directly applying?

JoelWilcox (approved):

integration-tests/mpp/android-module/build.gradle.kts
RBusarow: > Why do we apply this here compared to before when it was in the dependencies block at the end of this file?

I had been experimenting with a few different ways to make it work without the weird syntax, though they all failed.  I'll move it back.

> And why addLater instead of directly applying?

This is the standard behavior, but normally it's hidden by many layers of `dependencyNotation: Any` parameters.

All the dependencies from `libs` are some variation of `Provider<Dependency>`.  At this layer, the normal `dependencies.add(...)` only accepts a `Dependency` argument, and all those providers are added via `addLater(...)`.

integration-tests/code-generator/build.gradle.kts
RBusarow: Thanks!  I missed this one...

PR #813, opened 2023-12-12
Add the kotlinx binary compatibility plugin

RBusarow:
https://github.com/Kotlin/binary-compatibility-validator


compiler-utils/api/compiler-utils.api
JoelWilcox, line 1: Debating whether we will want to disable it for this module since we don't guarantee API compatibility on this and more-so provide it for convenience, but makes sense to start enabled and see whether we end up with noise problems.

JoelWilcox (approved):
Very cool! First time seeing this plugin

compiler-utils/api/compiler-utils.api
RBusarow: Those were pretty much my thoughts as well.  My rationale for leaving it enabled is that even though we don't guarantee anything, it's still nice to know?  Unless it gets annoying.

JoelWilcox: Yup I agree completely! And it's easy remove later if it does get annoying

PR #817, opened 2023-12-13
Gradle test framework base classes + some lifecycle-related tests

RBusarow:


gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
```diff
@@ -0,0 +1,90 @@
...
+          }
+
+          dir("src/main/java") {
+            injectClass()
+          }
+        }
+
+        val expected = when {
+          it.gradleVersion >= "8.5" -> "Calculating task graph as no cached configuration is available for tasks: compileKotlin"
+          it.gradleVersion >= "8.4" -> "Calculating task graph as no configuration cache is available for tasks: compileKotlin"
```
JoelWilcox, line 34: These read to me like the messages that would be displayed at the start when attempting to store the configuration cache rather than confirming that it was successful / compatible. Is there another message that we can/should be checking here similar to the `else` case?

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    shouldSucceed("compileKotlin")
+    shouldSucceed("compileKotlin") {
+      task(":compileKotlin")?.outcome shouldBe TaskOutcome.UP_TO_DATE
+    }
+  }
+
+  @TestFactory
+  fun `compileKotlin is FROM-CACHE after the project build directory is deleted`() = testFactory {
```
JoelWilcox, line 70: I love how clean and simple these test cases are

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/MoreAsserts.kt
```diff
@@ -0,0 +1,13 @@
+package com.squareup.anvil.plugin
+
+import io.kotest.matchers.file.shouldExist
+import io.kotest.matchers.shouldBe
+import java.io.File
+
+interface MoreAsserts {
+
+  infix fun File.shouldExistWithText(expectedText: String) {
```
JoelWilcox, line 9: Is this used in a followup PR? Okay to leave if-so, just making sure it's only temporarily dead code

RBusarow: Heh...  oops!

Yeah, it's used in some (currently failing) incremental tests.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/AnvilVersionMatrix.kt
```diff
@@ -0,0 +1,33 @@
...
+import com.rickbusarow.kase.gradle.KotlinDependencyVersion
+import com.rickbusarow.kase.gradle.KspDependencyVersion
+import com.rickbusarow.kase.gradle.VersionMatrix
+
+class AnvilVersionMatrix : VersionMatrix by VersionMatrix(agp + kotlin + gradle + dagger) {
+  companion object {
+    val agp = listOf("7.3.1", "7.4.2", "8.0.2", "8.1.1", "8.2.0").map(::AgpDependencyVersion)
+    val kotlin = listOf("1.8.21", "1.9.0", "1.9.10", "1.9.21").map(::KotlinDependencyVersion)
+    val gradle = listOf("8.5").map(::GradleDependencyVersion)
+    val dagger = listOf("2.46.1").map(::DaggerDependencyVersion)
```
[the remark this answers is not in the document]
RBusarow: You make good points.  I have plans for this but didn't document them.  I'll address that.

For posterity:

These should be independent of the current matrices in `ci.yml`.  Those are more limited since they change what the project builds with, whereas this matrix only defines what the consuming project uses.

Soon, I will need to move this matrix definition into `build-logic/conventions`, so that it can be broken into shards for CI.  The idea is that `ci.yml` is auto-updated via a task, and this specific `AnvilVersionMatrix.kt` in `:gradle-plugin` would be generated like a BuildConfig file.  This creates a single source of truth (in `build-logic/conventions`), keeps CI scalable, and makes it simple to run the full matrix on dev machines.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/BaseGradleTest.kt
```diff
@@ -0,0 +1,51 @@
...
+) : KaseGradleTest<GradleKotlinTestVersions>,
+  FileStubs,
+  MoreAsserts {
+
+  override val localM2Path: File
+    get() = localBuildM2Dir
+
+  override val kases: List<GradleKotlinTestVersions>
+    get() = versionMatrix.versions(GradleKotlinTestVersions)
+      .letIf(!fullTestRun) { it.takeLast(1) }
```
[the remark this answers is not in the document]
RBusarow: For clarity, `fullTestRun` is `true` by default, meaning that this behavior is disabled by default.

The last permutation will always have the last version from each list.  That's pretty intuitive once you've run some tests in the IDE.

image

If/when we get to a point where we start running tests against pre-release versions, then we can change this logic.  Personally, I don't foresee ever using this flag so I would have no objections to just removing it.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
RBusarow: Thanks!  That's the idea. :)

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
RBusarow: Nice catch!  You're right.  I was conflating two different things here.

The assertions have been updated.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/BaseGradleTest.kt
```diff
@@ -0,0 +1,51 @@
...
+abstract class BaseGradleTest(
+  override val versionMatrix: VersionMatrix = AnvilVersionMatrix(),
+) : KaseGradleTest<GradleKotlinTestVersions>,
+  FileStubs,
+  MoreAsserts {
+
+  override val localM2Path: File
+    get() = localBuildM2Dir
+
+  override val kases: List<GradleKotlinTestVersions>
```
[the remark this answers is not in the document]
RBusarow: I get your point, but *within* the context of the library, the meaning is clear.  The property holds a `List<Kase>`, and in this case (😁) those kases are versions, but even in Gradle-specific tests that won't always be so.

I think the best I could do would be to introduce an alias property here.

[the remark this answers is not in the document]
JoelWilcox: Hmm but the type is `GradleKotlinTestVersions`, that's why it feels like there's the disconnect between the name and the type. And I think an alias to have the type be `Kase` kinda goes in the opposite direction of what i was hoping because it introduces new terminology when there's already more "simple" terminology that can be used like the current type. IMO simpler, less context that is required to understand what is happening, the better. Not a hill I'm going to die on 😄, but I do feel like something like `versionCombos` better conveys what this is to any reader.

JoelWilcox (approved):

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/BaseGradleTest.kt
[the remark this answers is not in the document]
RBusarow: What I mean is that this is the interface that defines the property:

```kotlin
public interface HasKases<out K : Kase> {
  public val kases: List<K>
}
```

So it doesn't just relate to versions.

Meanwhile, by `alias` (not `typealias` ;)) I mean adding a property to `BaseGradleTest` like this:

```kotlin
abstract class BaseGradleTest(
  override val versionMatrix: VersionMatrix = AnvilVersionMatrix(),
) : KaseGradleTest<GradleKotlinTestVersions>,
  FileStubs,
  MoreAsserts {

  override val kases: List<GradleKotlinTestVersions>
    get() = // ...

  /** an alias for [kases] */
  open val versions: List<GradleKotlinTestVersions>
    get() = kases
}
```

PR #822, opened 2023-12-15
don't leak Anvil's annotation artifacts to the target project's compile classpath

RBusarow:


steve-the-edwards (approved):

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/ClasspathTest.kt
```diff
@@ -0,0 +1,146 @@
...
+   * Why can't we just use dependency-guard to check the classpath?
+   *
+   * Dependency-Guard can check the `runtimeClasspath` and `compileClasspath` configurations
+   * of the `gradle-plugin` project. That tells us what Anvil will contribute to a target project's
+   * `classpath` configuration (the one modified in the `buildScript` block).
+   *
+   * For this test, we're testing what dependencies Anvil adds to the target project's own
+   * `runtimeClasspath` and `compileClasspath` configurations.  Specifically, it adds the Anvil
+   * annotations dependency -- but is it added correctly as a `compileOnly` dependency,
+   * or is it being added as `implementation` or something else?
```
JoelWilcox, line 87: This comment is great 👍

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/ClasspathTest.kt
```diff
@@ -0,0 +1,146 @@
...
+   */
+  @TestFactory
+  fun `annotations aren't in the runtime classpath`() = listOf(
```
JoelWilcox, line 90: Nit: Since we also care about verifying that the annotations are included in the compile classpath, something like `annotations are only included in the compile classpath` might be better for encoding the expectations

JoelWilcox (approved):

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/ClasspathTest.kt
RBusarow: That's an excellent point.  Thanks!

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/ClasspathTest.kt
RBusarow: Thanks 😄  I think the coffee had just kicked in.

PR #826, opened 2023-12-19
Use a different build directory per Gradle build for main-build projects

RBusarow:
We been getting flakes in CI like this one:

```
> Cannot access output property 'outputDir' of task ':gradle-plugin:generateGradleTestBuildConfig'. Accessing unreadable inputs or outputs is not supported. Declare the task as untracked by using Task.doNotTrackState(). [...]
   > Failed to create MD5 hash for file '/home/runner/work/anvil/anvil/gradle-plugin/build/generated/sources/buildConfig/gradleTest/com/squareup/anvil/plugin/buildProperties/BuildProperties.kt' as it does not exist.
```

These are happening because the projects in the main "anvil" build are technically part of two different Gradle builds -- the main/root one, and the one that's included by the `:delegate` build.  Both of those builds are acting upon the same files, and Gradle doesn't share concurrency controls between its builds.  So, we have a shared mutable state (their build directories) and every build has a chance of creating a race condition like the one above.

The fix is to have different directories, which are `$projectDir/build/anvil-build` and `$projectDir/build/included-build`.

### DIFF
--- a/build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/BasePlugin.kt
+++ b/build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/BasePlugin.kt
@@ -2,7 +2,9 @@ package com.squareup.anvil.conventions
 
 import com.rickbusarow.kgx.extras
 import com.rickbusarow.kgx.javaExtension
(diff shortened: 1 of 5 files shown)


steve-the-edwards (approved):
LGTM!

Issue #751, opened 2023-09-27
Support KSP in factory generation


ZacSweers:
This issue is to track KSP support in factory gen, and different from #704

- [x] AnvilMergeAnnotationDetectorCheck.kt #756
- [x] AssistedFactoryGenerator.kt #795
- [x] AssistedInjectGenerator.kt #795
- [ ] BindsMethodValidator.kt
- [x] ComponentDetectorCheck.kt #753
- [x] InjectConstructorFactoryGenerator.kt #795
- [x] MapKeyCreatorGenerator.kt #788
- [x] MembersInjectorGenerator.kt #795
- [x] ProvidesMethodFactoryGenerator.kt #806
- [x] ContributesMultibinding #740
- [x] ContributesTo #742
- [x] AnvilAnnotationDetectorCheck #722
- [x] ContributesBinding #722
- [x] BindingModuleGenerator #877
- [x] ContributesSubcomponentGenerator #828

jhperezd:
👋 I'll be taking `ComponentDetectorCheck.kt` and `AnvilMergeAnnotationDetectorCheck.kt` to begin with.

r0adkll:
👋🏻 I'll be taking `MembersInjectorGenerator.kt` to start with as well

jhperezd:
👋 PRs for `ComponentDetectorCheck.kt` (https://github.com/square/anvil/pull/753) and `AnvilMergeAnnotationDetectorCheck.kt` (https://github.com/square/anvil/pull/756) are up. I'll take `InjectConstructorFactoryGenerator.kt` next.

ursusursus:
May I ask?

Is this about not having to use

```
anvil {
  generateDaggerFactories = true // default is false
}
```
such that one would then use `ksp dagger` everywhere instead, therefore avoiding the incremental compiliation issue?

ZacSweers:
That wouldn't be enough as you still need anvil's hint generation for contributed elements

IlyaGulya:
@ZacSweers 
Hello!
Would you accept help from the community to speed up this issue resolution?

ZacSweers:
Sure! The last two are gnarly, so it make take a bigger lift. I can write up more about them tomorrow

ZacSweers:
So the main issue is that both of those generates require doing class scanning, which complicates things in KSP due to it processing nodes that are not in the compilation source set.

This is obviously necessary for module/component merging, but ideally shouldn't be necessary for any of the factory generators.

[...]

PR #828, opened 2023-12-24
Support KSP in ContributesSubcomponentGenerator

ZacSweers:
Ref #751


compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentCodeGen.kt
ZacSweers, line 1: I renamed this file without code changes first before proceeding, but github still shows them as a separate file. Sorry :/, idk what else to do about this case

JoelWilcox: You're good, I can still see the diff by viewing the chunk of commits after the rename. Thanks for splitting it up!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ksp/KSAnnotationExtensions.kt
```diff
@@ -93,6 +94,29 @@ internal fun KSAnnotation.scopeOrNull(): KSType? {
...
+internal fun KSAnnotation.replaces(): List<KSClassDeclaration> =
+  (argumentAt("replaces")?.value as? List<KSType>).orEmpty().mapNotNull {
+    it.resolveKSClassDeclaration()
+  }
+
+@Suppress("UNCHECKED_CAST")
+internal fun KSAnnotation.exclude(): List<KSClassDeclaration> =
+  (argumentAt("exclude")?.value as? List<KSType>).orEmpty().mapNotNull {
+    it.resolveKSClassDeclaration()
+  }
```
JoelWilcox, line 107: Did you come across any cases here where it was expected to get a null back? I'm thinking we should throw an exception if one of the replaces/exclude types can't be resolved, otherwise if I'm reading correctly we'll silently skip one or more types.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentCodeGen.kt
```diff
@@ -36,207 +59,402 @@ import kotlin.reflect.KClass
...
+              message = "${clazz.qualifiedName?.asString()} is contributed to the Dagger graph, but the " +
+                "interface is not public. Only public interfaces are supported.",
+              node = clazz,
+            )
+          }
+
+          clazz.getKSAnnotationsByType(ContributesSubcomponent::class)
+            .forEach { annotation ->
+              for (it in annotation.replaces()) {
+                val scope = annotation.scope().resolveKSClassDeclaration() ?: continue
```
JoelWilcox, line 94: Similar thought here on throwing an exception if the type isn't found

compiler/src/main/java/com/squareup/anvil/compiler/codegen/ksp/KSAnnotationExtensions.kt
ZacSweers: It's possible if it refers to a type that's not yet been generated. In theory, the correct thing to do here would be to track this original type and return it to KSP to defer them to a later round. We don't currently support this in embedded code gen though, and there's not currently a way to easily defer these. This matches the current embedded behavior, and would be worth a wider cleanup in a later PR to add support for deferring unresolved types.

JoelWilcox: Agree on the first part around deferring eventually :+1:. Can you point me to the equivalent skipping logic in the embedded impl (or relevant test if you came across one)? When I reviewed before I didn't see that, and it looked like this case would bubble up as an exception somewhere along the line while creating the reference classes for the annotation arguments. It's possible I missed it since the embedded impl has a lot more levels of abstraction before getting to specifically looking for the `exclude` arguments, just want to verify that the behavior is the same.

ZacSweers: It will accept the ClassReference list as-is [here](https://github.com/square/anvil/blob/8aeaad1092ed92281f65d7d3e5db17ad6197075b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt#L73-L74), but default to empty if that value is null.

PSI will offer an error element for this case, but ClassReference doesn't ever account for these. Instead, the pattern I see is more or less that behavior is either undefined or it gracefully avoids those situations. In this case I _think_ it's the latter, but tbh I'm not super sure how the PSI lookups for classreferences work just glancing at the code. We can make them error if you want though 👍

ZacSweers: (Note that `exclude()` is not used in this PR, just opportunistically added it since it was missing. `replaces()` is what's used.)

JoelWilcox: > It will accept the ClassReference list as-is [here](https://github.com/square/anvil/blob/8aeaad1092ed92281f65d7d3e5db17ad6197075b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnnotationReference.kt#L73-L74), but default to empty if that value is null.
>
> PSI will offer an error element for this case, but ClassReference doesn't ever account for these. Instead, the pattern I see is more or less that behavior is either undefined or it gracefully avoids those situations. In this case I think it's the latter, but tbh I'm not super sure how the PSI lookups for classreferences work just glancing at the code. We can make them error if you want though 👍

Gotcha, yea the default to empty on the argument itself makes sense since it might not have been set. My main concern is around if one of the types in a provided argument list isn't resolvable resulting in a null value. 

I would say let's go ahead throw an exception in that case for the KSP impl for now (unless it reveals some tests covering this case where we don't throw an exception with the embedded impl). It's simple enough to change it back later if we find that there are cases where we want to ignore an unresolvable type. Basically just:
```
map {
  it.resolveKSClassDeclaration() ?: throw ...
}
  ```

ZacSweers:
@JoelWilcox sorry for the delay. I've merged latest main and added the throwing behavior requested: [1717ce5](https://github.com/square/anvil/pull/828/commits/1717ce52ff0927966bc6409565e5fc12c1959633)

JoelWilcox:
>@JoelWilcox sorry for the delay. I've merged latest main and added the throwing behavior requested: [1717ce5](https://github.com/square/anvil/pull/828/commits/1717ce52ff0927966bc6409565e5fc12c1959633)

No worries, thanks for updating it!

JoelWilcox (approved):

PR #830, opened 2023-12-25
Update to dagger 2.50

ZacSweers:
This updates Dagger to 2.50, which includes two code gen changes.

**1. Prohibit extension declarations for `@Binds` functions**

Starting in [Dagger 2.48](https://github.com/google/dagger/releases/tag/dagger-2.48), `@Binds` functions can no longer be extension functions.

Note that dagger also prohibits these on `@Provides` functions, but Anvil didn't have any tests or validation for this case. I added some opportunistically.

**2. Add support for new `dagger.internal.Provider` factories**

Dagger 2.50 introduced a new `Provider` type in preparation for its support of jakarta: https://github.com/google/dagger/releases/tag/dagger-2.50. The way this is implemented in AssistedFactory code gen is via a second `createFactoryProvider()` factory function that returns an instance of this new internal Provider type. This PR updates Dagger and adds support for this new type.

I had to also force a newer Guava version as it seems Dagger updated theirs and uses some newer APIs, though surprisingly this isn't exposed transitively on the classpath.


compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/AssistedFactoryCodeGen.kt
```diff
@@ -559,19 +559,24 @@ object AssistedFactoryCodeGen : AnvilApplicabilityChecker {
...
-                .addTypeVariables(typeParameters)
-                .addParameter(DELEGATE_FACTORY_NAME, generatedFactoryTypeName)
-                .returns(Provider::class.asClassName().parameterizedBy(baseFactoryTypeName))
-                .addStatement(
-                  "return %T.create(%T($DELEGATE_FACTORY_NAME))",
-                  InstanceFactory::class,
-                  implParameterizedTypeName,
-                )
-                .build(),
+              createFactory("createFactoryProvider", dagger.internal.Provider::class.asClassName()),
```
JoelWilcox, line 579: Does Dagger always generate both now? Or are they keying off anything to decide on one vs the other? Looks like all the tests passed so it should be fine, but it feels a little surprising

ZacSweers: It always generates both now. The original for backward compatibility, the new one for forward compatibility. They're going to support both going forward

JoelWilcox (approved):

ZacSweers:
errrr I don't know what changed on main that broke this so much? It was green before

ZacSweers:
Should be fixed up now 👍

ZacSweers:
... these pass locally for me. Ugh

ZacSweers:
nvm found a new issue. I think this is a slight behavior difference between dagger kapt and KSP though, gonna dig in. Weird that these passed before though, curious what changed on main to trigger these in a way that didn't fail previously 🤔

RBusarow:
I'm guessing it's related to the KotlinPoet bump.  The `KSTypeReference.toTypeName()` extension ([link](https://github.com/square/kotlinpoet/blob/76b3d25b386af88718f483dabf4158a433abc830/interop/ksp/src/main/kotlin/com/squareup/kotlinpoet/ksp/KsTypes.kt#L182-L195)) used to automatically resolve, but now it doesn't.  That would explain the missing variance here:

```
com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest > a factory class is generated for a generic class with two type parameters and a where clause[Use Dagger: false, mode: Ksp(symbolProcessorProviders=[])] FAILED
    keys with wrong values
    for key       : R
    expected value: [java.util.Set<? extends java.lang.String>]
    but got value : [java.util.Set<java.lang.String>]
    ---
    expected      : {T=[java.lang.Appendable, java.lang.CharSequence], R=[java.util.Set<? extends java.lang.String>]}
    but was       : {T=[java.lang.Appendable, java.lang.CharSequence], R=[java.util.Set<java.lang.String>]}
        at app//com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest$a factory class is generated for a generic class with two type parameters and a where clause$1.invoke(InjectConstructorFactoryGeneratorTest.kt:2354)
        at app//com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest$a factory class is generated for a generic class with two type parameters and a where clause$1.invoke(InjectConstructorFactoryGeneratorTest.kt:2331)
        at app//com.squareup.anvil.compiler.internal.testing.AnvilCompilationKt.compileAnvil(AnvilCompilation.kt:276)
        at app//com.squareup.anvil.compiler.internal.testing.AnvilCompilationKt.compileAnvil$default(AnvilCompilation.kt:230)
        at app//com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest.compile(InjectConstructorFactoryGeneratorTest.kt:2741)
        at app//com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest.compile$default(InjectConstructorFactoryGeneratorTest.kt:2737)
        at app//com.squareup.anvil.compiler.dagger.InjectConstructorFactoryGeneratorTest.a factory class is generated for a generic class with two type parameters and a where clause(InjectConstructorFactoryGeneratorTest.kt:2331)
```

ZacSweers:
Sure, but I haven't made any changes to `InjectConstructorFactoryCodeGen`, so how is this not also failing on main?

JoelWilcox (approved):

PR #832, opened 2023-12-29
Test Kotlin 2.0 Beta in CI

JoelWilcox:


compiler/src/main/java/com/squareup/anvil/compiler/IrUtils.kt
```diff
@@ -38,6 +39,7 @@ internal fun IrClass.requireClassId(): ClassId {
   )
 }
 
+@OptIn(UnsafeDuringIrConstructionAPI::class)
```
JoelWilcox, line 42: - [ ] Source-set split for these opt-ins / owner access

PR #833, opened 2023-12-31
[bugfix] Binding supertype which is narrower than return type is wrongly allowed

### LINKED ISSUE #750: Incorrect type validation for `@Binds`
Porting from https://github.com/google/dagger/issues/4065

---

> I was working on a lint and questioning if I understood `PsiType.isAssignableFrom` correctly.
> 
> The lint is to validate that a dagger `@Binds` function is correctly binding compatible types. For example, it will show a lint error in the IDE if someone tries to bind a `String` to an `Int`. Obviously Dagger checks this at compile-time, but a lint is nice to get instant feedback in the IDE.
> 
> There's a complex case I ran into though that has me wondering
> 
> Given these types
> 
> ```kotlin
> sealed interface ItemDetail {
>   object DetailTypeA : ItemDetail
> }
> 
> interface ItemMapper<T : ItemDetail>
> 
> class DetailTypeAItemMapper : ItemMapper<ItemDetail.DetailTypeA>
> ```
> 
> We have a bind function like so
> 
> ```kotlin
> @Module
> interface SomeModule {
>   @Binds fun validComplexBinding(real: DetailTypeAItemMapper): ItemMapper<ItemDetail>
> }
> ```
> 
> The lint check correctly reports the return type as not assignable from the bound instance. Dagger's compiler however accepts it as valid, and presumably in generated code does an unsafe cast that happens to work at runtime due to erased generics.
> 
> I wanted to raise this up as a possible issue, as I think Dagger maybe incorrectly allowing this and the lint check (via PSI) is correctly pointing out that `DetailTypeAItemMapper` isn't a subtype of `ItemMapper<ItemDetail>` because generics are invariant by default.

----

Notes since the issue on dagger: Dagger correctly catches this, it seems Anvil does not.

IlyaGulya:
This fixes #750


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -220,6 +226,14 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
 
     override fun enclosingClassesWithSelf(): List<Psi> = enclosingClassesWithSelf
 
+    override fun getDescriptor(): ClassDescriptor {
+      return module.findClassAcrossModuleDependencies(this.classId)
```
IlyaGulya, line 230: There's only one place that I'm a bit worried about. 
I'm not sure if there will be performance impact.
Should I cache class descriptor instead of resolving it every time?

JoelWilcox: Since the logic and fix in question is in the context of PSI, we would want to avoid utilizing and resolving descriptors here. Will leave a comment in the validator for what I think would be the preferred fix.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -74,50 +95,49 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
...
+    when {
+      parameterSuperTypes.any { it.isAssignableTo(returnType) } -> return
+      receiverSuperTypes.any { it.isAssignableTo(returnType) } -> return
     }
-  }
 
-  private fun MemberFunctionReference.Psi.parameterMatchesReturnType(): Boolean {
-    return parameterSuperTypes()
-      ?.contains(returnType().asClassReference())
-      ?: false
```
JoelWilcox, line 101: I would propose keeping the existing structure in the validator and modifying this method to utilize `TypeName` instead of `ClassReference` since it contains more of the information that we need in this case. You should be able to use
```
      ?.map { it.asTypeName() }
      ?.contains(returnType().asClassReference().asTypeName())
```
check for the full type info. Then `parameterSuperTypes()` would also be updated accordingly to provide `TypeReference`s.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -376,6 +377,36 @@ class BindsMethodValidatorTest(
...
+        @Module
+        interface SomeModule {
+          @Binds fun shouldBeInvalidComplexBinding(real: DetailTypeAItemMapper): ItemMapper<ItemDetail>
+        }
+      """.trimIndent(),
+    ) {
+      assertThat(exitCode).isEqualTo(ExitCode.COMPILATION_ERROR)
+      assertThat(messages).contains(
+        "@Binds methods' parameter type must be assignable to the return type",
+      )
```
JoelWilcox, line 406: Please also include an assert here verifying the more detailed portion of the error message that calls out the specific types that didn't match up. See `a binding with an incompatible parameter type fails to compile` for an example

IlyaGulya:
@JoelWilcox I've made all requested fixes. Please take a look.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -131,3 +132,41 @@ private val truePredicate: (Any?) -> Boolean = { true }
...
+        current = ""
+      }
+
+      else -> {
+        current += char
+      }
+    }
+  }
+
+  return builder.toString()
```
JoelWilcox, line 171: I think you could simplify this using a regex, something like `type.replace("([a-z]+\\.)+".toRegex(), "")`. Probably needs a little more tinkering to handle nested class cases like com.squareup.Foo.Bar

build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/KtlintConventionPlugin.kt
```diff
@@ -13,6 +13,18 @@ open class KtlintConventionPlugin : Plugin<Project> {
...
+        "root-build/generated",
+        "included-build/generated",
+      )
+      ktlint.filter {
+        it.exclude {
+          pathsToExclude.any { excludePath ->
+            it.file.path.contains(excludePath)
+          }
+        }
+      }
```
JoelWilcox, line 27: Is this necessary in the context of the fix? Unclear what the related error/issue is currently

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -54,70 +79,71 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
...
-      throw AnvilCompilationExceptionFunctionReference(
-        message = "@Binds methods' parameter type must be assignable to the return type. " +
-          "Expected binding of type $returnType but impl parameter of type " +
-          "${paramSuperTypes.first()} $superTypesMessage",
-        functionReference = function,
-      )
+    when (returnTypeName) {
+      in parameterSuperTypes -> return
+      in receiverSuperTypes -> return
+      else -> Unit
```
JoelWilcox, line 111: Nit: since we don't have an else case here it would feel a bit nicer to keep it as an `if (condition || condition) { // fail }` structure, which is also beneficial over the early returns since that means this check isn't required to stay at the bottom of the validations list as we add more or shift things around.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -54,8 +56,11 @@ class BindsMethodValidatorTest(
       )
       if (!useDagger) {
         assertThat(messages).contains(
-          "Expected binding of type Bar but impl parameter of type Foo only has the following " +
-            "supertypes: [Ipsum, Lorem]",
+          Errors.bindsParameterMustBeAssignable(
```
JoelWilcox, line 59: I'd prefer we keep the hardcoded expected value in this case, since it's possible a bug could be introduced in `bindsParameterMustBeAssignable()` later on which automatically changes the test to expect an incorrect value.

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
```diff
@@ -376,6 +381,45 @@ class BindsMethodValidatorTest(
...
+        }
+      """.trimIndent(),
+    ) {
+      assertThat(exitCode).isEqualTo(ExitCode.COMPILATION_ERROR)
+      assertThat(messages).contains(
+        "@Binds methods' parameter type must be assignable to the return type",
+      )
+      if (!useDagger) {
+        assertThat(messages).contains(
+          Errors.bindsParameterMustBeAssignable(
```
JoelWilcox, line 413: Same here

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -131,3 +132,41 @@ private val truePredicate: (Any?) -> Boolean = { true }
  * method throws an exception if more than one element is found.
  */
 internal fun <T> Iterable<T>.singleOrEmpty(): T? = singleOrEmpty(truePredicate)
+
+internal fun TypeName.toStringOfSimpleNamesOnly(): String {
+  return removePackageNames(toString())
+}
```
RBusarow, line 138: This extension (and the subsequent String parsing function) ignore all of the work that `TypeName` has already achieved.

If the type name has simple names, then it's a `ClassName` or a `ParameterizedTypeName`.  You could do something like:

```kotlin
internal fun TypeName.toStringOfSimpleNamesOnly(): String {
  return when (this) {
    is ParameterizedTypeName -> rawType.toStringOfSimpleNamesOnly()
    is ClassName -> simpleNames.joinToString(separator = ".")
    is LambdaTypeName,
    is TypeVariableName,
    is WildcardTypeName,
    Dynamic,
    -> error("???")
  }
}
```

From the context in which this is being called, the type name should be one of the first two.  If I was the one creating this extension, it would be inside the `BindsMethodValidator` class.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -54,70 +79,71 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
...
       )
     }
 
+    val parameters = function.parameters
+    val psiFunction = function.function
     val hasSingleBindingParameter =
-      (function.parameters.size == 1 && !function.function.isExtensionDeclaration()) ||
-        (function.parameters.isEmpty() && function.function.isExtensionDeclaration())
+      (parameters.size == 1 && !psiFunction.isExtensionDeclaration()) ||
+        (parameters.isEmpty() && psiFunction.isExtensionDeclaration())
```
RBusarow, line 91: This logic and the new logic below start off by doing the same thing -- trying to figure out what the bound parameter is.  You can simplify your logic to something like:

```kotlin
val bindingParameter = listOfNotNull(
    function.singleParameterTypeOrNull(),
    function.receiverTypeOrNull()
  )
   .singleOrNull()
   ?: throw AnvilCompilationExceptionFunctionReference(...)

val returnTypeName = function.returnTypeOrNull()?.asClassReference()?.asTypeName()
  ?: throw AnvilCompilationExceptionFunctionReference(...)

val paramSuperTypes = bindingParameter.superTypes()

if (returnTypeName !in paramSuperTypes) {
  throw AnvilCompilationExceptionFunctionReference(...)
}
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -54,70 +79,71 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
...
-  }
 
-  private fun MemberFunctionReference.Psi.parameterMatchesReturnType(): Boolean {
-    return parameterSuperTypes()
-      ?.contains(returnType().asClassReference())
-      ?: false
+    val paramSuperTypes =
+      parameterSuperTypes
+        .ifEmpty { receiverSuperTypes }
+        .map { it.toStringOfSimpleNamesOnly() }
```
RBusarow, line 117: I wouldn't bother trying to parse the simple names at all, since they're only used in the exception message.  Just print out the fully qualified name.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
```diff
@@ -54,70 +79,71 @@ internal class BindsMethodValidator : PrivateCodeGenerator() {
...
+  private fun TypeReference?.superTypes(): Sequence<TypeName> {
+    this ?: return emptySequence()
+    return sequence {
+      yield(this@superTypes)
+      yieldAll(
+        this@superTypes
+          .asClassReference()
+          .directSuperTypeReferences(),
+      )
+    }.map { it.asTypeName() }
```
RBusarow, line 147: I might be missing something, but it seems like we should continue to use `allSuperTypeClassReferences(includeSelf = true)`.  This version only returns the type and its immediate supertypes, whereas `allSuperTypeClassReferences()` is recursive.

IlyaGulya:
Wow, ok I'll try to get some time to fix this

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
IlyaGulya: Ok, I've just thought it would be nice to have, but sure, I will remove this part

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
IlyaGulya: Thanks, I see! But this part is no longer needed since I've removed simple names parsing.

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
IlyaGulya: Removed this completely

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
IlyaGulya: Yeah, that's much cleaner approach, thanks!

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
IlyaGulya: Unfortunately, `allSuperTypeClassReferences` lose generic type information (I assume that's because `ClassReference` can't contain (or should not contain?) information about which types are used as type arguments in specific binding declaration. That's why I'm using `TypeReference`s here instead of `ClassReference`s
I will try to make a similar to `allSuperTypeClassReferences` function which returns `TypeReference`s instead of `ClassReference`s

IlyaGulya: Nevermind, seems that I was wrong

build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/KtlintConventionPlugin.kt
IlyaGulya: It's not necessary in the context of fix, but without this I'm getting ktlint issues on generated code when I'm running ktlint locally:

```
Execution failed for task ':gradle-plugin:ktlintGradleTestSourceSetCheck'.
> A failure occurred while executing org.jlleitschuh.gradle.ktlint.worker.ConsoleReportWorkAction
   > KtLint found code style violations. Please see the following reports:
     - /Users/ilyagulya/Projects/Community/anvil/gradle-plugin/build/root-build/reports/ktlint/ktlintGradleTestSourceSetCheck/ktlintGradleTestSourceSetCheck.txt
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
IlyaGulya: Fixed

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
IlyaGulya: Fixed

compiler/src/test/java/com/squareup/anvil/compiler/dagger/BindsMethodValidatorTest.kt
IlyaGulya: Fixed

IlyaGulya:
@RBusarow I've made all requested fixes. Please take a look.
I've also introduced new extension `TypeReference.allSuperTypeReferences(includeSelf: Boolean = false): Sequence<TypeReference>` which works the same as `ClassReference.allSuperTypeClassReferences` but returns `TypeReference`s instead. 
This is required to keep correct generic type arguments in the `TypeName` and provide better error message. 
If I would use `allSuperTypeClassReferences` instead, then `TypeName` will not contain specific type argument which were used in the binding (i.e. `ItemMapper<ItemDetail.DetailTypeA>`), but instead will contain just generic type variable (i.e. `ItemMapper<T : ItemDetail>`)

compiler/src/main/java/com/squareup/anvil/compiler/codegen/dagger/BindsMethodValidator.kt
IlyaGulya: Also, I'm not quite sure that `allSuperTypeClassReferences` is recursive. Or, at least, I don't see it 🙂

RBusarow (approved):

PR #836, opened 2024-01-03
incremental compilation fixes

RBusarow:
We start off with new feature toggle: `trackSourceFiles`.

It's present in the Gradle DSL and as a Gradle property. 

```kotlin
// build.gradle.kts
anvil {
  trackSourceFiles = true // default is true
}
```

The majority of changes are behind that toggle, but not all.

### `GeneratedFileWithSources` replaces `GeneratedFile`

In order to support incremental compilation, we need to track what source files resulted in what generated files, then aggregate that data in `CodeGenerationExtension`. The existing `GeneratedFile` type is perfectly situated for this, but I wanted to make it explicit whether the code being generated is going to be incremental or not, so I extracted an interface from it and introduced a new type.

If `trackSourceFiles` is enabled but the consuming project is using a custom generator that returns the deprecated `GeneratedFile`, the build will fail.

It is possible to return a `GeneratedFileWithSources` with an empty list of source files. Anvil will then treat the generated file as if it was generated from all source files in the source set, and it will be invalidated by any change.

|                                                             | `trackSourceFiles == true`   | `trackSourceFiles == false`  |
|-------------------------------------------------------------|------------------------------|------------------------------|
| `GeneratedFile`                                             | :x: not supported            | :white_check_mark: supported |
| `GeneratedFileWithSources`                                  | :white_check_mark: supported | :white_check_mark: supported |
| `GeneratedFileWithSources` <br/> with an empty sources list | :white_check_mark: supported | :white_check_mark: supported | 

### `build/anvil/src-gen-<variant>`  has moved

We now write a cache file as well, and need to put it somewhere. The new structure is pretty comparable to what KSP and KGP do.

Prior to these changes, the generated output's folder structure resembled this:

```
<project-directory>
  ├ build
  │ └ anvil
  │   ├ src-gen-debug
  │   │ └ com/example/DebugClass_Factory.kt
  │   └ src-gen-main
  │     └ com/example/InjectClass_Factory.kt
  ├ src
  │ ├ debug
  │ │ └ kotlin
  │ │   └ com/example/DebugClass.kt
  │ └ main
  │   └ kotlin
  │     └ com/example/InjectClass.kt
  └ build.gradle.kts
```

Now, the `src-gen-main` directory is gone and we have this:

```
<project-directory>
  ├ build
  │ └ anvil
  │   ├ debug
  │   │ └ caches
  │   │ │ └ generated-files-cache.bin
  │   │ └ generated
  │   │   └ com/example/DebugClass_Factory.kt
  │   └ main
  │     └ caches
  │     │ └ generated-files-cache.bin
  │     └ generated
  │       └ com/example/InjectClass_Factory.kt
  ├ src
  │ ├ debug
  │ │ └ kotlin
  │ │   └ com/example/DebugClass.kt
  │ └ main
  │   └ kotlin
  │     └ com/example/InjectClass.kt
  └ build.gradle.kts
```

### Tracking Anvil output in Gradle

`AnvilPlugin` will add the generated source code and cache binary file as `TaskOutputs` of their corresponding `compileKotlin` task:

```kotlin
kotlinCompilation.compileTaskProvider.configure { task ->
  if (variant.variantFilter.trackSourceFiles) {
    task.outputs.dir(srcGenDir)
    task.outputs.dir(anvilCacheDir)
  }
}
```

This means that Gradle will handle cache/restore logic on its own when it's a fresh build, when it's from a (Gradle) build cache, or when the task is already up to date.

This is not enough to handle incremental builds, like when a single source file is changed or if a source file is deleted.

### File hashing

The files that the Kotlin compiler gives us in `CodeGenerationExtension.onAnalysisCompleted` aren't only the changed files. There is no built-in way for us to compare those inputs to the inputs of the last invocation of `onAnalysisCompleted`.

We now create an MD5 hash of every file that goes through `CodeGenerationExtension`, whether it's generated or human-written. These hashes are stored in `build/anvil/<variant name>/caches/generated-file-cache.bin`, which is written to disk after every invocation of `onAnalysisCompleted`.

### Tracking source <--> generated relationships

Thanks to [GeneratedFileWithSources](#generatedfilewithsources-replaces-generatedfile), we know exactly which source files resulted in which generated files. These relationships are also stored in `build/anvil/<variant name>/caches/generated-file-cache.bin`.

```mermaid
flowchart TD

  gen1.kt --> gen4.kt

  gen2.kt --> gen6.kt
  gen1.kt --> gen5.kt
  gen2.kt --> gen5.kt
  gen3.kt --> gen6.kt

  source1.kt --> gen1.kt
  source2.kt --> gen2.kt
  source2.kt --> gen3.kt
  source3.kt --> gen3.kt 
```

### Deleting out-of-date outputs

A file is considered to be *changed* if it is in the input files and its current MD5 hash is different from its cached hash, or if the cache doesn't contain a hash for that file.

A file is considered to be *deleted* if it's present in the cache but not on disk, AND it is not tracked as a generated file in the cache.

For every changed or deleted file, we delete all downstream generated files recursively. For example, in the above diagram, if `Source_2.kt` is changed, we delete `gen2.kt`, `gen3.kt`, `gen5.kt`, and `gen6.kt`.

```mermaid
flowchart TD
  
  classDef changed fill:#A00;
  
  source1[source1.kt]
  source2[source2.kt]:::changed
  source3[source3.kt]
  gen1[gen1.kt]
  gen2[gen2.kt]:::changed
  gen3[gen3.kt]:::changed
  gen4[gen4.kt]
  gen5[gen5.kt]:::changed
  gen6[gen6.kt]:::changed

  gen1 --> gen4

  gen2 --> gen6
  gen1 --> gen5
  gen2 --> gen5
  gen3 --> gen6

  source1 --> gen1
  source2 --> gen2
  source2 --> gen3
  source3 --> gen3 
```

### Restoring missing outputs

Because we include the generated files as `TaskOutputs` of the `compileKotlin` task, Gradle will *delete* those files any time it re-runs the task. This would break incremental builds, where we may only receive a subset of the source files, and therefore only re-generate a subset of the output.

Before we generate any new code, we restore any missing generated files from the `generated-file-cache.bin` cache.

## TODO

- [ ] finalize the default value of `trackSourceFiles`


compiler-api/src/main/java/com/squareup/anvil/compiler/api/GeneratedFile.kt
```diff
@@ -8,7 +8,30 @@ import java.io.File
...
+  replaceWith = ReplaceWith(
+    expression = "GeneratedFileWithSources(file, content, setOf(TODO(\"Add some source files using <somePsiElement>.containingFileAsJavaFile()\")))",
+    imports = [
+      "com.squareup.anvil.compiler.api.GeneratedFileWithSources",
+      "com.squareup.anvil.compiler.internal.containingFileAsJavaFile",
+    ],
+  ),
+  level = DeprecationLevel.WARNING,
 )
+public data class GeneratedFile(
```
JoelWilcox, line 22: What do you think about adding a separate constructor here and deprecating the current no-source-files constructor instead of introducing the new type? That should result in fewer changes for folks migrating their own generators now, as well as later when we would presumably rename `GeneratedFileWithSources` back to just `GeneratedFile` when this original one gets fully deleted (reason being that the inclusion of sources feels odd to have encoded in the name once this current form is gone). From a quick glance that seems like it should also remove the need for the new interface. E.g.
```
public class GeneratedFile {

  public val file: File
  public val content: String
  public val sourceFiles: Set<File>

  public constructor(file: File, content: String, sourceFiles: Set<File>) {
    this.file = file
    this.content = content
    // <sourceFiles assertions go here instead of init>
    this.sourceFiles = sourceFiles
  }

  @Deprecated("stop using this etc")
  public constructor(file: File, content: String) {
    this.file = file
    this.content = content
    this.sourceFiles = emptySet()
  }
}
```

RBusarow: I did consider the second constructor, but I wanted a clean break.  That's for two main reasons:

1. `GeneratedFile` is a data class, so binary compatibility is a bit tricky.  It's best to just get away from that.
2. I want to add a warning or exception if the incremental behavior is enabled, but a (custom) generator returns results that don't track anything.  That's technically different than a result that simply doesn't have any source files.

build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/BasePlugin.kt
```diff
@@ -135,6 +137,12 @@ abstract class BasePlugin : Plugin<Project> {
         target.layout.buildDirectory.set(target.file("build/included-build"))
       }
     }
+
+    // Set the kase working directory ('<build directory>/kase') as a System property,
+    // so that it's in the right place for projects with relocated directories.
+    target.tasks.withType(Test::class.java).configureEach { task ->
```
handstandsam, line 143: Can you add a link to https://github.com/rickbusarow/kase/blob/255db67f40d5ec83e31755bc9ce81b1a2b08cf11/kase/src/main/kotlin/com/rickbusarow/kase/files/HasWorkingDir.kt#L93-L96

compiler-api/src/main/java/com/squareup/anvil/compiler/api/AnvilContext.kt
```diff
@@ -43,6 +43,14 @@ public interface AnvilContext {
    */
   public val disableComponentMerging: Boolean
 
+  /**
+   * Track the source files for each generated file. This allows for two new behaviors:
+   * - Generated code is "invalidated" and deleted when the source file is changed or deleted.
+   * - Generated code is cached in a way that Gradle understands,
+   *   and will be restored from cache along with other build artifacts.
+   */
+  public val trackSourceFiles: Boolean
```
handstandsam, line 52: Enables incremental compilation support.  This is achieved by maintaining a link between source that exists before Anvil is run and the Anvil generated files.  This allows Anvil to react to changes to sources used as input to Anvil.

handstandsam (approved):
Reviewed on a call with @RBusarow

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
```diff
@@ -0,0 +1,209 @@
...
+    |${filesToMd5.toList().joinToString("\n") { "${it.first} -> ${it.second.hex()}" }}
+    |========================
+  """.trimMargin()
+
+  override fun equals(other: Any?): Boolean {
+    if (this === other) return true
+    if (other !is GeneratedFileCache) return false
+
+    if (generatedToSources != other.generatedToSources) return false
+    if (sourcesToGenerated != other.sourcesToGenerated) return false
```
JoelWilcox, line 171: Should we be checking `generatedToContent` here too?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtensionTest.kt
```diff
@@ -83,36 +81,155 @@ class CodeGenerationExtensionTest {
...
+      assertThat(codeGenerator.isApplicableCalls).isEqualTo(1)
+      assertThat(codeGenerator.getCallsForInput(componentInterface)).isEqualTo(1)
+
+      assertThat(exitCode).isEqualTo(OK)
+      assertThat(classLoader.loadClass("com.squareup.test.Abc")).isNotNull()
+    }
+  }
+
+  @Test
+  fun `a code generator that does not track sources can compile with trackSourceFiles disabled`() {
```
JoelWilcox, line 202: could you also add a test for the case where trackSourceFIles is enabled and there's at least one source file tracked?

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
```diff
@@ -89,12 +92,168 @@ class LifecycleTest : BaseGradleTest() {
       )
     }
 
-    shouldSucceed("compileKotlin")
+    shouldSucceed("compileJava", withHermeticTestKit = true)
```
JoelWilcox, line 95: Why do we check `compileJava` now?

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -93,6 +93,26 @@ public abstract class AnvilExtension @Inject constructor(
...
+  /**
+   * Enables incremental compilation support.
+   *
+   * This is achieved by maintaining a link between source files that exist before Anvil is run
+   * and the Anvil-generated files. This allows for two new behaviors:
+   * - Generated code is "invalidated" and deleted when the source file is changed or deleted.
+   * - Generated code is cached in a way that Gradle understands,
+   *   and will be restored from cache along with other build artifacts.
+   *
+   * By default this feature is enabled.
```
JoelWilcox, line 105: I'm thinking maybe we should default this to disabled first and have it be opt-in while we get feedback on it, verify no major issues pop up, and also do benchmarking for performance impact. Then we can switch it to enabled by default in a follow-up release

JoelWilcox (commented):
Partial review, still working my way through most of the tests

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -62,78 +87,93 @@ internal class CodeGenerationExtension(
...
-        .mapNotNull { (file, content) ->
-          val virtualFile = LightVirtualFile(
-            file.relativeTo(codeGenDir).path,
-            KotlinFileType.INSTANCE,
-            content,
-          )
+    fun Collection<FileWithContent>.toKtFile(): List<KtFile> {
+      return mapNotNull { (file, content) ->
+        val virtualFile = LightVirtualFile(
+          file.path,
```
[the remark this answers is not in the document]
RBusarow: Nope.  As a matter of fact, I did that very early on and it caused lots of headaches.

The Psi files that are created by Kotlinc have absolute paths, so they all should.  I do relativize the paths before they go into the cache, though.  At that point, they're relative to the project directory.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -0,0 +1,545 @@
...
+      task(":compileKotlin")?.outcome shouldBe TaskOutcome.SUCCESS
+    }
+
+    rootAnvilMainGenerated.injectClassFactory.shouldNotExist()
+    rootAnvilMainGenerated.resolve("com/squareup/test/OtherClass_Factory.kt").shouldExist()
+  }
+
+  @TestFactory
+  fun `a generated factory is deleted if the @Inject annotation is removed`() =
+    withTrackSourceFiles { (trackSourceFiles) ->
```
JoelWilcox, line 117: Why do we use this / care about the non-trackSourceFiles case for this test (but not the others)?

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -0,0 +1,545 @@
...
+        gradlePropertiesFile(
+          """
+            com.squareup.anvil.trackSourceFiles=true
+          """.trimIndent(),
+        )
+      }
+
+      shouldSucceed("compileJava")
+
+      rootAnvilMainGenerated.injectClassFactory.shouldExist()
```
JoelWilcox, line 195: Could you add a factory content assertion here too? I think the explicitness would be valuable since we're focused on the generated content in this case

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -0,0 +1,545 @@
...
+
+    shouldSucceed("compileJava") {
+      task(":compileKotlin")?.outcome shouldBe TaskOutcome.SUCCESS
+    }
+
+    rootAnvilMainGenerated.injectClassFactory.shouldExist()
+    rootAnvilMainGenerated.resolve("com/squareup/test/OtherClass_Factory.kt").shouldExist()
+    rootAnvilMainGenerated.resolve(
+      "anvil/hint/binding/com/squareup/test/OtherClass.kt",
+    ).shouldExist()
```
JoelWilcox, line 351: These would all still be true even without the second compilation right? It feels like we want to utilize a different assert here based on the generated content and possibly changes to the updated source to verify that the supertype change was picked up in the recompilation. My concern right now would be that `compileKotlin` could run successfully but silently not pick up the source file difference and these assertions would still pass

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -0,0 +1,545 @@
...
+      }
+
+      componentModuleFile.shouldExist()
+      componentModuleFile.readText() shouldContain """
+        @Module
+        @ContributesTo(Any::class)
+        public abstract class AppComponentAnvilModule {
+          @Binds
+          public abstract fun bindTypeB(otherClass: OtherClass): TypeB
+        }
```
JoelWilcox, line 476: Nice, yea this is the type of content assertion I think we should include in the other test

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
RBusarow: That's my thought as well, except that while testing against a snapshot or otherwise-special build, it's more convenient to just have it enabled.  So I was thinking leave it enabled by default up until release, then disable it by default, then enable it in a follow-up.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/GeneratedFileWithSources.kt
```diff
@@ -0,0 +1,70 @@
...
+        )
+      }
+    }
+    require(sourceFiles.none { it == file }) {
+      "GeneratedFileWithSources must not contain the generated file as a source file."
+    }
+    require(sourceFiles.none { it.isAbsolute && !it.exists() }) {
+      "If a source file is also a generated file, its path should be relative.\n" +
+        sourceFiles.joinToString("\n")
+    }
```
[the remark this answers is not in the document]
RBusarow: I re-wrote this to clarify things.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
RBusarow: Updated

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
```diff
@@ -0,0 +1,209 @@
...
+  }
+
+  fun add(generated: GeneratedFileWithSources) {
+
+    val generatedAbsolute = GeneratedAbsoluteFile(generated.file)
+    val generatedRelative = generatedAbsolute.relativeTo(projectDir)
+
+    generatedToContent[generatedRelative] = generated.content
+    val generatedAsSource = generatedRelative.asSourceFile()
+    filesToMd5.computeIfAbsent(generatedAsSource) { generatedAsSource.md5() }
```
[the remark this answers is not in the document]
RBusarow: This is an intentional optimization.

The version without `computeIfAbsent { }` is for ensuring that the hash reflects the latest state of the source code.

The version with `computeIfAbsent { }` is called against files that are almost certainly already cached as part of the current compilation, so we skip the work if we can.

This logic has now been consolidated inside the `addMd5()` function.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/FileCacheOperations.kt
```diff
@@ -0,0 +1,121 @@
...
+
+        // If it's only a source file, don't delete it.
+        is SourceRelativeFile -> cache.removeSource(file)
+      }
+    }
+
+    val unchangedGenerated =
+      unchangedInput.flatMapTo(mutableSetOf()) { cache.getGeneratedFilesRecursive(it) }
+
+    println(
```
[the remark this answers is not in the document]
RBusarow: I missed this while I was cleaning up.  It was just for debugging.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/FileCacheOperations.kt
```diff
@@ -0,0 +1,121 @@
...
+  /**
+   * @param inputSourceFiles The source files that have changed since the last compilation.
+   */
+  fun restoreFromCache(inputSourceFiles: List<SourceAbsoluteFile>) {
+
+    val (changedInput, unchangedInput) = inputSourceFiles.partition { cache.hasChanged(it) }
+
+    val changedSource = changedInput
+      .mapTo(mutableSetOf()) { it.relativeTo(projectDir) }
+      .let { if (it.isNotEmpty()) it + SourceRelativeFile.ANY else it }
```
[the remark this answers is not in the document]
RBusarow: Moved and I added a comment at its new location.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -149,14 +189,78 @@ internal class CodeGenerationExtension(
...
+      |
+      |   // build.gradle(.kts)
+      |   anvil {
+      |     trackSourceFiles = true
+      |   }
+      |
+      |or enable the property in `gradle.properties`:
+      |
+      |   # gradle.properties
+      |   com.squareup.anvil.trackSourceFiles=true
```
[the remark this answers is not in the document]
RBusarow: Nice catch!  Thanks for noticing. :)

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -149,14 +189,78 @@ internal class CodeGenerationExtension(
     // PrivateCodeGenerators don't impact other code generators. Therefore, they can be called a
     // single time at the end.
     privateCodeGenerators.generateCode(anvilModule.allFiles.toList())
+    maybeCacheFiles(filesAsAbsoluteFiles, generatedFiles)
```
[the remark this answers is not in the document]
RBusarow: Things have been moved around a lot and I missed this.  In its current state, it'll only be accessed/initialized when `trackSourceFiles == true`.

compiler/src/test/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtensionTest.kt
RBusarow: That's the current base case, so it's tested over 1000 times, but I see the logic of it. 😄 

Added

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
RBusarow: Out of an abundance of caution. 😄 

As its name suggests, the `cmmpilekotlin` task only goes as far as running the Kotlin compiler.  But this test is creating a condition where the files generated from a previous compilation would be missing.  It's one of those cases where the Kotlin task actually succeeds but the subsequent Java compilation task fails.  I'll add a clarifying comment.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
RBusarow: The other tests fail if Anvil isn't tracking the source files.  This one should work in both cases, so the extra test case is just making sure I don't introduce a regression.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
RBusarow: added

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
RBusarow: There isn't actually any difference in any of the generated output between those two compilations.

The output change only occurs when a merge happens, so I added an `AppComponent` with `@MergeComponent(Any::class)`.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
JoelWilcox: > That's my thought as well, except that while testing against a snapshot or otherwise-special build, it's more convenient to just have it enabled. So I was thinking leave it enabled by default up until release, then disable it by default, then enable it in a follow-up.

It would be easy to get released in the wrong state that way though, and just creates additional work along the way. Shouldn't enabling for the snapshot be really easy now by just setting it in your gradle.properties file?

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
```diff
@@ -89,12 +93,172 @@ class LifecycleTest : BaseGradleTest() {
...
+            kotlinFile(
+              "com/squareup/lib1/InjectClass.kt",
+              """
+              package com.squareup.lib1
+              
+              import javax.inject.Inject
+              
+              class InjectClass @Inject constructor()
+              """.trimIndent(),
+            )
```
JoelWilcox, line 176:
```suggestion
            injectClass("com.squareup.lib1")
```

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
JoelWilcox: Nice, updated assertions are perfect

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/LifecycleTest.kt
JoelWilcox: Didn't mean to submit the review yet 🤡

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
```diff
@@ -0,0 +1,238 @@
...
+  fun addGeneratedFile(generated: GeneratedFileWithSources) {
+
+    val generatedAbsolute = AbsoluteFile(generated.file)
+    val generatedRelative = generatedAbsolute.relativeTo(projectDir)
+
+    generatedToContent[generatedRelative] = generated.content
+
+    addMd5(generatedRelative, overwrite = true)
+
+    filesToMd5.computeIfAbsent(generatedRelative) { generatedRelative.md5() }
```
JoelWilcox, line 92: It looks like the second md5 line is a duplicate/no-op now

```suggestion
    addMd5(generatedRelative, overwrite = true)
```

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
```diff
@@ -0,0 +1,238 @@
...
+  /**
+   * Returns `true` if the [sourceFile] has changed since the last time it was added to the cache.
+   */
+  fun hasChanged(sourceFile: FileType): Boolean {
+    return when (sourceFile) {
+      is AbsoluteFile -> hasChanged(sourceFile.relativeTo(projectDir))
+      is RelativeFile -> {
+        if (sourceFile == RelativeFile.ANY) return true
+        val currentMd5 = sourceFile.md5()
+        val previousMd5 = filesToMd5.put(sourceFile, currentMd5)
```
JoelWilcox, line 156: This should be using `get()` right? Right now it has a side effect of modifying the tracked MD5

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
JoelWilcox: Adding on to the previous comment, I think changing to false now is also preferred so that we're testing a snapshot as close to release as possible which lets us verify the expected default behavior etc

JoelWilcox (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
RBusarow: This is intentional, but it's tricky.

This function must always calculate the file's current md5 in order to do the comparison.  As such, we might as well cache it.  The `filesToMd5.put(...)` returns the previously-stored value or null, so we still get to do the same comparison -- and now we get the cache update for free.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
RBusarow: I blame CoPilot?

Nice catch.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/incremental/GeneratedFileCache.kt
JoelWilcox: 👍 that makes sense for optimizing. I don't remember if there's already a test for this, but if not it's probably worth adding one to make it explicit that this behavior is intentional. I would have assumed it was a bug and changed it to `get` if I was reading this sometime in the future since the method name `hasChanged` suggests it's only checking things

PR #838, opened 2024-01-05
enable explicit API for all published modules/source sets

RBusarow:
This change is mostly just to make the IDE happy.  The `compiler-utils` and `gradle-plugin` projects have the explicit types and modifiers in most places, but the flag isn't enabled so the IDE complains.


build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/BasePlugin.kt
```diff
@@ -98,15 +98,23 @@ abstract class BasePlugin : Plugin<Project> {
...
+                .replaceFirstChar(Char::lowercase),
+            )
+
+          return sourceSetName == "test" || sourceSetName == "gradleTest"
         }
 
-        val fromInt = when (val targetInt = target.jvmTargetInt()) {
-          8 -> JVM_1_8
-          else -> JvmTarget.fromTarget("$targetInt")
+        if (extension.explicitApi.get() && !isTestOrGradleTest()) {
```
JoelWilcox, line 113: Is the thought here just that applying it for test source sets doesn't really matter and adds unnecessary noise?

JoelWilcox (approved):

JoelWilcox:
- [ ] Need to block merging on this for a sec while pushing out the release

build-logic/conventions/src/main/kotlin/com/squareup/anvil/conventions/BasePlugin.kt
RBusarow: Exactly!

PR #851, opened 2024-01-12
binding extension properties to Gradle properties

RBusarow:
This change makes it so that all of Anvil's existing configuration options can be set via a `gradle.properties` or a `-P_______` command line option.

For any setting, Anvil take the first non-null value out of:
1. Settings in the Gradle DSL:
   ```kotlin
   anvil {
     generateDaggerFactories.set(true)
   }
   ```
3. Properties set via the command line like:
   ```sh
   ./gradlew build -Pcom.squareup.anvil.generateDaggerFactories=true
   ```
4. Properties specified in a project-level, root-project-level, or `~/.gradle/`-level `gradle.properties`:
   ```properties
   com.squareup.anvil.generateDaggerFactories = true
   ```
5. Hard-coded default values (never null)

Practically speaking, this makes it possible to *locally* feature-toggle things (like the upcoming incremental fix) without touching a project's build logic.

### DIFF
--- a/gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/AnvilExtensionTest.kt
+++ b/gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/AnvilExtensionTest.kt
@@ -1,23 +1,71 @@
 package com.squareup.anvil.plugin
 
+import com.rickbusarow.kase.Kase4
+import com.rickbusarow.kase.gradle.GradleDependencyVersion
+import com.rickbusarow.kase.gradle.GradleKotlinTestVersions
+import com.rickbusarow.kase.gradle.GradleProjectBuilder
+import com.rickbusarow.kase.gradle.HasGradleDependencyVersion
+import com.rickbusarow.kase.gradle.HasKotlinDependencyVersion
+import com.rickbusarow.kase.gradle.KotlinDependencyVersion
 import com.rickbusarow.kase.kase
-import io.kotest.assertions.asClue
+import com.rickbusarow.kase.times
 import io.kotest.matchers.string.shouldContain
+import org.junit.jupiter.api.DynamicNode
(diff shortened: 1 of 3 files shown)


gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/AnvilExtensionTest.kt
```diff
@@ -0,0 +1,114 @@
...
+            """.trimIndent(),
+          )
+        }
+
+        "the extension property is $default when not set anywhere".asClue {
+
+          shouldSucceed("printProperty") {
+            output shouldContain "$propertyName: $default"
+          }
+        }
```
JoelWilcox, line 43: Could you add some helper methods for this test factory? It would make it easier to quickly scan the test for a high-level idea of what is being tested since there's quite a bit going on. E.g. this highlighted chunk could be something like `generateProjectAndVerifyDefaultProperty` or just one for generating the project and leaving the assertion in place.

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/AnvilExtensionTest.kt
```diff
@@ -0,0 +1,114 @@
...
+          } else {
+            gradlePropertiesFile(
+              """
+              com.squareup.anvil.$propertyName=${!default}
+              """.trimIndent(),
+            )
+          }
+        }
+
+        "the extension property is ${!default} when only set in gradle.properties".asClue {
```
JoelWilcox, line 65: Similarly, could you break this up into multiple test factories? It looks like this is actually testing multiple use cases that would benefit from being their own tests. (not set anywhere case, only set in gradle.properties case, etc). Helper methods will likely be less important then too aside from the initial root project setup.

JoelWilcox (approved):

PR #861, opened 2024-01-31
support building against Kotlin 1.9.10 via system property arguments

RBusarow:


JoelWilcox:
Can we work on getting register bumped instead of re-adding 1.9.10 support?

RBusarow:
> Can we work on getting register bumped instead of re-adding 1.9.10 support?

That's a much bigger lift, especially if we go to 1.9.22 thanks to the [Compose changes](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.8).

JoelWilcox:
> That's a much bigger lift, especially if we go to 1.9.22 thanks to the [Compose changes](https://developer.android.com/jetpack/androidx/releases/compose-compiler#1.5.8).

Going to check in with MDX, we can merge this later if it becomes the blocking factor after the incremental fixes are in but the required changes for bumping in register looked pretty minimal previously

.github/workflows/ci.yml
```diff
@@ -169,8 +169,14 @@ jobs :
...
-      - name : Publish Snapshots 1.9
-        run : ./gradlew clean publish --no-build-cache --no-daemon --stacktrace --no-parallel
+      - name : Publish Release 1.9.22
+        run : ./gradlew publish --no-build-cache --no-daemon --stacktrace
+        env :
+          ORG_GRADLE_PROJECT_mavenCentralUsername : ${{ secrets.SONATYPE_NEXUS_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword : ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+
+      - name : Publish Release 1.9.10
+        run : ./gradlew publish --no-build-cache --no-daemon --stacktrace -Doverride_kotlin=1.9.10 -PVERSION_NAME=2.4.10-1-9-10-SNAPSHOT
```
gabrielittner, line 179: If you end up going with this, could you swap the release order so that the version without suffix for the newer Kotlin version is published second? The publishing order influences https://repo.maven.apache.org/maven2/com/squareup/anvil/gradle-plugin/maven-metadata.xml and with the current order renovate bot suggest updating `2.4.9` to `2.4.9-1-8`

RBusarow: Absolutely!  I know that pain well and that's an easy ~fix~ bandaid.

Long term, I would love to use [feature variants](https://docs.gradle.org/7.0/userguide/feature_variants.html#feature_variants) to ship multiple Kotlin version variants to the same artifact/pluginID, without the need for an overloaded version name.  I'm not certain it's viable, but I would definitely like to try.

PR #866, opened 2024-02-05
move snapshot publishing into its own workflow

RBusarow:
This allows us to manually trigger a publish-snapshot build against a different branch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -149,31 +149,7 @@ jobs:
           check-latest: true
 
       - name: Publish to Maven Local
-        run: ./gradlew clean publishToMavenLocal --no-build-cache --no-daemon --stacktrace --no-parallel
-
-  publish-snapshot:
-    runs-on: ubuntu-latest
-    if: github.repository == 'square/anvil' && github.ref == 'refs/heads/main'
-    timeout-minutes: 25
-    needs:
-      - test-ubuntu
-      - test-gradle-plugin
-      - gradle-wrapper-validation
-      - publish-maven-local
-
-    steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
-      - uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4
-        with:
-          distribution: 'zulu'
-          java-version: '17'
-          check-latest: true
-
-      - name: Publish Snapshots 1.9
-        run: ./gradlew clean publish --no-build-cache --no-daemon --stacktrace --no-parallel
-        env:
-          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
-          ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+        run: ./gradlew clean publishToMavenLocal --no-build-cache --no-daemon --stacktrace
 
   test-gradle-plugin:
     runs-on: ubuntu-latest
@@ -336,7 +312,6 @@ jobs:
       - ktlint
       - lint
       - publish-maven-local
-      - publish-snapshot
       - test-gradle-plugin
       - kapt-for-dagger-factories
       - instrumentation-tests
--- a/.github/workflows/publish-snapshot.yml
+++ b/.github/workflows/publish-snapshot.yml
@@ -1,0 +1,28 @@
+name: Publish snapshot
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+
+jobs:
+  publish-snapshot:
+    runs-on: ubuntu-latest
+    if: github.repository == 'square/anvil'
+    timeout-minutes: 25
+
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+      - uses: gradle/wrapper-validation-action@85cde3f5a1033b2adc2442631c24b530f1183a1a # v2
+      - uses: actions/setup-java@387ac29b308b003ca37ba93a6cab5eb57c8f5f93 # v4
+        with:
+          distribution: 'zulu'
+          java-version: '17'
+          check-latest: true
+
+      - name: Publish Release 1.9
+        run: ./gradlew clean publish --no-build-cache --no-daemon --stacktrace
+        env:
+          ORG_GRADLE_PROJECT_mavenCentralUsername: ${{ secrets.SONATYPE_NEXUS_USERNAME }}
+          ORG_GRADLE_PROJECT_mavenCentralPassword: ${{ secrets.SONATYPE_NEXUS_PASSWORD }}


.github/workflows/ci.yml
```diff
@@ -336,7 +312,6 @@ jobs :
       - ktlint
       - lint
       - publish-maven-local
-      - publish-snapshot
```
JoelWilcox, line 339: Is there another way to include the snapshot in the all-checks job?

I'm wondering if this change is worth it since we can already publish manual snapshots when needed prior to the change hitting main

.github/workflows/ci.yml
```diff
@@ -148,32 +148,8 @@ jobs :
...
-          java-version : '11'
-          check-latest : true
-
-      - name : Publish Snapshots 1.9
-        run : ./gradlew clean publish --no-build-cache --no-daemon --stacktrace --no-parallel
-        env :
-          ORG_GRADLE_PROJECT_mavenCentralUsername : ${{ secrets.SONATYPE_NEXUS_USERNAME }}
-          ORG_GRADLE_PROJECT_mavenCentralPassword : ${{ secrets.SONATYPE_NEXUS_PASSWORD }}
+      - name: Publish to Maven Local
+        run: ./gradlew clean publishToMavenLocal --no-build-cache --no-daemon --stacktrace
```
JoelWilcox, line 152: Why is `--no-parallel` removed?

.github/workflows/ci.yml
RBusarow: We can't do that anymore, with the recent changes to Square's OSS publishing.

.github/workflows/ci.yml
RBusarow: It hasn't been necessary since [0.25.0](https://github.com/vanniktech/gradle-maven-publish-plugin/blob/main/CHANGELOG.md#0250-2023-03-23) from March 2023.

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,28 @@
...
+jobs:
+  publish-snapshot:
+    runs-on: ubuntu-latest
+    if: github.repository == 'square/anvil'
+    timeout-minutes: 25
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v2
+      - uses: actions/setup-java@v2
```
JoelWilcox, line 18: We had `- publish-maven-local` here before, was that just an optimization to let us reuse the results of the local publishing step or are we losing out on anything but not being able to depend on it now?

.github/workflows/publish-snapshot.yml
```diff
@@ -0,0 +1,28 @@
...
+  push:
+    branches:
+      - main
+
+jobs:
+  publish-snapshot:
+    runs-on: ubuntu-latest
+    if: github.repository == 'square/anvil'
+    timeout-minutes: 25
+
```
JoelWilcox, line 14: Now that we don't have a dependency on
```
 - test-ubuntu
  - test-gradle-plugin
```
      
That means we'll be publishing a snapshot even if one of those fails right?

.github/workflows/ci.yml
```diff
@@ -148,32 +148,8 @@ jobs :
           java-version : '11'
           check-latest : true
 
-      - name : Publish to Maven Local
-        run : ./gradlew clean publishToMavenLocal --no-build-cache --no-daemon --stacktrace --no-parallel
-
-  publish-snapshot :
-    runs-on : ubuntu-latest
-    if : github.repository == 'square/anvil' && github.ref == 'refs/heads/main'
```
JoelWilcox, line 156: Looking around to see what other options we have since workflows can't reference each other and it seems like we lose some of the existing checks by moving this out of the main ci workflow; what do you think about us just modifying this branch condition as needed when making temporary snapshots? If I'm understanding correctly, it looks like we'd need to do something like that with the new workflow file anyway since it's still gating on `main`

.github/workflows/publish-snapshot.yml
RBusarow: Even before the change, it wouldn't have given any real benefit.  `ci.yml` doesn't do any caching/restoring of the build artifacts, so the jars and whatnot created by the local publishing task are all just discarded.  That `publish-maven-local` shard is essentially just a way to ensure that the publishing configuration works.

.github/workflows/publish-snapshot.yml
RBusarow: Sort of, but that's what required checks and "Require branches to be up to date before merging" are for.  Our `main` branch can never be red unless an admin intentionally bypasses those restrictions.

With that said, the `ci.yml` workflow is still triggered by pushes to `main`, so if something is broken we'll still get notified of it.

.github/workflows/ci.yml
RBusarow: The new workflow does not need to be edited in order to publish other branches.  Here's the `on` block:

```yaml

 on:
   workflow_dispatch:
   push:
     branches:
       - main
```

With that logic, it will run automatically any time there's a push to `main`.  But the `workflow_dispatch` node means that you can trigger it manually from the `Actions` tab, then select the branch you want to target.

It winds up looking like this:

image

.github/workflows/publish-snapshot.yml
JoelWilcox: That's fair 👍

.github/workflows/ci.yml
JoelWilcox: Oh nice, thanks for explaining! One last question/concern is differentiating the snapshots. My understanding right now is that running the workflow on a PR would clobber the existing snapshot generated from main unless someone first manually updated the version in gradle.properties which could cause issues if someone was already relying on the existing snapshot. Is there a way we can guarantee the PR-initiated workflow would use a custom version name?

gabrielittner: You could do something like this https://github.com/freeletics/khonshu/blob/main/.github/workflows/publish-snapshot.yml#L36-L39 that would would publish the snapshot as `<branch-name>-SNAPSHOT` if the workflow doesn't run on main. The env var takes precedence over `VERSION_NAME` in `gradle.properties`.

JoelWilcox: :+1: that setup looks awesome, thanks for sharing!

JoelWilcox: Discussed more with Rick offline, for now folks using [pinned versions](https://medium.com/@vanniktech/pinning-a-snapshot-dependency-to-a-specific-version-b477f79469fc) is sufficient for this concern. Maybe we'll look at updating the workflow logic later as well but for now this should be good

JoelWilcox (approved):

.github/workflows/publish-snapshot.yml
RBusarow: It's never a good idea to rely upon a dynamically-versioned dependency like a snapshot.  I would hope that if anyone is using our snapshots like that, they're also using [specifically pinned versions](https://medium.com/@vanniktech/pinning-a-snapshot-dependency-to-a-specific-version-b477f79469fc) to ensure consistency.

PR #877, opened 2024-02-16
Restructure binding module code gen

ZacSweers:
This PR reimplements binding module generation to perform all merging at the IR level and eliminate intermediate _merged_ binding modules.

At a high level, this does the following
- Updates `@ContributesBinding` and `@ContributesMultiBinding` to instead generate binding modules per-binding.
- Each binding modules is annotated with a new `@InternalBindingMarker` annotation that contains a minimal amount of metadata
  - `originClass` - the origin class that contributed the binding
  - `isMultibinding` - indicates if the binding is a multibinding
  - `priority` - corresponds to `ContributesBinding.Priority`.
  - `qualifierKey` - the computed "qualifier key"
- Each binding module is annotated with `@ContributesTo` and propagated via that infrastructure.
- During `ModuleMergerIr`, these binding modules are merged and metadata read from `@InternalBindingMarker` to properly allow for `excludes`, `replaces`, priority, etc all work correctly.
- Update tests for this new infrastructure, preserving as much of the existing binding module tests (unit and integration) while removing only now-irrelevant ones (i.e. ones that explicitly expect a BindingModule of the old style to always be generated).

This also functionally completes #751!

---

- [x] Update ContributesBinding and ContributesMultibinding to generate contributed binding modules
- [x] Update ContributesBinding and ContributesMultibinding tests for new impl
- [x] Aggregate contributed bindings in `ModuleMergerIr` + new modeling using `@InternalBindingMarker`
- [x] Finish reimplementing prioritization and replacements in `ModuleMergerIr`. This is a little tricky because `BindingModuleMerger` speaks classes but `ModuleMergerIr` speaks modules.
- [x] Update tests


gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -447,34 +464,73 @@ class IncrementalTest : BaseGradleTest() {
...
+          priority = "NORMAL",
+        )
+        public interface OtherClassAsComSquareupTestLibTypeBToKotlinAnyBindingModule {
           @Binds
-          public abstract fun bindTypeB(otherClass: OtherClass): TypeB
+          public fun bindTypeB(real: OtherClass): TypeB
         }
       """.trimIndent()
+
+      // TODO this doesn't work, which... suggests to me that IC might be broken here?
```
RBusarow, line 530: It seems like this is actually a Kotlin issue.  The `.kt` file with the invalid `TypeA` binding is not present at the start of the second compile task, but KGP/Kotlinc isn't deleting is the corresponding `.class` file.  I suspect this is because of a collision with the `[...]/OtherClassBindingModule.kt` name.

I ran this test with KGP's [build reports](https://kotlinlang.org/docs/gradle-compilation-and-caches.html#build-reports) enabled.  For the second compile task, where `TypeA` isn't bound at all, it should be deleting all the `TypeA` classes.  Instead, the report includes this:

([full log](https://github.com/square/anvil/files/14532629/second_compilation-collision.txt))
```
Process SignatureChanged(fqName = com.squareup.test.lib.OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule)
Process ParentsChanged(fqName = com.squareup.test.lib.OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule)
Deleting /lib/build/classes/kotlin/main/META-INF/lib.kotlin_module on clearing cache for /lib/build/anvil/main/generated/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
Deleting /lib/build/classes/kotlin/main/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModuleKt.class on clearing cache for /lib/build/anvil/main/generated/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
[...]
Deleting /lib/build/classes/kotlin/main/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeBToKotlinAnyBindingModule.class on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassBindingModule.kt
Deleting /lib/build/classes/kotlin/main/META-INF/lib.kotlin_module on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassBindingModule.kt
```


It's deleting the `TypeA` hint's `.class` file because of the `TypeA` hint's `.kt` file, but it's deleting the `TypeB` module `.class` file because `OtherClassBindingModule.kt` changed, and it's ignoring the `TypeA` module's `.class` file.  It then creates the `TypeB` module `.class` file again, which is how we wind up with both.

This wasn't a problem before your changes because that file's contents didn't actually change.  Coincidentally, all the changing `.class` files were linked to uniquely-named `.kt` files.

So, I forked your branch and updated the generators so that they create one file per module, and the file name matches the module: [diff](https://github.com/ZacSweers/anvil/compare/z/ksp/newBindsModule...square:anvil:rick/z/ksp/newBindsModule-individual-files).

([full log](https://github.com/square/anvil/files/14532636/second_compilation-fixed.txt))
```
Deleting /lib/build/classes/kotlin/main/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.class on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
Deleting /lib/build/classes/kotlin/main/META-INF/lib.kotlin_module on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
[...]
Process SignatureChanged(fqName = com.squareup.test.lib.OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule)
Process ParentsChanged(fqName = com.squareup.test.lib.OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule)
Deleting /lib/build/classes/kotlin/main/META-INF/lib.kotlin_module on clearing cache for /lib/build/anvil/main/generated/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
Deleting /lib/build/classes/kotlin/main/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModuleKt.class on clearing cache for /lib/build/anvil/main/generated/anvil/hint/merge/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule.kt
[...]
Deleting /lib/build/classes/kotlin/main/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeBToKotlinAnyBindingModule.class on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeBToKotlinAnyBindingModule.kt
Deleting /lib/build/classes/kotlin/main/META-INF/lib.kotlin_module on clearing cache for /lib/build/anvil/main/generated/com/squareup/test/lib/OtherClassAsComSquareupTestLibTypeBToKotlinAnyBindingModule.kt
```

So, it still deletes the brand new `TypeB` `.class` files and recreates them, but the important part is that it deletes that lingering `TypeA` `.class` file and it doesn't come back.

But your `mergedModules()` function is still finding an `OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule` reference somewhere, now with an exception when it tries to access the `Component.modules` field:

```
java.lang.ClassNotFoundException: com.squareup.test.lib.OtherClassAsComSquareupTestLibTypeAToKotlinAnyBindingModule
[...]
```

I really don't know why.  The generated stubs don't reference it, I don't see it in any decompiled .class files, and I can't find it anywhere with kotlinx-metadata.

The code compiles without issue.  I even turned the `:app` module into an application, with a `main` function that initializes the component and injects that `TypeB` dependency, and that code executes without issue.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Contribution.kt
```diff
@@ -0,0 +1,176 @@
...
+
+  // Can't be private until we move off jvmTarget 8
+  fun generateBindingModule(): TypeSpec {
+    val contribution = this
+    // Combination name of origin, scope, and boundType
+    val suffix = "As" +
+      contribution.boundType.generateClassNameString().capitalize() +
+      "To" +
+      contribution.scope.generateClassNameString().capitalize() +
+      bindingModuleNameSuffix
```
RBusarow, line 46: Could we break this up with some underscores?

ZacSweers: Yeah we could I was less worried about readability given this is never accessed directly by users and trying to minimize chars to avoid too-long class names

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
ZacSweers: Very strange. Let me see if I can get a couple eyes from the compiler/KGP side to see if this is a bug we should file for the former. In the meantime, do you think we should switch to the 1:1 approach from your fork? I can try digging in on the phantom binding module ref

RBusarow (approved):

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
ZacSweers: Looks like things are green now with the switch to 1:1 generation 🥳

RBusarow: LGTM!

ZacSweers:
@RBusarow wanna take a look at the updated 1:1 impl before I merge?

RBusarow:
> @RBusarow wanna take a look at the updated 1:1 impl before I merge?

Sorry, I should have been more clear with my comment [here](https://github.com/square/anvil/pull/877#discussion_r1523716421).  I have taken a look and I am happy with it.

PR #886, opened 2024-02-26
Wire up gradle to backend options + add future component merging backend option for future use

ZacSweers:
This allows enabling the new KSP code gen support


ZacSweers:
We could rename `backend` to `factoryGenBackend` to make it more explicit if we want, since it wasn't exposed before. No strong feelings either way

RBusarow:
> We could rename `backend` to `factoryGenBackend` to make it more explicit if we want, since it wasn't exposed before. No strong feelings either way

I like `factoryGenBackend`.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/ComponentMergingBackend.kt
```diff
@@ -0,0 +1,10 @@
+package com.squareup.anvil.compiler.api
+
+/** Possible modes of component merging. */
+public enum class ComponentMergingBackend {
+  /** Component merging runs as an IR plugin during kapt stub generation. */
+  IR,
```
RBusarow, line 6: How about something like...

```kotlin
public enum class AnvilBackend {
  EMBEDDED_BACKEND,
  KSP_BACKEND
}
public enum class ComponentMergingBackend {
  EMBEDDED_MERGING,
  KSP_MERGING
}
```

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -113,6 +113,33 @@ public abstract class AnvilExtension @Inject constructor(
   public val trackSourceFiles: Property<Boolean> = objects.property(Boolean::class.java)
     .conventionFromProperty("com.squareup.anvil.trackSourceFiles", false)
 
+  /**
+   * Enables the new KSP backend for Anvil _source generation_. This is an experimental feature that
```
RBusarow, line 117: >  _source generation_.

I'm not clear on the distinction between these two options.  Technically, everything Anvil generates is source.

compiler-api/src/main/java/com/squareup/anvil/compiler/api/ComponentMergingBackend.kt
ZacSweers: I don't think that's quite right, as "embedded" in this case means two different things - AnalysisHandlerExtension and IR. I was thinking the specificity was better in this case since `AnvilBackend.EMBEDDED` will soon(er than IR) go away.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
ZacSweers: IR merging doesn't generate source? Unless you count in-memory IR or its downstream compiled bytecode as source 😅

compiler-api/src/main/java/com/squareup/anvil/compiler/api/ComponentMergingBackend.kt
ZacSweers: [abef3a1](https://github.com/square/anvil/pull/886/commits/abef3a153795c6ecf96fab909fa36bf6103a2268)

RBusarow: I was approaching this as just a question of whether it's executing as part of KGP (ish) or KSP.  Except KGP isn't really correct either.

I have no strong feeling about it, though, and I'm comfortable with the `IR` name.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
RBusarow: > in-memory IR

I guess I do?  I've never really questioned that. 🤷 

> compiled bytecode

no :)

RBusarow (approved):

ZacSweers:
The API change is real but feels ok to say since it wasn't usable before anyway?

PR #895, opened 2024-03-06
add ClassGraph support in tests

RBusarow:
[ClassGraph](https://github.com/classgraph/classgraph)

It's nicer ergonomics than using reflection APIs.


RBusarow:
folding this into PR #910

PR #901, opened 2024-03-09
Auto wire KSP when running in KSP mode

ZacSweers:
This propagates values from `AnvilExtension` to KSP args automatically when KSP mode is enabled, easing use in projects. Otherwise, KSP processors end up being inert since they require reading an `AnvilContext` from these args. This also automatically adds the Anvil compiler to the corresponding KSP configuration.

Most of this logic is adapted from what's been working well in Moshi-IR's plugin: https://github.com/ZacSweers/MoshiX/blob/main/moshi-ir/moshi-gradle-plugin/src/main/kotlin/dev/zacsweers/moshix/ir/gradle/MoshiGradleSubplugin.kt#L30-L69


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
```diff
@@ -93,6 +103,58 @@ internal open class AnvilPlugin : KotlinCompilerPluginSupportPlugin {
         getConfiguration(target, "main").extendsFrom(commonConfiguration)
       }
     }
+
+    target.afterEvaluate {
```
ZacSweers, line 107: This isn't my favorite. Another alternative is we could maybe exchange the properties for a function that calls through to this setup synchronously

ZacSweers: We may need to do the synchronous approach actually. While testing this, I encoutnered this issue https://github.com/google/ksp/issues/1789. Alternatively, we could move to just using static gradle properties only

ZacSweers: nvm, I think this is probably enough for now, and advanced cases like ours can just apply ksp eagerly if need be

RBusarow: The function would be my choice.  `afterEvaluate { }` is the worst.

We could still have Gradle properties, just not in the DSL.  Then the plugin could just explicitly call the function if they're toggled.

ZacSweers: alright let me see what I can do!

ZacSweers: Alright took a crack at it, how's this look? https://github.com/square/anvil/pull/901/commits/fc18e8c144c3f87856487663155554d8ee79f859

RBusarow (approved):
Aside from some curious formatting, LGTM. 👍

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -114,31 +138,92 @@ public abstract class AnvilExtension @Inject constructor(
...
+            "ksp${
+              it.targetName.replaceFirstChar {
+                if (it.isLowerCase()) {
+                  it.titlecase(
+                    Locale.US,
+                  )
+                } else {
+                  it.toString()
+                }
+              }
```
RBusarow, line 203: Are you okay, KtLint? 😓

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilPlugin.kt
ZacSweers: Hmm, seems the deferred cli args breaks CC. Let me see if I can improve this

ZacSweers: Fixed in https://github.com/square/anvil/pull/901/commits/cf56054c9fa621cefd0a4527db440838f821b7df

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -160,13 +160,16 @@ public abstract class AnvilExtension @Inject constructor(
    * com.squareup.anvil.useKspComponentMergingBackend=true
    * ```
    *
+   * @param project the Gradle project this is operating on. Annoying but necessary to pass this
```
RBusarow, line 163: 😁

You can make `Project` a receiver, so that it's invisible to the DSL.

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
ZacSweers: I don't like everything ktfmt does but definitely found it more reliable than ktlint heh

gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
ZacSweers: I thought about that, but are you worried about groovy accessors?

RBusarow: I am now 😓

PR #902, opened 2024-03-10
Fix contributed interface merging when KSP backend is enabled

ZacSweers:
I had assumed that `ModuleMergerIr` somewhere also handled merging contributed interfaces, but that's actually still done separately in `InterfaceMerger`. With the KSP controls, we currently don't add this impl if we enable the KSP backend. Easy enough fix for now is to move it back up to before, but this'll be another thing we need to do to fully support K2 (i.e. `InterfaceMergerIr`).


ZacSweers:
Superseded by #907

PR #907, opened 2024-03-11
Migrate interface merging to IR

### LINKED ISSUE #906: Migrate interface merging in IR
This'll be necessary for K2 support + also hopefully faster. I've got a WIP prototype ready once #877 is in

ZacSweers:
Resolves #906. More or less a straight re-implementation of the existing impl with changes made to use IR analogues as needed.


compiler/src/main/java/com/squareup/anvil/compiler/InterfaceMergerIr.kt
```diff
@@ -0,0 +1,276 @@
...
+      .mapNotNull { contributedSubcomponent ->
+        contributedSubcomponent.classId
+          .generatedAnvilSubcomponent(clazz.classId)
+          .createNestedClassId(Name.identifier(PARENT_COMPONENT))
+          .referenceClassOrNull(pluginContext)
+      }
+  }
+}
+
+// TODO is this necessary in IR? Think all supertypes should already be resolved
```
ZacSweers, line 246: I'm checking in kotlin-lang to see if this is still necessary, I suspect it's not

ZacSweers: Still necessary, it's only direct supertypes by default

RBusarow (approved):

PR #910, opened 2024-03-11
rework incremental cache restoration

RBusarow:
- Use all cached "root source files" as the basis for cache restoration, instead of using the list of `KtFile`s passed in by the compiler.  The file list from the compiler can be incremental, not including all extant source files.  With the incomplete list, any missing files that were generated from an unchanged source file would not be restored.
- Add all restored files to the list of `KtFile`s passed to code generators, to be treated as source files.
- Add a `BiMap<RelativeFile /* File */, FileId /* Int */>` to cached data and use those IDs in the other maps, to shrink the binary size a bit.


JoelWilcox (approved):
lgtm

PR #922, opened 2024-03-13
treat any input `KtFile`s as changed/invalidated

### LINKED ISSUE #898: New incremental compilation fails when deleting a contributed subcomponent
Version 2.5.0-beta03

To reproduce

1. Add this to `sample/library`:
    ```
    @ContributesSubcomponent(Any::class, AppScope::class)
    interface DescriptionSubcomponent {
      fun fatherProvider(): FatherProvider
      fun motherProvider(): MotherProvider
    
    
      @ContributesSubcomponent.Factory
      interface Factory {
        fun create(): DescriptionSubcomponent
      }
    
      interface ParentComponent {
        fun descriptionSubcomponentFactory(): Factory
      }
    }
    ```
2. Run `./gradlew :delegate:sample:app:assembleDebug -Pcom.squareup.anvil.trackSourceFiles=true`
3. Remove the contributed subcomponent again
4. Run `./gradlew :delegate:sample:app:assembleDebug -Pcom.squareup.anvil.trackSourceFiles=true`

Step 4 fails with this error
```
> Task :delegate:sample:app:kaptDebugKotlin FAILED
/Users/gabrielittner/Downloads/anvil/sample/app/build/tmp/kapt3/stubs/debug/anvil/component/com/squareup/anvil/sample/appcomponent/DescriptionSubcomponentA.java:30: error: @Binds methods' parameter type must be assignable to the return type - anvil.component.com.squareup.anvil.sample.appcomponent.DescriptionSubcomponentA.SubcomponentModule.bindSubcomponentFactory(anvil.component.com.squareup.anvil.sample.appcomponent.DescriptionSubcomponentA.SubcomponentFactory)
        public abstract error.NonExistentClass bindSubcomponentFactory(@org.jetbrains.annotations.NotNull()
                                               ^
/Users/gabrielittner/Downloads/anvil/sample/app/build/tmp/kapt3/stubs/debug/anvil/component/com/squareup/anvil/sample/appcomponent/DescriptionSubcomponentA.java:17: error: @Subcomponent.Factory types must have exactly one method that returns the @Subcomponent type - anvil.component.com.squareup.anvil.sample.appcomponent.DescriptionSubcomponentA.SubcomponentFactory
    public static abstract interface SubcomponentFactory {
```

It looks like the code in the library module is properly cleaned up but the code that the app module generates isn't. Deleting `sample/app/build` and then running again fixes the issue.

RBusarow:
fixes #898

By "input `KtFile`s" I'm referring to the `files: List<KtFile>` parameters in `CodeGeneratorExtension.doAnalysis(...)` and `CodeGeneratorExtension.onAnalysisCompleted(...)`.

For any given compilation, on the very first round of analysis/generation, the `files` parameter will roughly correspond to the incremental files from Gradle's `InputChanges` instance.

We should treat these files as changed because:

1. If they're also in our internal cache, with the same contents (MD5 hash), then there must be some reason that Gradle decided to pass them as input again and the compiler agreed. This is most likely because of a classpath change, either in a dependency project or an external library.

   In this case, since Anvil uses classpath discovery to generate binding modules and merged interfaces, its expected output can easily change even if the source files in that module haven't.

2. If they come from `InputChanges`, they're changed by definition.

3. If they're input files, then we'll be generating code for them again anyway and it will be up-to-date.

With this change, a source file is considered to be unchanged (eligible for restoration) if it exists on the disk and in the cache with the same content hash and it wasn't passed as an input file.  Since they aren't inputs, they won't be passed to the code generators, and we need to restore any downstream generated files from the cache.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
@@ -140,3 +140,35 @@ internal fun File.requireDelete() {
     throw AnvilCompilationException("Could not delete file: $this")
   }
 }
+
+/**
+ * Transforms the elements of the receiver collection and adds the results to a set.
+ *
+ * @param destination The destination set where the transformed
+ *   elements are placed. By default, it is an empty mutable set.
+ * @param transform Maps elements of the receiver collection to the output set.
+ * @receiver The collection to be transformed.
+ * @return A set containing the transformed elements from the receiver collection.
+ */
+internal inline fun <C : Collection<T>, T, R> C.mapToSet(
+  destination: MutableSet<R> = mutableSetOf(),
+  transform: (T) -> R,
+): Set<R> {
+  return mapTo(destination, transform)
+}
+
+/**
+ * Transforms the elements of the receiver array and adds the results to a set.
+ *
+ * @param destination The destination set where the transformed
+ *   elements are placed. By default, it is an empty mutable set.
+ * @param transform Maps elements of the receiver collection to the output set.
+ * @receiver The array to be transformed.
+ * @return A set containing the transformed elements from the receiver array.
+ */
+internal inline fun <T, R> Array<T>.mapToSet(
+  destination: MutableSet<R> = mutableSetOf(),
+  transform: (T) -> R,
+): Set<R> {
+  return mapTo(destination, transform)
+}
--- a/gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
+++ b/gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
@@ -1,6 +1,7 @@
 package com.squareup.anvil.plugin
 
 import com.rickbusarow.kase.Kase1
+import com.rickbusarow.kase.asClueCatching
 import com.rickbusarow.kase.files.DirectoryBuilder
 import com.rickbusarow.kase.gradle.dsl.buildFile
 import com.rickbusarow.kase.gradle.getValue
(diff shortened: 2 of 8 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
```diff
@@ -140,3 +140,35 @@ internal fun File.requireDelete() {
...
+/**
+ * Transforms the elements of the receiver collection and adds the results to a set.
+ *
+ * @param destination The destination set where the transformed
+ *   elements are placed. By default, it is an empty mutable set.
+ * @param transform Maps elements of the receiver collection to the output set.
+ * @receiver The collection to be transformed.
+ * @return A set containing the transformed elements from the receiver collection.
+ */
+internal inline fun <C : Collection<T>, T, R> C.mapToSet(
```
ZacSweers, line 153: Really wish this was in the stdlib sometimes heh

gradle-plugin/src/gradleTest/java/com/squareup/anvil/plugin/IncrementalTest.kt
```diff
@@ -557,6 +560,158 @@ class IncrementalTest : BaseGradleTest() {
       )
     }
 
+  @TestFactory
+  fun `a generated merged Subcomponent is deleted when a dependency module's @ContributesSubcomponent type is deleted`() =
```
ZacSweers, line 564: 👌

ZacSweers (approved):

compiler/src/main/java/com/squareup/anvil/compiler/Utils.kt
RBusarow: > sometimes

Yeah, pretty much every time I have to type it out manually.  Then after a handful of times I'll usually give in and add it. 😀

PR #926, opened 2024-03-14
Don't use internal AGP API

ZacSweers:
This just implements it in AnvilExtension instead


gradle-plugin/src/main/java/com/squareup/anvil/plugin/AnvilExtension.kt
```diff
@@ -250,6 +249,11 @@ public abstract class AnvilExtension @Inject constructor(
       .orElse(default),
   )
 
+  private fun <T> Property<T>.setDisallowChanges(value: T?) {
```
RBusarow, line 252: This is the way.

RBusarow (approved):

PR #931, opened 2024-03-15
fix interface based @ContributesSubcomponent.Factory in KSP

gabrielittner:
Currently any `@ContributesSubcomponent.Factory` that is an interface fails the KSP codegen with "A factory must have exactly one abstract function returning the subcomponent ...".  The `getAllFunctions()` returns non abstract methods like equals/hashCode for interfaces, so we need to filter for interfaces as well.


ZacSweers (approved):
Thanks!

PR #942, opened 2024-03-27
update `ij_kotlin.*` .editorconfig properties to match KtLint behavior

RBusarow:
The biggest discrepancy between IDE format and KtLint format is regarding blank lines in `when` blocks.  The IDE wants to add them, but IntelliJ doesn't.  So, I wind up reverting a lot of formatting changes before every pull request. 😠

I ran IDE format on the whole project after the .editorconfig changes, and the diff was satisfyingly small.


.editorconfig
```diff
@@ -537,7 +538,7 @@ ij_kotlin_extends_list_wrap = on_every_item
 ij_kotlin_field_annotation_wrap = normal
 ij_kotlin_finally_on_new_line = false
 ij_kotlin_if_rparen_on_new_line = true
-ij_kotlin_import_nested_classes = true
+ij_kotlin_import_nested_classes = false
```
RBusarow, line 541: This wonderful little property makes the IDE import top-level types, then use the partially-qualified references in code.  So for example, you'll have references of `PropertyReference.Psi` instead of the ambiguous `Psi`.

.editorconfig
```diff
@@ -547,15 +548,18 @@ ij_kotlin_keep_first_column_comment = true
 ij_kotlin_keep_indents_on_empty_lines = false
 ij_kotlin_keep_line_breaks = true
 ij_kotlin_lbrace_on_next_line = false
+ij_kotlin_line_break_after_multiline_when_entry = false
```
RBusarow, line 551: This makes it consistent with KtLint's formatting.

.editorconfig
```diff
@@ -547,15 +548,18 @@ ij_kotlin_keep_first_column_comment = true
...
+ij_kotlin_method_annotation_wrap = off
 ij_kotlin_method_call_chain_wrap = on_every_item
 ij_kotlin_method_parameters_new_line_after_left_paren = true
 ij_kotlin_method_parameters_right_paren_on_new_line = true
 ij_kotlin_method_parameters_wrap = normal
-ij_kotlin_name_count_to_use_star_import = 999
-ij_kotlin_name_count_to_use_star_import_for_members = 999
+ij_kotlin_name_count_to_use_star_import = 2147483647
+ij_kotlin_name_count_to_use_star_import_for_members = 2147483647
+ij_kotlin_packages_to_use_import_on_demand = unset
```
RBusarow, line 562: No more accidental `java.util.*` imports.

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Contribution.kt
```diff
@@ -24,7 +24,6 @@ import dagger.Module
 import dagger.Provides
 import dagger.multibindings.IntoMap
 import dagger.multibindings.IntoSet
-import java.util.Comparator
```
RBusarow, line 27: KtLint didn't catch this because there's a `Comparator` reference at the bottom of the file.  It's actually referencing the built-in `kotlin.Comparator` typealias, but it seems KtLint errs on the side of caution.

JoelWilcox (approved):

PR #946, opened 2024-04-02
delay `@ContributesSubcomponent` generation until the last analysis rounds

RBusarow:
The `ContributesSubcomponentHandlerGenerator` is responsible for generating these:

```kotlin
@MergeSubcomponent(scope = Any::class)
public interface SubcomponentInterfaceA : SubcomponentInterface {
  // ...
}
```

It used to generate those interfaces as soon as it scanned the triggers, but that can cause problems if another `@ContributesSubcomponent` interface is being generated, and that intends to replace a human-written one.

Now, the handler is a `PrivateCodeGenerator` and it doesn't start parsing/generating until all the "public" code generation is done.  This means it gets the full picture of the possible graph, but it still finishes generating before any IR merging happens.

### DIFF
--- a/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
+++ b/compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
@@ -89,10 +89,10 @@ internal class CodeGenerationExtension(
 
     // Either sync the local file state with our internal cache,
     // or delete any existing generated files (if caching isn't enabled).
-    val createdChanges = syncGeneratedDir(files)
+    val syncCreatedChanges = syncGeneratedDir(files)
     didSyncGeneratedDir = true
 
-    if (createdChanges) {
+    if (syncCreatedChanges) {
       // Tell the compiler to analyze the generated files *before* calling `analysisCompleted()`.
       // This ensures that the compiler will update/delete any references to stale code in the
       // ModuleDescriptor, as well as updating/deleting any stale .class files.
@@ -118,10 +118,8 @@ internal class CodeGenerationExtension(
     anvilModule.addFiles(files)
 
     val generatedFiles = generateCode(
-      codeGenerators = codeGenerators,
       psiManager = psiManager,
       anvilModule = anvilModule,
-      files = files,
     )
 
     if (trackSourceFiles) {
@@ -189,10 +187,8 @@ internal class CodeGenerationExtension(
   }
 
   private fun generateCode(
-    codeGenerators: List<CodeGenerator>,
     psiManager: PsiManager,
     anvilModule: RealAnvilModuleDescriptor,
-    files: Collection<KtFile>,
   ): MutableCollection<FileWithContent> {
 
     val anvilContext = commandLineOptions.toAnvilContext(anvilModule)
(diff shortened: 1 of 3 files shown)


compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -118,10 +118,8 @@ internal class CodeGenerationExtension(
     anvilModule.addFiles(files)
 
     val generatedFiles = generateCode(
-      codeGenerators = codeGenerators,
```
RBusarow, line 121: This is just class-level, so no need to pass the variable around internally.

ZacSweers (approved):
lgtm though I don't think this fixes the issue @gabrielittner was reporting as his was KSP-specific

gabrielittner:
> though I don't think this fixes the issue @gabrielittner was reporting as his was KSP-specific

gabrielittner:
> though I don't think this fixes the issue @gabrielittner was reporting as his was KSP-specific

Yes, it's still happening with this change. The issue is that since there is no KSP implementation of `ContributesSubcomponentHandlerGenerator` the `SubcomponentInterfaceA` types won't be generated at all when enabling KSP mode.

gabrielittner:
@RBusarow This change seems to break custom code generators in embedded mode. https://github.com/freeletics/khonshu on branch `anvil-snapshots`  if you run `./gradlew -p sample/simple assembleDebug` there is no output from custom generators. The build does not fail but checking for example `sample/simple/feature/main/implementation/build/anvil` there should be a generated file called `KhonshuMainScreen` which doesn't exist. The branch is using the latest snapshot but I've locally tested that this change is the one that broke it (commit `a3cf4876` is still working, `95ade2c1` is broken).

compiler/src/main/java/com/squareup/anvil/compiler/codegen/CodeGenerationExtension.kt
```diff
@@ -227,47 +223,56 @@ internal class CodeGenerationExtension(
...
-        }
-
-    var newFiles = nonPrivateCodeGenerators.generateCode(files)
-
-    while (newFiles.isNotEmpty()) {
-      // Parse the KtFile for each generated file. Then feed the code generators with the new
-      // parsed files until no new files are generated.
-      newFiles = nonPrivateCodeGenerators.generateCode(newFiles)
+    fun List<CodeGenerator>.loopGeneration() {
+      var newFiles = generateAndCache(anvilModule.allFiles.toList())
```
gabrielittner, line 254: Previously `generateAndCache` (which was called `generateCode` before) was called once with `files` (like 258 of the removed code `var newFiles = nonPrivateCodeGenerators.generateCode(files)`). Always using `anvilModule.allFiles.toList()` can break custom code generators that use [topLevelFunctionReferences](https://github.com/square/anvil/blob/main/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt#L73-L80) or [topLevelPropertyReferences](https://github.com/square/anvil/blob/main/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/AnvilModuleDescriptor.kt#L82-L89) because it won't contain any `KtFile` that doesn't contain a `KtClassOrObject`. So a `KtFile` that only contains top level functions and/or properties will not be passed to code generators anymore and because of that `topLevel*References` won't include functions/properties from those files.

PR #974, opened 2024-04-19
Consolidate hint packages

ZacSweers:
Needs to figure out rebasing since my fork got borked by the force push


ZacSweers:
Continued in #975

PR #987, opened 2024-04-26
shorter names for binding modules and hints

RBusarow:


compiler/src/main/java/com/squareup/anvil/compiler/codegen/Contribution.kt
```diff
@@ -197,6 +197,40 @@ internal sealed class Contribution {
...
+      val md5Hash = MessageDigest.getInstance("MD5")
+        .apply {
+          update(originType.canonicalName.toByteArray())
+          update(boundType.canonicalName.toByteArray())
+          update(scopeType.canonicalName.toByteArray())
+          update(suffix.toByteArray())
+        }
+        .digest()
+        .take(8)
+        .joinToString("") { "%02X".format(it) }
```
ZacSweers, line 217: sidenote - okio has some nice APIs for this

ZacSweers (approved):

r0adkll (approved):

compiler/src/main/java/com/squareup/anvil/compiler/codegen/Contribution.kt
RBusarow: I've been tempted to add Okio, and also kotlinx-serialization or Moshi for the incremental caching.  But that would mean adding things to the compiler classpath and/or adding the shadow plugin to our build, and it just didn't seem to be worth the hassle.

PR #1033, opened 2024-07-16
consider `ignoreQualifier` arguments when checking for duplicate bindings

### LINKED ISSUE #986: Duplicate `ContributesMultibinding` check does not consider `ignoreQualifier`
Having `ContributesMultibinding` once without `ignoreQualifier` and once with results in an error that the contributions need to be distinct.
```
@Named("test")
@ContributesMultibinding(AppScope::class, DeepLinkHandler::class)
@ContributesMultibinding(AppScope::class, DeepLinkHandler::class, ignoreQualifier = true)
class FooDeepLink @Inject constructor() : DeepLinkHandler {}
```

```
e: com.squareup.anvil.compiler.api.AnvilCompilationException: Back-end (JVM) Internal error: com.freeletics.feature.foo.FooDeepLinkHandler contributes multiple times to the same scope using the same bound type: [DeepLinkHandler]. Contributing multiple times to the same scope with the same bound type is forbidden and all scope - bound type combinations must be distinct.
```

The distinct check should consider `ignoreQualifier` as well since it results in 2 distinct contributions.

RBusarow:
fixes #986

### DIFF
--- a/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
@@ -4,6 +4,7 @@ import com.google.common.truth.Truth.assertThat
 import com.squareup.anvil.annotations.MergeComponent
 import com.squareup.anvil.annotations.MergeSubcomponent
 import com.squareup.anvil.annotations.compat.MergeInterfaces
+import com.squareup.anvil.annotations.compat.MergeModules
 import com.squareup.anvil.compiler.anyQualifier
 import com.squareup.anvil.compiler.componentInterface
 import com.squareup.anvil.compiler.contributingInterface
@@ -23,7 +24,7 @@ import javax.inject.Named
 class BindingModuleQualifierTest : AnvilAnnotationsTest(
   MergeComponent::class,
   MergeSubcomponent::class,
-  MergeInterfaces::class,
+  MergeModules::class,
 ) {
 
   @TestFactory fun `the Dagger binding method has a qualifier without parameter`() = testFactory {
--- a/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
+++ b/compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
@@ -3,51 +3,30 @@ package com.squareup.anvil.compiler.codegen
 import com.google.common.truth.Truth.assertThat
 import com.squareup.anvil.annotations.MergeComponent
 import com.squareup.anvil.annotations.MergeSubcomponent
-import com.squareup.anvil.annotations.compat.MergeModules
+import com.squareup.anvil.annotations.compat.MergeInterfaces
 import com.squareup.anvil.compiler.anyQualifier
-import com.squareup.anvil.compiler.compile
 import com.squareup.anvil.compiler.componentInterface
 import com.squareup.anvil.compiler.contributingInterface
 import com.squareup.anvil.compiler.generatedBindingModule
 import com.squareup.anvil.compiler.generatedMultiBindingModule
 import com.squareup.anvil.compiler.internal.testing.isAbstract
-import com.squareup.anvil.compiler.isFullTestRun
 import com.squareup.anvil.compiler.mergedModules
 import com.squareup.anvil.compiler.parentInterface
 import com.squareup.anvil.compiler.subcomponentInterface
+import com.squareup.anvil.compiler.testing.AnvilAnnotationsTest
 import dagger.Binds
 import dagger.Provides
 import dagger.multibindings.IntoSet
-import org.junit.Test
-import org.junit.runner.RunWith
-import org.junit.runners.Parameterized
-import org.junit.runners.Parameterized.Parameters
+import org.junit.jupiter.api.TestFactory
 import javax.inject.Named
-import kotlin.reflect.KClass
 
-@RunWith(Parameterized::class)
-class BindingModuleQualifierTest(
-  private val annotationClass: KClass<out Annotation>,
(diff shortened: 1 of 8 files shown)


compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.rickbusarow.kase.files.HasWorkingDir
+import com.rickbusarow.kase.files.TestLocation
+import com.rickbusarow.kase.kases
+import com.squareup.anvil.compiler.isFullTestRun
+import kotlin.reflect.KClass
+
+abstract class AnvilAnnotationsTest(
+  firstAnnotation: KClass<out Annotation>,
+  vararg fullTestRunAnnotations: KClass<out Annotation>,
+) : KaseTestFactory<
```
JoelWilcox, line 16: Is there a way to use `KaseTestFactory` via composition or a test rule instead of introducing a base class?

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
```diff
@@ -3,51 +3,30 @@ package com.squareup.anvil.compiler.codegen
...
-    fun annotationClasses(): Collection<Any> {
-      return buildList {
-        add(MergeComponent::class)
-        if (isFullTestRun()) {
-          add(MergeSubcomponent::class)
-          add(MergeModules::class)
-        }
-      }
-    }
-  }
```
JoelWilcox, line 48: Is the move to kase necessary here? I feel like the further these details move away from the test, the harder it is to reason about and maintain vs keeping it simple and using the Junit apis directly. E.g. it's no longer clear that this test suite is impacted by the `isFullTestRun()` check

compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.rickbusarow.kase.KaseTestFactory
+import com.rickbusarow.kase.ParamTestEnvironmentFactory
+import com.rickbusarow.kase.files.HasWorkingDir
+import com.rickbusarow.kase.files.TestLocation
+import com.rickbusarow.kase.kases
+import com.squareup.anvil.compiler.isFullTestRun
+import kotlin.reflect.KClass
+
+abstract class AnvilAnnotationsTest(
+  firstAnnotation: KClass<out Annotation>,
```
JoelWilcox, line 14: Rename to something like `requiredAnnotation`? To help signify what it's used for and why it's separate from `fullTestRunAnnotations`

compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
```diff
@@ -0,0 +1,56 @@
...
+import com.rickbusarow.kase.Kase1
+import com.rickbusarow.kase.KaseTestFactory
+import com.rickbusarow.kase.ParamTestEnvironmentFactory
+import com.rickbusarow.kase.files.HasWorkingDir
+import com.rickbusarow.kase.files.TestLocation
+import com.rickbusarow.kase.kases
+import com.squareup.anvil.compiler.isFullTestRun
+import kotlin.reflect.KClass
+
+abstract class AnvilAnnotationsTest(
```
JoelWilcox, line 13: Maybe add some KDOC for this and the intended use cases

JoelWilcox (approved):

compiler/src/test/java/com/squareup/anvil/compiler/codegen/BindingModuleQualifierTest.kt
RBusarow: It's not *strictly* necessary.  Most of these tests were failing at some point during or after `ignoreQualifier` change, and it's much easier to debug them when you have the consistent working directories from Kase.  Any time I run into a similar situation with JUnit 4 tests, I've been migrating them over to Kase.

I would also argue that the mechanism for parameterized tests in JUnit 4 is more magical than simple -- at least for me.

compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
RBusarow: No to the Rule, technically yes to the base class.  But the intent is for it to be used for the other tests with annotation parameters as well (there are five `Binding___Test` in total).  The setup for each is identical except for those parameters.

compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
RBusarow: Done: https://github.com/square/anvil/pull/1033/commits/4df2fb9e48bc26d9b0a30ec5058b0b023408b9ce

compiler/src/test/java/com/squareup/anvil/compiler/testing/AnvilAnnotationsTest.kt
```diff
@@ -0,0 +1,92 @@
...
+import com.rickbusarow.kase.KaseTestFactory
+import com.rickbusarow.kase.ParamTestEnvironmentFactory
+import com.rickbusarow.kase.files.HasWorkingDir
+import com.rickbusarow.kase.files.TestLocation
+import com.rickbusarow.kase.kases
+import com.squareup.anvil.compiler.isFullTestRun
+import kotlin.reflect.KClass
+
+/**
+ * Creates test factories to run the same test against multiple annotations.
```
JoelWilcox, line 14: Doc comment looks great!

PR #1038, opened 2024-07-19
include inherited functions in Subcomponent Factory checks

### LINKED ISSUE #1003: `ContributesSubcomponent.Factory` doesn't support inherited functions
The document said that

> This follows all the rules of `Subcomponent.Factory`, except it must appear in classes
> annotated with `@ContributesSubcomponent` instead of `@Subcomponent`.

But `ContributesSubcomponent.Factory` actually doesn't support inherited function.

```kotlin
package com.squareup.test

import com.squareup.anvil.annotations.ContributesSubcomponent
import com.squareup.anvil.annotations.ContributesSubcomponent.Factory
import com.squareup.anvil.annotations.ContributesTo
import com.squareup.anvil.annotations.MergeComponent

@ContributesSubcomponent(Any::class, parentScope = Unit::class)
interface SubcomponentInterface {
    @ContributesTo(Unit::class)
    interface AnyParentComponent {
        fun createFactory(): ComponentFactory
    }

    interface Creator {
        fun createComponent(): SubcomponentInterface
    }

    @Factory
    abstract class ComponentFactory : Creator
}

@MergeComponent(Unit::class)
interface ComponentInterface
```

The code above will fail if we're using `@ContributesSubcomponent.Factory` but will pass for `@Subcomponent.Factory`

Another thing is, while I haven't tested it, the KSP part of the code gen might support the above code, since [in here](https://github.com/square/anvil/blob/main/compiler/src/main/java/com/squareup/anvil/compiler/codegen/ContributesSubcomponentCodeGen.kt#L205-L207) it use `getAllFunctions` that will returns declared and inherited functions.

RBusarow:
fixes #1003

Our checks for a single abstract method in a `@ContributesSubcomponent.Factory`-annotated interface do not consider functions that are defined in a supertype.  The existing `ClassReference.functions` property was somewhat ambiguously named, which probably contributed to the error.

### DIFF
--- a/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
+++ b/compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
@@ -66,8 +66,49 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
   public val packageFqName: FqName get() = classId.packageFqName
 
   public abstract val constructors: List<MemberFunctionReference>
+
+  @Deprecated(
+    "renamed to `declaredMemberFunctions`.  " +
+      "Use `memberFunctions` to include inherited functions.",
+    replaceWith = ReplaceWith("declaredMemberFunctions"),
+  )
   public abstract val functions: List<MemberFunctionReference>
+
+  /**
+   * All functions that are declared in this class, including overrides.
+   * This list does not include inherited functions that are not overridden by this class.
+   */
+  public abstract val declaredMemberFunctions: List<MemberFunctionReference>
+
+  /**
+   * All functions declared in this class or any of its super-types.
+   */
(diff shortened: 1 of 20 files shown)


compiler-utils/src/main/java/com/squareup/anvil/compiler/internal/reference/ClassReference.kt
```diff
@@ -66,8 +66,49 @@ public sealed class ClassReference : Comparable<ClassReference>, AnnotatedRefere
...
   public abstract val properties: List<MemberPropertyReference>
+
+  /**
+   * All properties that are declared in this class, including overrides.
+   * This list does not include inherited properties that are not overridden by this class.
+   */
+  public abstract val declaredMemberProperties: List<MemberPropertyReference>
+
+  /**
+   * All properties declared in this class or any of its super-types.
```
JoelWilcox, line 105: Nice, the new api and comments are very clear

JoelWilcox (approved):

PR #1074, opened 2024-12-03
Hide api for enabling KSP for upcoming release because it's currently only a partial implementation in the main Anvil repo

RBusarow:
This is grabbed from the [Kotlin 2.0.21](https://github.com/square/anvil/tree/joel.kotlin-2.0.21) branch, since we want to disable the KSP apis in 2.5.0 as well.


RBusarow:
cc @JoelWilcox

JoelWilcox (approved):

PR #1075, opened 2024-12-03
Add Anvil ROADMAP doc

JoelWilcox:


docs/ROADMAP.md
```diff
@@ -0,0 +1,47 @@
...
+2. Anvil 3.0 \- Introduce Anvil Component generation, remove dependency on both Dagger and KAPT. At this point Anvil will become a completely independent Kotlin compiler plugin which broadly follows the existing contract and API surface. We don't currently have plans for KMP support since it's not something we need internally, but it's something we're open to re-evaluating after the release of Anvil 3.0.  
+3. Long-tail: Evolve the API to match new capabilities and take advantage of our lack of dependency on Dagger, with a focus on ease of migration for consumers.
+
+# **Context**
+
+This section focuses on providing context for how we ended up with the roadmap items above and the decisions made along the way.
+
+## **Breaking from Dagger**
+
+The decision to break our dependency on Dagger stems from a number of constraints Anvil is limited by today, including:
```
jingibus, line 21: Error reporting, too, right? Library users will always get Dagger errors, which they then have to interpret alongside what might be happening in Anvil

jingibus (approved):

docs/ROADMAP.md
JoelWilcox: Good catch! Added a point for error reporting

docs/ROADMAP.md
```diff
@@ -0,0 +1,48 @@
...
+
+# **Roadmap**
+
+The long term vision is currently bucketed into three major milestones:
+
+1. Introduce support for FIR and the K2 compiler. This will tentatively happen in two phases:  
+   1. K2 baseline support for all primary Anvil functionality (contributions and merging), with KAPT utilized instead of KSP. (For those that need Dagger-KSP interop or K2 support now, we endorse Zac Sweers's Anvil-KSP fork: [https://github.com/ZacSweers/anvil](https://github.com/ZacSweers/anvil).)  
+   2. K2 support for generating dagger factories.  
+2. Anvil 3.0 \- Introduce Anvil Component generation, remove dependency on both Dagger and KAPT. At this point Anvil will become a completely independent Kotlin compiler plugin which broadly follows the existing contract and API surface. We don't currently have plans for KMP support since it's not something we need internally, but it's something we're open to re-evaluating after the release of Anvil 3.0.
```
bddckr, line 12: <sup>If you consider this to be out of scope to discuss here, feel free to let me know. No hard feelings 😄</sub>

Are there any thoughts on the recently emerged Amazon project https://github.com/amzn/kotlin-inject-anvil Perhaps Anvil maintainers consider `kotlin-inject` itself, or `kotlin-inject-anvil` on top of it to be quite different from what Anvil 3.0 will be?

I'm curious about it in the context of Anvil 3.0, as I'd like to understand long-term vision differences (if any) and whether there's opportunity the three projects to join forces. The good old "which horse will I bet on" question 😅

vRallev: I can't speak for Square anymore, but conceptional I expect Anvil 3.0 to be similar to the existing Anvil 2.0 with Dagger 2. `kotlin-inject` has some limitations, e.g. it requires KSP which makes builds slower. How much in the current environments I haven't measured. But what `kotlin-inject-anvil` provides today is KMP support. For my current projects that's required and I cannot go back to Dagger and Anvil and likely Anvil 3.0 neither. So `kotlin-inject-anvil` will be funded for a long time :-)

That being said, what works best for you depends on your requirements. Can you wait for Anvil 3.0? Is Anvil 2.0 blocking Kotlin 2.X for you? How fast do you need to upgrade? Do you see problems with incremental compilation? Do you want to use Kotlin Multiplatform?

bddckr: Thanks for all the explanations!

This makes a lot of sense, and continues to differentiate Anvil compared to every other DI solution available right now.

I've personally been fine on the latest beta and just updated to 2.5.0. No interest in updating to the latest K2 versions just yet.

Looking forward to the road to 3.0, thanks a lot!

RBusarow (approved):

PR #1107, opened 2025-02-03
initial k2 base classes and k2-compatible testing support

RBusarow:


compiler-k2/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
+plugins {
+  id("conventions.library")
+  id("conventions.publish")
+  kotlin("kapt")
```
JoelWilcox, line 4:
```suggestion
 alias(libs.plugins.kotlin.kapt)
```

compiler-k2/build.gradle.kts
```diff
@@ -0,0 +1,35 @@
...
+}
+
+conventions {
+  explicitApi = true
+}
+
+publish {
+  configurePom(
+    artifactId = "compiler-k2",
+    pomName = "Anvil Compiler Implementation for Kotlin 2+",
```
JoelWilcox, line 14: Might be worth clarifying this as the K2 compiler (or Kotlin languageLevel 2) specifically since the existing implementation also supports Kotlin 2.

```suggestion
    pomName = "Anvil Compiler Implementation for the K2 compiler",
```

(same for the name in compiler-k2-api)

compiler-testing/build.gradle.kts
```diff
@@ -0,0 +1,59 @@
...
+    // compiler requires this flag when doing so.
+    "-Xjvm-default=all",
+    "-opt-in=org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi",
+  )
+  addTasksToIdeSync("generateBuildConfig")
+}
+
+publish {
+  configurePom(
+    artifactId = "compiler-k2-testing",
```
JoelWilcox, line 32: Should the module name + build config package name also reflect this? Current name could cause confusion as being non-K2 or supporting both K1 and K2

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CompilationModeTest.kt
```diff
@@ -0,0 +1,76 @@
...
+public sealed interface CompilationMode {
+  public val useKapt: Boolean
+
+  public data class K2(
+    override val useKapt: Boolean,
+
+  ) : CompilationMode
+
+  @Deprecated("K1 is not implemented yet", level = DeprecationLevel.ERROR)
+  public data class K1(
```
JoelWilcox, line 43: Is this needed? Probably relates to my other question of whether this module is intended to just be for K2 testing, but also this appears to be unused for now

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CompilationModeTest.kt
```diff
@@ -0,0 +1,76 @@
...
+    override val useKapt: Boolean,
+
+  ) : CompilationMode {
+    init {
+      error("K1 is not implemented yet")
+    }
+  }
+}
+
+@Execution(ExecutionMode.SAME_THREAD, reason = "KSP tests aren't thread-safe yet.")
```
JoelWilcox, line 53: Is this still needed now that all of the KSP logic has been removed?

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/Compile2.kt
```diff
@@ -0,0 +1,471 @@
...
+import java.io.PrintStream
+import java.net.URI
+import java.net.URLClassLoader
+import javax.tools.Diagnostic
+import javax.tools.DiagnosticCollector
+import javax.tools.JavaFileObject
+import javax.tools.SimpleJavaFileObject
+import javax.tools.ToolProvider
+
+public fun CompilationEnvironment.compile2(
```
JoelWilcox, line 36: I think this would be a good place to document why we have compile2 and any issues encountered with KCT / how compile2 resolves them, especially for folks who may run into those issues in the future when updating their custom code generator tests

JoelWilcox (commented):
First pass looks good so far! Mostly left some notes on documentation and clarifying K1 vs K2 vs Kotlin 2 in a few places

compiler-testing/build.gradle.kts
RBusarow: The artifact id was out of date.  It's updated now to `compiler-testing`.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CompilationModeTest.kt
RBusarow: In the future, possibly.  It's intended to help in adding K2 support to all the existing tests.

japplin (commented):
I think the changes look good but I'd prefer we avoid adding extra test utils and abstractions until we need them.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/Compile2.kt
```diff
@@ -0,0 +1,471 @@
...
+
+    return Compile2Result(
+      rootDir = config.rootDir,
+      libsDir = config.jarOutputDir,
+      classFilesDir = config.classFilesDir,
+      classpathFiles = config.compilationClasspath,
+      exitCode = exitCode,
+      classLoader = classLoader,
+    ).apply {
+
```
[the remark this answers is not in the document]
RBusarow: It is formatted per our .editorconfig

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/HostClasspath.kt
```diff
@@ -0,0 +1,236 @@
...
+          // ["build", "libs", "compiler-2.5.0-SNAPSHOT.jar"]
+          .takeLast(3)
+          // ["build", "libs"]
+          .dropLast(1)
+
+        libsDir == relativeJarParents
+      }
+  }
+
+  private fun anvilModuleJar(moduleName: String, afterVersion: String = ""): File {
```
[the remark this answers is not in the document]
RBusarow: No, it wouldn't, but that's getting pretty obscure.  For the specific use-case you're thinking of, that convention plugin doesn't re-map the build directories of included builds, so it wouldn't matter.

RBusarow:
> I think the changes look good but I'd prefer we avoid adding extra test utils and abstractions until we need them.

The overwhelming majority is moved or copied over from K1 testing utils.  I'm avoiding adding `compiler` or the test fixtures as dependencies to the k2 modules because they add services to the compiler classpath and adds lots of K1-specific code, so for now that means some duplication.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/Compile2.kt
```diff
@@ -0,0 +1,471 @@
...
+      // We use the string version so that it's source compatible.
+      if (kotlinVersion < "2.1.0") {
+        // aka `args.useKapt4 = true`
+        args.freeArgs += "-Xuse-kapt4"
+      } else {
+        // aka `args.useK2Kapt = true`
+        args.freeArgs += "-Xuse-k2-kapt"
+      }
+
+      val processingClasspath = listOf(
```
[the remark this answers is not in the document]
RBusarow: It's the kapt processor classpath.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/HostClasspath.kt
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  public val autoServiceAnnotations: File by lazy {
+    findInClasspath(group = "com.google.auto.service", module = "auto-service-annotations")
+  }
+
+  public val dagger: File by lazy {
+    findInClasspath(group = "com.google.dagger", module = "dagger")
+  }
+  public val daggerCompiler: File by lazy {
```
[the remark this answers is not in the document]
RBusarow: There are several different classpaths that need to be populated for each test compilation.  This class finds all the relevant jars and/or class files within your local environment.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/HostClasspath.kt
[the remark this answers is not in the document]
japplin: Ah right this is just for configuring the test files right? So that would mean a project ie. android-register which is using anvil would still be able to use the out dir?

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/Compile2.kt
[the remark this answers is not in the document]
japplin: While that is true, I think it is still quite abnormal to start any lambda block with a blank line.

japplin:
> > I think the changes look good but I'd prefer we avoid adding extra test utils and abstractions until we need them.
> 
> The overwhelming majority is moved or copied over from K1 testing utils. I'm avoiding adding `compiler` or the test fixtures as dependencies to the k2 modules because they add services to the compiler classpath and add lots of K1-specific code, so for now that means some duplication.

Yup I realize these are copy pasted over from the old tests and I agree duplicating this type of setup isn't a major concern, with the only caveat being it might become a challenge with sharing any tests between the k1 and k2 implementations. 

My main point here was more along the lines of generally not adding code which is unused since it may never be needed and it makes the core changes, establishing compile2, more difficult to review as well as the additional code itself. I'm not pushing for you to necessarily go back and remove it now but just a general point for future prs.

RBusarow:
> My main point here was more along the lines of generally not adding code which is unused since it may never be needed and it makes the core changes, establishing compile2, more difficult to review as well as the additional code itself. I'm not pushing for you to necessarily go back and remove it now but just a general point for future prs.

Right.  This PR is expressly for adding K2-compatible testing support.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CompilationEnvironment.kt
```diff
@@ -0,0 +1,141 @@
...
+  override val testEnvironmentFactory: DefaultTestEnvironment.Factory
+    get() = DefaultTestEnvironment.Factory()
+}
+
+public interface CompilationEnvironment : HasWorkingDir {
+  public val mode: CompilationMode
+    get() = CompilationMode.K2(useKapt = false)
+
+  /**
+   * A convenience overload of [compile2] that accepts raw Kotlin and Java source code strings.
```
japplin, line 32: These docs are really helpful

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/HostClasspath.kt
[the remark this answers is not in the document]
RBusarow: Yep.  Register would be unaffected, even when we're consuming Anvil as an included build.

japplin (approved):

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/Utils.kt
```diff
@@ -0,0 +1,109 @@
...
+internal const val COMPONENT_PACKAGE_PREFIX = "anvil.component"
+
+internal const val REFERENCE_SUFFIX = "_reference"
+internal const val SCOPE_SUFFIX = "_scope"
+
+/**
+ * Returns the single element matching the given [predicate], or `null` if element was not found.
+ * Unlike [singleOrNull] this method throws an exception if more than one element is found.
+ */
+internal inline fun <T> Iterable<T>.singleOrEmpty(predicate: (T) -> Boolean): T? {
```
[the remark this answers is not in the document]
RBusarow: It's actually used quite a bit in the rest of the project, but this copy wasn't necessary.  It's gone.

PR #1110, opened 2025-02-11
Generate dagger factories for @Inject constructor

japplin:
Generate top level factory classes for constructor injection in k2.


compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/constructor/inject/FirInjectConstructorFactoryGenerationExtensionTest.kt
```diff
@@ -0,0 +1,61 @@
...
+        val param0: String,
+        val param1: Int,
+      )
+      """.trimIndent(),
+    ) {
+      val expectedString = "ExpectedString"
+      val expectedInt = 77
+      val injectFactoryClass = classGraph.allClassesAsMap["foo.InjectClass_Factory"]!!.loadClass()
+      val factoryInstance = injectFactoryClass.constructors.first()
+        .newInstance(Provider { expectedString }, Provider { expectedInt })
```
JoelWilcox, line 53: Should we be calling the companion object's `create` method instead here to verify everything was generated, either here or as a separate test? I think right now all of the tests would continue to pass even if we were no longer generating the `create` method

JoelWilcox (approved):
Looks good overall, main request is increasing test coverage a bit to catch any regressions to the `create()` function. Other things are mostly organizational or topics for us to think about and discuss more

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/fir/GeneratedDeclarationsIrBodyFiller.kt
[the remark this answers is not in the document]
japplin: Ah good catch

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/fir/constructor/inject/FirInjectConstructorFactoryGenerationExtension.kt
```diff
@@ -0,0 +1,190 @@
...
+    return createMemberProperty(
+      owner = context.owner,
+      key = Key,
+      name = param.name,
+      returnType = param.resolvedReturnType,
+    ).apply {
+      // Assigns property values from the constructor arguments
+      replaceInitializer(
+        buildPropertyAccessExpression {
+          val b = this@buildPropertyAccessExpression
```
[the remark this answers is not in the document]
japplin: Ah yea this was just there from the prototype code and I missed cleaning it up.

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/constructor/inject/FirInjectConstructorFactoryGenerationExtensionTest.kt
japplin: Ah yes good call I'll update the test to use the static create method

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/fir/constructor/inject/InjectConstructorGenerationModel.kt
```diff
@@ -0,0 +1,161 @@
...
+import org.jetbrains.kotlin.fir.types.constructClassLikeType
+import org.jetbrains.kotlin.fir.types.constructType
+import org.jetbrains.kotlin.name.CallableId
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.Name
+
+@OptIn(ExperimentalTopLevelDeclarationsGenerationApi::class)
+internal class InjectConstructorGenerationModel(
+  private val extension: FirExtension,
+  private val session: FirSession,
```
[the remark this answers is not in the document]
JoelWilcox: Makes sense! I left it like this in the @ContributesBinding PR as well, but was able to make them just constructor args and not properties for now.

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/names/DaggerNames.kt
```diff
@@ -0,0 +1,7 @@
+package com.squareup.anvil.compiler.k2.names
+
+import org.jetbrains.kotlin.name.FqName
+
+internal object DaggerNames {
```
RBusarow, line 5: The `Names` object I created in the big POC branch was intended to aid discoverability with IDE auto-complete, since everything's hierarchal under `Names`.  It also has the benefit of not adding any new top-level types that start with Dagger, Anvil, or Java.

Would you mind continuing with that pattern instead?

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/ir/AnvilFirCompilerPluginRegistrar.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import com.google.auto.service.AutoService
+import org.jetbrains.kotlin.backend.common.extensions.IrGenerationExtension
+import org.jetbrains.kotlin.compiler.plugin.CompilerPluginRegistrar
+import org.jetbrains.kotlin.compiler.plugin.ExperimentalCompilerApi
+import org.jetbrains.kotlin.config.CompilerConfiguration
+
+@OptIn(ExperimentalCompilerApi::class)
+@AutoService(CompilerPluginRegistrar::class)
+public class AnvilIrCompilerPluginRegistrar : CompilerPluginRegistrar() {
```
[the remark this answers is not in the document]
japplin: I moved this out in response to this thread: https://github.com/square/anvil/pull/1110#discussion_r1951688455.
I guess I could have just moved renamed to not mention fir and moved it out of the fir package, I'll make that change now.
cc @JoelWilcox

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/names/DaggerNames.kt
japplin: I noticed that @JoelWilcox  pr also has this DaggerNames object: https://github.com/square/anvil/pull/1112/files#diff-c901a8b204aaf65a92fd66334072a95ae6bbbe36fa2ccf0ce57c65bcb16811f7. So we'll need to align on something across both pr's. My main reason for moving away from the Names.dagger pattern was not wanting to declare objects with lowercase names. I'm open to other options for organizing these names, maybe just one large flat hierarchy of Names.daggerBinds, Names.daggerModule, Names.javaInject etc.

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/ir/AnvilFirCompilerPluginRegistrar.kt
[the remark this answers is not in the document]
japplin: Oh I just reread Joels comment again and realized you are both suggesting the same thing

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/names/DaggerNames.kt
JoelWilcox: I would also like to avoid having lowecase-named objects, but otherwise don't have strong feelings on how we organize these names. The discoverability aspect makes sense to me :+1:

RBusarow: I also didn't/don't like creating lowercase objects, but I rationalize it away by telling myself it's a mini DSL, so the normal conventions don't really apply.

But if you guys would like to capitalize the inner objects, I can live with it.

JoelWilcox: Throwing this out as an option too which would still allow a more-DSL-esque style of accessing the different names, maintain easy discoverability, and allow using specific sets directly if you already know what you want:

Names.kt
```
object Names {
  val dagger = DaggerNames
  val otherGroup = OtherGroupNames
  ...
}
```

DaggerNames.kt
```
internal object DaggerNames {
  val factory = FqName("dagger.internal.Factory")
}
```

japplin: Oh that's an interesting idea, but I think it could lead to inconsistent usages with some call sites accessing DaggerNames.factory directly and some accessing it via Names.dagger.factory. I'm really leaning towards just keeping this really simple with just a single Names object and then having everything prefix with dagger/javax/anvil etc and sorting the vals alphabetically.

JoelWilcox: sounds good to me!

JoelWilcox (approved):

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/constructor/inject/FirInjectConstructorFactoryGenerationExtensionTest.kt
```diff
@@ -0,0 +1,63 @@
...
+      class InjectClass @javax.inject.Inject constructor(
+        val param0: String,
+        val param1: Int,
+      )
+      """.trimIndent(),
+    ) {
+      val expectedString = "ExpectedString"
+      val expectedInt = 77
+      val injectFactoryClass =
+        classGraph.allClassesAsMap["com.squareup.test.InjectClass_Factory"]!!.loadClass()
```
RBusarow, line 52: nit: You can just do `classLoader.loadClass(...)` for these.

`injectClassFactory` and `injectClass` are actually defined in `CommonNames` and give you the `Class<*>` instances.  I just forgot to make them public.

RBusarow (approved):

PR #1112, opened 2025-02-12
Generate Dagger modules for @ContributesBinding with basic configuration

JoelWilcox:
This is a minimal FIR impl that handles generating the associated Dagger module with @ContributesTo and @Module annotations, as well as the expected @Binds method. Support for other configurations like ignoring qualifiers and replacing contributions will come as followup incremental changes.


compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/BindingModuleData.kt
```diff
@@ -0,0 +1,85 @@
...
+import org.jetbrains.kotlin.fir.symbols.impl.FirClassSymbol
+import org.jetbrains.kotlin.fir.types.ConeKotlinType
+import org.jetbrains.kotlin.fir.types.builder.buildResolvedTypeRef
+import org.jetbrains.kotlin.fir.types.classId
+import org.jetbrains.kotlin.fir.types.constructClassLikeType
+import org.jetbrains.kotlin.name.ClassId
+import org.jetbrains.kotlin.name.Name
+
+@OptIn(ExperimentalTopLevelDeclarationsGenerationApi::class)
+public class BindingModuleData(
```
JoelWilcox, line 29: I kept the structure here mostly in-line with the POC branch until we can discuss concerns/alternatives (related https://github.com/square/anvil/pull/1110/files#r1951710072), and also because I'd like us to get the merging impl + related integration tests in before doing any heavier refactoring.

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/util/AnvilNames.kt
```diff
@@ -0,0 +1,10 @@
+package com.squareup.anvil.compiler.k2.util
+
+import com.squareup.anvil.annotations.ContributesBinding
+import com.squareup.anvil.annotations.ContributesTo
+import com.squareup.anvil.compiler.internal.fqName
+
+internal object AnvilNames {
+  val contributesBinding = ContributesBinding::class.fqName
+  val contributesTo = ContributesTo::class.fqName
```
RBusarow, line 9: The entire `:compiler-utils` dependency is being added for the sake of this `fqName` extension and the `joinSimpleNames` extension.  Whatever API we wind up using for defining the `FqName`s, it would be good if we could just define the values using strings.

compiler-k2/src/main/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/ContributesBindingSessionComponent.kt
```diff
@@ -0,0 +1,57 @@
...
+   * generated module IDs.
+   * E.g. Key: "Foo_BindingModule", Value: ClassSymbol<Foo>
+   */
+  public val generatedIdsToMatchedSymbols: Map<ClassId, FirClassSymbol<*>> by lazy {
+    session.predicateBasedProvider.getSymbolsByPredicate(hasContributesBindingAnnotation)
+      .filterIsInstance<FirClassSymbol<*>>()
+      .associateBy {
+        it.classId.asClassName()
+          .joinSimpleNames(suffix = "_BindingModule")
+          .asClassId()
```
RBusarow, line 36: We should really just have a `ClassId` version of `joinSimpleNames`.

```kotlin

/**
 * Joins the simple names of a class with the given [separator], [prefix], and [suffix].
 *
 * ```
 * val normalName = ClassName("com.example", "Outer", "Middle", "Inner")
 * val joinedName = normalName.joinSimpleNames(separator = "_", suffix = "Factory")
 *
 * println(joinedName) // com.example.Outer_Middle_InnerFactory
 * ```
 * @throws IllegalArgumentException if the resulting class name is too long to be a valid file name.
 */
@ExperimentalAnvilApi
public fun ClassId.joinSimpleNames(
  separator: String = "_",
  prefix: String = "",
  suffix: String = "",
): ClassId = ClassId.topLevel(
  packageFqName.child(
    Name.identifier(
      relativeClassName.pathSegments()
        .joinToString(separator = separator, prefix = prefix, postfix = suffix),
    ),
  ),
)
```

compiler-k2/build.gradle.kts
```diff
@@ -23,13 +23,18 @@ dependencies {
   api(project(":compiler-k2-api"))
   api(libs.kotlin.compiler)
 
+  implementation(project(":compiler-utils"))
+  implementation(libs.dagger2)
```
RBusarow, line 27: If we just use `ClassId.joinSimpleNames(...)` and define the FqNames using strings, we can drop both these dependencies.

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CommonNames.kt
```diff
@@ -42,7 +42,7 @@ internal operator fun ClassLoader.get(fqName: String): Class<*> = loadClass(fqNa
 internal operator fun ScanResult.get(fqName: FqName): ClassInfo = getClassInfo(fqName)
 internal operator fun ScanResult.get(fqName: String): ClassInfo = getClassInfo(fqName)
 
-internal val Compile2Result.contributingObject: Class<*>
+public val Compile2Result.contributingObject: Class<*>
```
RBusarow, line 45: Sorry, I didn't notice this since I only wrote tests in this project.  I'll get the rest of the visibility changes.

compiler-k2/build.gradle.kts
```diff
@@ -23,13 +23,18 @@ dependencies {
...
+
   kapt(libs.auto.service.processor)
 
   testCompileOnly(libs.auto.service.annotations)
 
   testImplementation(project(":compiler-testing"))
+  testImplementation(project(":compiler-utils"))
```
RBusarow, line 34: This is unused.

compiler-k2/build.gradle.kts
```diff
@@ -23,13 +23,18 @@ dependencies {
...
   testImplementation(libs.junit.jupiter.api)
   testImplementation(libs.kase)
   testImplementation(libs.kotest.assertions.api)
   testImplementation(libs.kotest.assertions.core.jvm)
+  testImplementation(libs.truth)
```
RBusarow, line 39: 😭 

Could you give the Kotest apis a shot?

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/ContributesBindingFirExtensionTest.kt
```diff
@@ -0,0 +1,101 @@
...
+
+      @ContributesBinding(
+        scope = Int::class, 
+        boundType = ParentInterface::class
+      )
+      object ContributingObject : ParentInterface
+      """.trimIndent(),
+    ) {
+      // This is an implicit assertion and will fail if the annotation is missing
+      contributingObjectBindingModule.daggerModule
```
RBusarow, line 66:
```suggestion
      classGraph.shouldContainClass(CommonNames.contributingObject.bindingModule)
```

That `shouldContainClass` should be infix.  That's my bad.  The `bindingModule` extension doesn't exist, but it should.

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/TestUtils.kt
RBusarow, line 1: Why not put this in `compiler-testing`?

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/CommonNames.kt
JoelWilcox: All good! I think defaulting to limited visibility until it's needed makes sense anyway

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/ContributesBindingFirExtensionTest.kt
JoelWilcox: Similar thoughts here as my response to the kotest comment for the syntactic changes, but would you like to still see a more explicit assert instead of the implict one?

compiler-k2/build.gradle.kts
RBusarow: The majority of tests written within the last year have used Kotest assertions -- any time there was a new test class.  All assertions in new test tooling use Kotest assertions.  Going back to creating new test classes with Truth is an unnecessary step backwards.

Many of the existing tools make assumptions about hints, generated file names, KCT, and other K1 implementation details that don't really transfer over.  The allegories I'm writing in the K2 tooling do not make those assumptions, specifically so that they can have the interop we need.

compiler-k2/src/test/kotlin/com/squareup/anvil/compiler/k2/fir/contributions/ContributesBindingFirExtensionTest.kt
RBusarow: It's not the end of the world, but yes.  It's one line that's clear about its purpose and gives an `AssertionError` as its exception.

compiler-k2/build.gradle.kts
JoelWilcox: Updating for posterity: discussed offline with Rick and we agreed to focus on using Kotest for new test suites going forward to prioritize consistency across the K2 tests :+1:

RBusarow (approved):

PR #1116, opened 2025-02-18
`:compiler-k2-utils` and complementary testing apis

RBusarow:


compiler-k2-utils/src/main/kotlin/com/squareup/anvil/compiler/k2/utils/names/ClassIds.kt
```diff
@@ -0,0 +1,167 @@
...
+ * val entry by kotlinMap.child()
+ *
+ * entry.asFqNameString() shouldBe "kotlin.collections.Map.Entry"
+ * ```
+ * @see ClassId.createNestedClassId
+ */
+public fun ClassId.child(nameString: String): ClassId = nested(nameString)
+
+/**
+ * alias for [com.squareup.anvil.compiler.testing.nested]
```
JoelWilcox, line 154: Looks like this may be copy pasta / could use an update

compiler-k2-utils/build.gradle.kts
JoelWilcox, line 1: Not something that we need to decide now but I'm wondering if we want to treat this module similarly to the existing `compiler-utils` module where it's expressly not a stable API but still made available for consumers' convenience

compiler-k2-utils/src/main/kotlin/com/squareup/anvil/compiler/k2/utils/names/ClassIds.kt
RBusarow: Yep...  I'll get this at some point soon.

compiler-k2-utils/build.gradle.kts
RBusarow: That was my thinking as well.

PR #1123, opened 2025-02-21
WIP - consuming existing unit tests in K1 and K2.

RBusarow:


compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
```diff
@@ -0,0 +1,2819 @@
...
+import dagger.Lazy
+import dagger.internal.Factory
+import io.kotest.matchers.shouldBe
+import org.jetbrains.kotlin.cli.common.ExitCode
+import org.junit.jupiter.api.Disabled
+import org.junit.jupiter.api.TestFactory
+import java.io.File
+import javax.inject.Provider
+
+class InjectConstructorFactoryGeneratorTest : CompilationModeTest() {
```
JoelWilcox, line 20: It looks like this is duplicated instead of moved right now, can we delete the original? This will help to prevent unintentional diverging between the two as well as preserve the git history which can be really helpful for some test cases. Also with this test being as large as it is, it will help to highlight and confirm what the new changes are.

compiler/build.gradle.kts
```diff
@@ -17,7 +17,7 @@ conventions {
   kotlinCompilerArgs.addAll(
     // The flag is needed because we extend an interface that uses @JvmDefault and the Kotlin
     // compiler requires this flag when doing so.
-    "-Xjvm-default=all",
+    // "-Xjvm-default=all",
```
JoelWilcox, line 20: Delete or uncomment for this

compiler/build.gradle.kts
```diff
@@ -60,6 +61,6 @@ dependencies {
   testImplementation(libs.truth)
 
   testRuntimeOnly(libs.kotest.assertions.core.jvm)
-  testRuntimeOnly(libs.junit.vintage.engine)
+  // testRuntimeOnly(libs.junit.vintage.engine)
```
JoelWilcox, line 64: Same here

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
```diff
@@ -0,0 +1,2819 @@
...
+      
+      import javax.inject.Inject
+      
+      class InjectClass @Inject constructor() {
+        @Inject constructor(string: String)
+      }
+      """,
+      expectedExitCode = ExitCode.COMPILATION_ERROR,
+    ) {
+      // assertThat(
```
JoelWilcox, line 2635: Is this one intentionally commented out? Does it just need a skipWhen(mode.isK2)?

compiler-testing/src/main/kotlin/com/squareup/anvil/compiler/testing/compilation/Compile2.kt
```diff
@@ -85,70 +86,106 @@ public class Compile2Compilation(
...
         .toTypedArray()
 
+      if (config.languageVersion < LanguageVersion.KOTLIN_2_0) {
+        val anvilPluginOptions = buildList {
+
+          fun option(optionName: String, value: String) {
+            add("plugin:com.squareup.anvil.compiler:$optionName=$value")
+          }
+
+          if (!config.mode.isK2) {
```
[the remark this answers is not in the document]
RBusarow: heh.. yep!  Once upon a time, I thought this was the right spot for the branching.  Then I forgot to delete this.

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
RBusarow: I hadn't noticed this in the diff.  I'd certainly prefer to keep its git history.  I'll figure out what I need to do to fix that.

RBusarow: I split it into two different commits.  It still doesn't present nicely in the PR diff, but the history is preserved. 🤷 

image

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
RBusarow: I commented it out because I needed to move some utilities around before it would compile, then it slipped through the cracks.  Thanks for catching it!

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
JoelWilcox: Awesome thanks for taking the time to adjust it! Unfortunately yea sometimes github doesn't handle the diff the same as git goes

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
JoelWilcox, line 2635: This test is showing as commented out for me still, leaving a comment in case that was meant to be included in the last push. Otherwise everything looks good!

JoelWilcox (approved):

compiler-tests/src/test/java/com/squareup/anvil/compiler/dagger/InjectConstructorFactoryGeneratorTest.kt
RBusarow: I'll email you a zip file with the change.

PR #1153, opened 2025-10-01
Upgrade Kotlin to 2.2.20 and Dagger to 2.57.2

JoelWilcox:
- Also changes constructor visibility to private for generated Dagger `Factory`s and `MembersInjector`s to match code generation changes introduced in Dagger 2.57


compiler/src/test/java/com/squareup/anvil/compiler/TestUtils.kt
```diff
@@ -356,6 +357,15 @@ fun Class<*>.contributedProperties(): List<KClass<*>>? {
     .filterIsInstance<KClass<*>>()
 }
 
+/**
+ * Assumes that there is only one constructor. For MembersInjector factories in particular, we
+ * make their constructor private to match behavior in Dagger 2.57, but under the hood Kotlin
+ * generates a public synthetic constructor with additional parameters that we must filter out.
+ */
+internal fun Class<*>.primaryConstructor(): Constructor<*> {
```
vRallev, line 365: Not sure if this is the best name. When reading `primaryConstructor()` I think of Kotlin's primary constructor. But this isn't necessarily what this function returns. Maybe `singleConstructor()`? Not the end of the world, though.

vRallev (approved):

compiler/src/test/java/com/squareup/anvil/compiler/TestUtils.kt
JoelWilcox: That's fair, `singleConstructor` sounds like a good alternative to me

vRallev (approved):
Fix looks good!

